PR #195, opened 2015-01-08
Cfy 1863 support installation arguments

noakup:


plugins/plugin-installer/plugin_installer/tasks.py
```diff
@@ -54,12 +54,18 @@ def install(ctx, plugins, **kwargs):
...
+    logger.debug('Installing {0} from {1} with args: {2}'
+                 .format(name, url, installation_args))
+
+    plugin_temp_dir = extract_plugin_dir(url)
+
+    install_package(url, installation_args, plugin_temp_dir)
+    module_paths = extract_module_paths(plugin_temp_dir)
     update_includes(module_paths)
 
+    shutil.rmtree(plugin_temp_dir)
```
iliapolo: this should be within a 'finally' clause.

plugins/plugin-installer/plugin_installer/tasks.py
```diff
@@ -79,23 +85,33 @@ def update_includes(module_paths, includes_path=None):
...
 
 
-def install_package(url):
+def install_package(url, installation_args, plugin_dir):
 
     """
     Installs a package onto the worker's virtualenv.
 
-    :param url: A URL to the package archive.
+    :param url:                 A URL to the package archive.
```
iliapolo: looks like this parameter is not used anymore

plugins/plugin-installer/plugin_installer/tasks.py
```diff
@@ -110,57 +126,85 @@ def extract_module_paths(url):
...
+                           download_dir=None,
+                           only_download=False)
+        return plugin_dir
+    except Exception as e:
+        if os.path.exists(plugin_dir):
+            shutil.rmtree(plugin_dir)
+        raise NonRecoverableError('Failed to download and unpack plugin from '
+                                  '{0}. Reported error: {1}'
+                                  .format(plugin_url, e.message))
+
```
iliapolo: we use str(e) instead of e.message since it usually contains more information

plugins/plugin-installer/plugin_installer/tasks.py
```diff
@@ -110,57 +126,85 @@ def extract_module_paths(url):
...
+        req_set = pip.req.RequirementSet(build_dir=None,
+                                         src_dir=None,
+                                         download_dir=None)
+        req_set.unpack_url(link=pip.index.Link(plugin_url),
+                           location=plugin_dir,
+                           download_dir=None,
+                           only_download=False)
+        return plugin_dir
+    except Exception as e:
+        if os.path.exists(plugin_dir):
```
iliapolo: 'plugin_dir' may not be defined if tempfile.mkdtemp() fails. its a very unlikely scenario, but still its better to be protected from it.

plugins/plugin-installer/plugin_installer/tasks.py
```diff
@@ -110,57 +126,85 @@ def extract_module_paths(url):
...
 
-def get_url(blueprint_id, plugin):
+def get_url_and_args(blueprint_id, plugin_dict):
 
-    source = plugin['source']
+    source = plugin_dict['source']
+    if source:
+        source = str(source).strip()
+    else:
+        raise NonRecoverableError('Plugin source is missing')
```
iliapolo: if 'source' is not in plugin_dict than a KeyError will be raised and this code wont be reached.
a better way to do it is 
source = plugin_dict.get('source','').strip()

plugins/plugin-installer/plugin_installer/tasks.py
```diff
@@ -110,57 +126,85 @@ def extract_module_paths(url):
...
+
+    install_args = ''
+    if 'installation_args' in plugin_dict:
+        install_args = plugin_dict['installation_args']
+
+    if install_args:
+        install_args = str(install_args).strip()
+
```
iliapolo: no need to use str() here...it can either be a string or a unicode object, both of which have the strip() method.
also, the same logic from above can be applied
install_args = plugin_dict.get('installation_args','').strip()

plugins/plugin-installer/plugin_installer/tasks.py
```diff
@@ -110,57 +126,85 @@ def extract_module_paths(url):
...
-    # Else, assume its a relative path from <blueprint_home>/plugins
-    # to a directory containing the plugin project.
-    # in this case, the archived plugin will reside in the manager file server.
-
-    blueprint_plugins_url = '{0}/{1}/plugins'.format(
-        utils.get_manager_file_server_blueprints_root_url(),
-        blueprint_id
-    )
-    return '{0}/{1}.zip'.format(blueprint_plugins_url, source)
+        if not (schema in ['http', 'https']):
```
iliapolo: this kind of condition is usually written as

if schema not in ['http', 'https']

plugins/plugin-installer/plugin_installer/tests/mock-with-install-args-plugin/requirements.txt
```diff
@@ -0,0 +1 @@
+https://github.com/cloudify-cosmo/softlayer-python/archive/master.zip
```
iliapolo: i think we plan on deleting this repo soon no? since we dont need this fork anymore...
maybe use cloudify-rest-client? its pretty light and should install quickly

PR #197, opened 2015-01-18
CFY-1884 added a test for ctx.operation.retry

idanmo:


tests/resources/dsl/operation_retry.yaml
```diff
@@ -0,0 +1,16 @@
+tosca_definitions_version: cloudify_dsl_1_0
+
+imports:
+  - cloudify.types
```
idanmo: change to cloudify/types/types.yaml

PR #201, opened 2015-01-23
CFY 2026 execute_operation kwargs override

ran-z:


resources/rest-service/cloudify/types/types.yaml
```diff
@@ -240,6 +240,8 @@ workflows:
       operation: {}
       operation_kwargs:
         default: {}
+      allow_kwargs_override:
+        default: null
```
dankilman: Why did you work so hard to enable a null value in the dsl parser when this property is in fact a boolean property, so the default value could simply be false?

PR #203, opened 2015-01-30
Failure healing policies tests fixes

simon-bar:
I have also updated one url in types.yaml


dankilman:
If I merge this pull request, the autoheal system test will fail as it is still using the failing_node property

simon-bar:
That's right. That's why we have prepared appropriate changes in cloudify-system-tests and the docs as well

PR #214, opened 2015-02-11
add amqp address to config

m1keil:


rest-service/manager_rest/server.py
```diff
@@ -155,26 +155,9 @@ def reset_state(configuration=None):
...
-    if 'file_server_uploaded_blueprints_folder' in yaml_conf:
-        obj_conf.file_server_uploaded_blueprints_folder = \
-            yaml_conf['file_server_uploaded_blueprints_folder']
-    if 'file_server_resources_uri' in yaml_conf:
-        obj_conf.file_server_resources_uri = \
-            yaml_conf['file_server_resources_uri']
-    if 'rest_service_log_path' in yaml_conf:
-        obj_conf.rest_service_log_path = \
-            yaml_conf['rest_service_log_path']
+    for key, value in yaml_conf.iteritems():
```
nir0s: This should be yaml_conf.items() rather than iteritems(). iteritems is deprecated.
wonderful consolidation though.

m1keil: that's py3 you are talking about. in py2 iteritems is returns generator and items returns list.

nir0s: It provides the same functionality anyhow.. just deprecated in 3..

rest-service/manager_rest/config.py
```diff
@@ -19,6 +19,7 @@ class Config(object):
     def __init__(self):
         self._db_address = 'localhost'
```
nir0s: I would actually make localhost be a private case for a host. As the db address should be abstracted by a name, I would change `db_address` and `amqp_address` to a default set of names, and for specific cases, would inject "localhost" instead.

PR #215, opened 2015-02-16
Cfy 2173 configure riemann for external hosts

nir0s:
Riemann now receives an environment variable called HOSTING_VM to allow it to connect to different host other than localhost when initiating an AMQP connection.


dankilman:
This PR will require setting the HOSTING_VM environment variable any time I want to locally execute an integration test. The default should be "localhost" and you should inject this env variable when needed

PR #227, opened 2015-03-13
Cfy 2289 runtime credentials worker installer

szpotona:


plugins/agent-installer/worker_installer/__init__.py
```diff
@@ -50,14 +50,24 @@ def wrapper(*args, **kwargs):
...
+            agent_config = kwargs['cloudify_agent']
+        else:
+            if ctx.type == context.NODE_INSTANCE and \
+                    ctx.node.properties.get('cloudify_agent'):
+                agent_config = ctx.node.properties['cloudify_agent']
+            else:
+                agent_config = {}
         prepare_connection_configuration(ctx, agent_config)
-
+        kwargs['cloudify_agent'] = agent_config
```
iliapolo: why is this needed? 
line 76 is the one who injects the agent_config dictionary in the kwargs as 'agent_config'. 

i don't see where this assignment is used

PR #239, opened 2015-03-31
Cfy 2263 modify file based userstore implementation to use user inputs

noakup:


rest-service/manager_rest/test/base_test.py
```diff
@@ -150,6 +150,43 @@ def create_configuration(self):
         test_config.file_server_resources_uri = FILE_SERVER_RESOURCES_URI
         test_config.rest_service_log_path = self.rest_service_log
 
+        # security config
+        test_config.secured_server = self._secured
+        if self._secured:
```
dankilman: why is this required here in the base test if it weren't required before?

PR #249, opened 2015-04-20
CFY-2504 if service does not exist, no status will be returned

adamlavie:


dankilman:
Merged by accident, so created a PR to revert and merged it as well. Sorry, need to open a new PR

PR #291, opened 2015-08-15
Cfy 2926 elasticsearc index rotation

nir0s:


rest-service/manager_rest/resources.py
```diff
@@ -1176,27 +1176,42 @@ def get(self, deployment_id, **_):
...
+        else:
+            return 'logstash-*'
+
     def _query_events(self):
         """
         List events for the provided Elasticsearch query
         """
         verify_json_content_type()
-        return _query_elastic_search(index='cloudify_events',
+        return _query_elastic_search(index=self._set_index_name(),
```
nir0s: research flask.g object to store both es connection object and make the index exists check only once.

PR #292, opened 2015-08-20
CFY-3316 use the common create_import_resolver method to get the reso…

yaelpeled:
…lver

### DIFF
--- a/rest-service/manager_rest/blueprints_manager.py
+++ b/rest-service/manager_rest/blueprints_manager.py
@@ -24,26 +24,17 @@ from dsl_parser import constants
 from dsl_parser import exceptions as parser_exceptions
 from dsl_parser import functions
 from dsl_parser import tasks
-from dsl_parser.import_resolver.default_import_resolver import \
-    DefaultResolverValidationException, \
-    DEFAULT_RESLOVER_RULES_KEY, \
-    DefaultImportResolver
+from dsl_parser import utils as dsl_parser_utils
 from manager_rest import models
 from manager_rest import manager_exceptions
 from manager_rest.workflow_client import workflow_client
 from manager_rest.storage_manager import get_storage_manager
 from manager_rest.utils import maybe_register_teardown
-from manager_rest.utils import get_class_instance
-
 
 LIMITLESS_GLOBAL_PARALLEL_EXECUTIONS_VALUE = -1
 
 
 class DslParseException(Exception):
-    pass
-
-
-class ResolverInstantiationError(Exception):
     pass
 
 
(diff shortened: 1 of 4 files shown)


rest-service/manager_rest/blueprints_manager.py
```diff
@@ -890,29 +873,9 @@ def _get_resolver(self):
...
-                except DefaultResolverValidationException, ex:
-                    raise ResolverInstantiationError(
-                        'Wrong parameters ({0}) configured for '
-                        'the default resolver ({1}): {2}'
-                        .format(params,
-                                DefaultImportResolver.__name__, str(ex)))
-        current_app.resolver = resolver
+        resolver_section = self._get_resolver_section(context)
+        current_app.resolver = dsl_parser_utils.create_import_resolver(
+            resolver_section)
```
noakup: regarding update_provider_context (not modified here) - why does it contain _update_import_resolver_in_app? should it be separate? seems like the meaning is not the same

rest-service/manager_rest/resources.py
```diff
@@ -1383,7 +1384,7 @@ def post(self):
             get_blueprints_manager().update_provider_context(update, context)
```
noakup: in fact you just want to check _update_import_resolver_in_app, that is called as part of update_provider_context. seems to me like it should be extracted from it and called separately.

rest-service/manager_rest/test/test_import_resolver.py
```diff
@@ -58,195 +42,52 @@ def _update_provider_context(self, resolver_section=None):
...
+        self.assertEqual(self.app.application.resolver,
+                         'mock expected import resolver')
+
+    def test_resolver_update_in_app(self):
+        # the next test updates the resolver in the app
+        self.test_upload_blueprint_with_resolver()
+        curr_app_resolver = self.app.application.resolver
         # upload blueprint again and check that
         # the expected resolver passed to the parser
         with mock.patch('dsl_parser.tasks.parse_dsl') as mock_parse_dsl:
```
noakup: I think you already test the same thing in "test_upload_blueprint_with_resolver"

rest-service/manager_rest/utils.py
```diff
@@ -103,7 +103,7 @@ def get_class(class_path):
 
     module = importlib.import_module(class_module_str)
     if not hasattr(module, class_name):
-        raise ValueError('module "{0}", does not contain class "{1}"'
+        raise ValueError('module {0}, does not contain class {1}'
                          .format(class_module_str, class_name))
```
noakup: let's remove that comma, I don't know what I was thinking!

PR #306, opened 2015-09-10
CFY 3397 REST api for setting global parallel execution limit

ran-z:


rest-service/manager_rest/resources_v2.py
```diff
@@ -18,13 +18,18 @@
 
 from manager_rest import resources
 from manager_rest.resources import (marshal_with,
-                                    exceptions_handled)
+                                    exceptions_handled,
+                                    verify_json_content_type,
+                                    verify_parameter_in_request_body)
 from manager_rest.resources import verify_and_convert_bool
 from manager_rest import models
+from manager_rest import responses
```
dankilman: Instead of importing `responses`, you should follow the current convention of importing `ProviderContext` in `responses_v2` and use it. If you disagree with this convention, then all other places should follow the same pattern.

rest-service/manager_rest/resources_v2.py
```diff
@@ -300,3 +305,54 @@ def get(self, _include=None, filters=None, **kwargs):
...
+        """
+        modifies provider context configuration
+        """
+        verify_json_content_type()
+        request_json = request.json
+        verify_parameter_in_request_body('global_parallel_executions_limit',
+                                         request_json)
+
+        provider_ctx = get_storage_manager().get_provider_context()
+        bootstrap_ctx = provider_ctx.context['cloudify']
```
dankilman: I think the safer `context.get('cloudify', {})` is preferable.

rest-service/manager_rest/resources_v2.py
```diff
@@ -300,3 +305,54 @@ def get(self, _include=None, filters=None, **kwargs):
...
+                ' deployment workers mode is disabled')
+
+        limit = request_json['global_parallel_executions_limit']
+        if type(limit) is not int:
+            raise manager_exceptions.BadParametersError(
+                'global_parallel_executions_limit parameter should be of type'
+                ' int, but is instead of type {0}'.format(
+                    type(limit).__name__))
+
+        bootstrap_ctx['transient_deployment_workers_mode'][
```
dankilman: if you take the `get` approach suggested previously and we switch to default `True` then `transient_deployment_workers_mode` may not exist in `bootstrap_ctx`. Also, you will need to put it back into `provider_ctx.context`

PR #310, opened 2015-09-30
CFY-3035 CFY-3040 snapshots and agents migration

konradxyz:


nir0s:
Please solve conflicts.

resources/rest-service/cloudify/install_agent.py
```diff
@@ -0,0 +1,199 @@
...
+    return logger
+
+
+def _return(value, old_agent_version):
+    ctx.returns(value)
+    # Due to bug in celery:
+    if os.name == 'nt' and old_agent_version.startswith('3.2'):
+        from celery import current_task
+        from cloudify.celery import celery
+        celery.backend.mark_as_done(current_task.request.id, value)
```
dankilman: can you provide some details on this?

konradxyz: The same issue was mentioned in here:
https://github.com/cloudify-cosmo/cloudify-agent/blob/master/cloudify_agent/app.py#L74
I am not sure what is an exact cause of issue.
I've noticed that for windows all tasks' backends are disabled - event though backend of main celery object is properly configured. So results are not sent back to the rabbitmq.
This is not an issue anymore because new agents are using  workaround with setting backend_url via configuration update rather than constructor.

resources/rest-service/cloudify/install_agent.py
```diff
@@ -0,0 +1,199 @@
+import json
```
idanmo: Missing license header.

rest-service/manager_rest/blueprints_manager.py
```diff
@@ -61,6 +60,9 @@ def deployments_list(self, include=None, filters=None, pagination=None):
         return self.sm.deployments_list(include=include, filters=filters,
                                         pagination=pagination)
 
+    def snapshots_list(self, include=None):
```
idanmo: This call should support pagination, please look here:
https://github.com/cloudify-cosmo/cloudify-manager/pull/322/files

rest-service/manager_rest/blueprints_manager.py
```diff
@@ -102,16 +107,80 @@ def update_execution_status(self, execution_id, status, error):
...
+                    wf_id=wf_id,
+                    task_mapping=deployment_env_creation_task_name,
+                    deployment=deployment,
+                    execution_parameters=exec_params)
 
         return self.sm.update_execution_status(execution_id, status, error)
 
+    def _get_conf_for_snapshots_wf(self):
+        config_instance = config.instance()
+        snapshots_conf = {
```
idanmo: For making the code more readable can you change it to something like:
```python
snapshots_conf = {
  'file_server_root': config.instance().file_server_root,
  ...
}
```

rest-service/manager_rest/blueprints_manager.py
```diff
@@ -260,7 +336,29 @@ def execute_workflow(self, deployment_id, workflow_id,
 
         return new_execution
 
-    def _execute_system_workflow(self, deployment, wf_id, task_mapping,
+    def _check_for_any_active_executions(self):
+        executions = [
+            e.id
+            for e in self.executions_list(is_include_system_workflows=True)
```
idanmo: `self.executions_list` accepts a filter argument where you can filter by multiple states instead of reading all executions.

rest-service/manager_rest/blueprints_manager.py
```diff
@@ -260,7 +336,29 @@ def execute_workflow(self, deployment_id, workflow_id,
...
+        if executions:
+            raise manager_exceptions.ExistingRunningExecutionError(
+                'You cannot start a system-wide execution if there are '
+                'other executions running. '
+                'Currently running executions: {0}'
+                .format(executions))
+
+    def _check_for_active_system_wide_execution(self):
+        for e in self.executions_list(is_include_system_workflows=True):
+            if e.status not in e.END_STATES and e.deployment_id is None:
```
idanmo: Same comment as above.

rest-service/manager_rest/blueprints_manager.py
```diff
@@ -807,25 +914,27 @@ def create_filters_dict(**kwargs):
                 filters[key] = val
         return filters or None
 
-    def _create_deployment_environment(self, deployment, deployment_plan):
+    def _create_deployment_environment(self, deployment, dep_plan):
```
idanmo: Please change back to `deployment_plan`

rest-service/manager_rest/es_storage_manager.py
```diff
@@ -190,6 +192,9 @@ def blueprints_list(self, include=None, filters=None, pagination=None):
                                     filters=filters,
                                     include=include)
 
+    def snapshots_list(self, include=None):
```
idanmo: Add support for filters and pagination.

rest-service/manager_rest/resources.py
```diff
@@ -252,8 +237,7 @@ def _process_plugins(self, file_server_root, blueprint_id):
                                         'plugins', final_zip_name)
             self._zip_dir(plugin_dir, target_zip_path)
 
-    @staticmethod
-    def _zip_dir(dir_to_zip, target_zip_path):
+    def _zip_dir(self, dir_to_zip, target_zip_path):
```
idanmo: Why not `@staticmethod`?

rest-service/manager_rest/resources_v2.py
```diff
@@ -92,6 +93,204 @@ def _create_filter_params_list_description(parameters, list_type):
...
+class Snapshots(SecuredResource):
+
+    @swagger.operation(
+        responseClass='List[{0}]'.format(responses_v2.Snapshot.__name__),
+        nickname='list',
+        notes='Returns a list of existing snapshots.'
+    )
+    @exceptions_handled
+    @marshal_with(responses_v2.Snapshot)
+    def get(self, _include=None):
```
idanmo: Add support for filters and pagination (see Plugins for example).

rest-service/manager_rest/resources_v2.py
```diff
@@ -92,6 +93,204 @@ def _create_filter_params_list_description(parameters, list_type):
...
+        consumes=[
+            "application/json"
+        ]
+    )
+    @exceptions_handled
+    @marshal_with(responses_v2.Execution)
+    def put(self, snapshot_id):
+        verify_json_content_type()
+        request_json = request.json
+        verify_parameter_in_request_body('include_metrics', request_json)
```
idanmo: Verification can include the parameter type (`param_type` argument).
Perhaps we should have default values for these parameters and make them optional?

rest-service/manager_rest/resources.py
konradxyz: It seems like unnecessary change.
I will change all methods that does not require access to object instance to classmethods  - they are calling each other frequently and it is easier to use cls.method_name than UploadedBlueprintsManager.method_name

rest-service/manager_rest/resources_v2.py
```diff
@@ -92,6 +93,204 @@ def _create_filter_params_list_description(parameters, list_type):
...
+            'allowMultiple': False,
+            'dataType': 'binary',
+            'paramType': 'body'}],
+        consumes=[
+            "application/octet-stream"
+        ]
+    )
+    @exceptions_handled
+    @marshal_with(responses_v2.Snapshot)
+    def put(self, snapshot_id):
```
idanmo: This API is not consistent with other APIs we have.
/archive is only used for downloading the resource (for instance, /blueprints/<id>/archive).
For being consistent, upload snapshot from URL should be a flag in request body as implemented for uploading a blueprint from URL.

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+    generate_create_dep_tasks_graph
+
+
+_METADATA_FILE = 'metadata.json'
+
+# metadata fields
+_M_HAS_CLOUDIFY_EVENTS = 'has_cloudify_events'
+_M_VERSION = 'snapshot_version'
+
+_VERSION = '3.3'
```
idanmo: The version should be retrieved by using REST API's version call.
Note that the version includes milestone information as well (don't know if relevant here).

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+        try:
+            while s:
+                obj, idx = decoder.raw_decode(s)
+                n += 1
+                yield json.dumps(obj)
+                s = s[idx:]
+        except:
+            pass
+
+    assert not n or not s
```
idanmo: Perhaps raise an appropriate exception with a reasonable error message instead of this assertion?

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+
+
+def _copy_data(archive_root, config, to_archive=True):
```
idanmo: This method is a little bit complicated for what it is supposed to do :/
Adding a small paragraph explaining the logic would also be good and state that `to_archive=False` is passed when restoring a snapshot.

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+
+    if es_node['_type'] == 'blueprint':
+        source = es_node['_source']
+        if 'description' not in source:
+            source['description'] = ''
+
+
+def _restore_elasticsearch(ctx, tempdir, es, metadata):
+    ctx.send_event('Deleting all ElasticSearch data')
+    elasticsearch.helpers.bulk(
```
idanmo: Why is it necessary to erase all data from ES?
Isn't there a validation for not allowing to restore a snapshot if the manager is not "clean"?

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+        os.makedirs(snapshots_dir)
+
+    metadata = {}
+
+    # files/dirs copy
+    _copy_data(tempdir, config)
+
+    # elasticsearch
+    es = _create_es_client(config)
+    has_cloudify_events = es.indices.exists(index=_EVENTS_INDEX_NAME)
```
idanmo: We changed logstash configuration to create a new events index per day/week and the index name is: "logstash-*".
Look here: https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/resources.py#L1183

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+    os.makedirs(snapshot_dir)
+
+    shutil.make_archive(
+        os.path.join(snapshot_dir, snapshot_id),
+        'zip',
+        tempdir
+    )
+
+    # end
+    shutil.rmtree(tempdir)
```
idanmo: Perhaps use the try..finally pattern here for making sure everything gets cleaned up in a case of failure.

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+    storage_scan = (e for e in storage_scan if e['_id'] != execution_id)
+
+    event_scan = elasticsearch.helpers.scan(
+        es,
+        index=_EVENTS_INDEX_NAME if has_cloudify_events else 'logstash-*'
+    )
+
+    with open(os.path.join(tempdir, _ELASTICSEARCH), 'w') as f:
+        for item in itertools.chain(storage_scan, event_scan):
+            f.write(json.dumps(item) + '\n')
```
idanmo: Use `os.linesep` instead of `\n`

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+
+
+def _dump_influxdb(tempdir):
+    influxdb_file = os.path.join(tempdir, _INFLUXDB)
+    influxdb_temp_file = influxdb_file + '.temp'
+    subprocess.call(_INFLUXDB_DUMP_CMD.format(influxdb_temp_file), shell=True)
```
idanmo: Check the status code of this process execution?

rest-service/manager_rest/resources_v2.py
xwombat: Upload snapshot from URL is a flag in request body which is managed by UploadedDataManager (in files.py) from which UploadedSnapshotsManager and UploadedBlueprintsManager are deriving, so behavior in snapshots and blueprints is identical.
Regarding upload snapshots method in SnapshotsIdArchive endpoint, we did it because in SnapshotId endpoint there already exists put method that creates snapshot and it is consistent with deployments where there is also put method for creating deployment in DeploymentId endpoint. And because of that we chose SnapshotsIdArchive as place where to put it.
If you want we can, for example, put it in SnapshotsId as post method.

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+    with open(influxdb_temp_file, 'r') as f, open(influxdb_file, 'w') as g:
+        for obj in _get_json_objects(f):
+            g.write(obj + '\n')
+
+    os.remove(influxdb_temp_file)
+
+
+def _is_compute(node):
+    return 'cloudify.nodes.Compute' in node.type_hierarchy
```
idanmo: `cloudify.nodes.Compute` can be imported from `cloudify.constants.COMPUTE_NODE_TYPE`

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+                    'node_instance_ids': {
+                        'default': []
+                    }
+                },
+                'plugin': 'default_workflows'
+            }
+    if es_node['_type'] == 'node':
+        source = es_node['_source']
+        type_hierarchy = source.get('type_hierarchy', [])
+        if 'cloudify.nodes.Compute' in type_hierarchy:
```
idanmo: Can be imported from `cloudify.constants`

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+                    'retry_interval': None,
+                    'max_retries': None,
+                    'executor': 'central_deployment_agent',
+                    'operation': 'cloudify_agent.operations.create_agent_amqp'
+                }
+
+    if es_node['_type'] == 'blueprint':
+        source = es_node['_source']
+        if 'description' not in source:
+            source['description'] = ''
```
idanmo: There's an additional field: `main_file_name` which holds the name of the main blueprint file.
Look here: https://github.com/cloudify-cosmo/cloudify-manager/pull/309/files

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+
+    # cloudify_events -> cloudify_events, logstash-* -> logstash-*
+    def get_data_itr():
+        for line in open(os.path.join(tempdir, _ELASTICSEARCH), 'r'):
+            elem = json.loads(line)
+            _update_es_node(elem)
+            yield elem
+
+    # logstash-* -> cloudify_events
+    def l_to_ce():
```
idanmo: Perhaps name this function better? logstash_to_cloudify_events?

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+    ctx.send_event('Restoring ElasticSearch data')
+    elasticsearch.helpers.bulk(es, data_iter)
+    es.indices.flush()
+
+
+def _restore_influxdb_3_3(ctx, tempdir):
+    ctx.send_event('Restoring InfluxDB metrics')
+    influxdb_f = os.path.join(tempdir, _INFLUXDB)
+    if os.path.exists(influxdb_f):
+        subprocess.call(_INFLUXDB_RESTORE_CMD.format(influxdb_f), shell=True)
```
idanmo: Check process exit code

workflows/cloudify_system_workflows/snapshot.py
xwombat: We have already implemented that validation (plus flag to control it) but it will be pushed as separate pull request.

workflows/cloudify_system_workflows/snapshot.py
xwombat: Yes, and inside _dump_elasticsearch that is correctly handled.

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -0,0 +1,574 @@
...
+            tasks_graph.execute()
+        finally:
+            dep_ctx.internal.stop_local_tasks_processing()
+            dep_ctx.internal.stop_event_monitor()
+        ctx.send_event('Successfully created {0}\'s deployment '
+                       'environment'.format(dep_id))
+
+
+@workflow(system_wide=True)
+def restore(ctx, snapshot_id, recreate_deployments_envs, config, **kwargs):
```
idanmo: Perhaps it would be good to add some logs during the process of creating a snapshot or restoring one.

workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -119,14 +115,24 @@ def create(ctx, **kwargs):
...
         ctx.send_event('Starting deployment policy engine core'),
         ctx.execute_task('riemann_controller.tasks.create',
-                         kwargs=kwargs.get('policy_configuration', {})))
+                         kwargs=policy_configuration or {}))
 
+    return graph
+
+
+@workflow
+def create(ctx, deployment_plugins_to_install, workflow_plugins_to_install,
```
idanmo: `from cloudify import ctx` ?

workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -119,14 +115,24 @@ def create(ctx, **kwargs):
...
+    graph = generate_create_dep_tasks_graph(
+        ctx,
+        deployment_plugins_to_install,
+        workflow_plugins_to_install,
+        policy_configuration)
     return graph.execute()
 
 
 @workflow
 def delete(ctx, **kwargs):
```
idanmo: `from cloudify import ctx` ?

workflows/cloudify_system_workflows/snapshot.py
idanmo: `from cloudify import ctx` ?

xwombat: Do you think that we should add logs or events? What is general convention to choose between them?

rest-service/manager_rest/resources_v2.py
idanmo: Lets decide after we merge the initial implementation.

workflows/cloudify_system_workflows/snapshot.py
idanmo: But even if you decide to restore the snapshot (force), its probably not a good idea to clear all data from ES. ES may contain data the user needs..

workflows/cloudify_system_workflows/deployment_environment.py
simon-bar: Hmm it has always been like this, but we can change it, why not. It shouldn't break anything.

simon-bar: Well, maybe we should leave it. It should be a separate PR in my opinion, as it is unrelated to snapshots..

PR #318, opened 2015-10-21
CFY-3592 support multiple values in filters

ChenRoth:
introduce multiple-value filters to REST service v2

### DIFF
--- a/rest-service/manager_rest/test/test_list_filters.py
+++ b/rest-service/manager_rest/test/test_list_filters.py
@@ -48,6 +48,13 @@ class ResourceListFiltersTestCase(BaseListTest):
                                       'filters {0}, got {1}'
                                       .format(filter_fields, response[0]))
 
+    def test_deployments_list_with_filters_multiple_values(self):
+        filter_fields = \
+            {'id': [self.first_deployment_id, self.sec_deployment_id]}
+        self._test_multiple_values_filter('deployments',
+                                          filter_fields,
+                                          2)
+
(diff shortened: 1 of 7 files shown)


rest-service/manager_rest/test/test_list_filters.py
```diff
@@ -280,3 +320,29 @@ def test_plugins_list_no_filters(self):
             self.assertIn(plugin['uploaded_at'],
                           (first_plugin_response['uploaded_at'],
                            sec_plugin_response['uploaded_at']))
+
+    def _test_multiple_values_filter(self, url, filter_fields, expected_count):
+        type_name = url.replace('-', ' ').replace('/', '')
```
ChenRoth: use client.x.list() instead of get()

tests/workflow_tests/test_rest_service_list_filters.py
```diff
@@ -31,6 +31,11 @@ def setUp(self):
             self.sec_blueprint_id = self._put_two_deployments()
         self.deployment_id_filter = {'deployment_id': self.first_deployment_id}
         self.blueprint_id_filter = {'blueprint_id': self.first_blueprint_id}
+        self.multiple_value_filters = \
+            {'deployment_id': [self.first_deployment_id,
+                               self.sec_deployment_id],
+             'node_id': ['webserver',
+                         'compute']}
```
ChenRoth: belongs in test scope, move it there

PR #322, opened 2015-10-25
CFY-3904 Use elastic search pagination capabilities

dankilman:


rest-service/manager_rest/es_storage_manager.py
```diff
@@ -152,36 +151,28 @@ def _build_filter_terms_query(filters=None):
...
         """
         terms_lst = []
-        query = None
+        body = {}
+        if pagination:
+            if not skip_size:
+                body['size'] = pagination.get('page_size', DEFAULT_SEARCH_SIZE)
+            if 'offset' in pagination:
+                body['from'] = pagination['offset']
+        elif not skip_size:
```
adamlavie: I am guessing you added this parameter in order to be able to get all the results and not be confined by a limit.

rest-service/manager_rest/es_storage_manager.py
```diff
@@ -152,36 +151,28 @@ def _build_filter_terms_query(filters=None):
         :return: an elasticsearch query string containing the given filters
```
adamlavie: change this to be a bit more accurate

rest-service/manager_rest/es_storage_manager.py
dankilman: i added this so the "delete deployment" doesn't fail with "unsupported parameter: size"

PR #325, opened 2015-10-28
CFY-3682 - Add implementation for authrization based on roles

noakup:


rest-service/manager_rest/cloudify_acl_handler.py
```diff
@@ -0,0 +1,27 @@
...
+#  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  * See the License for the specific language governing permissions and
+#  * limitations under the License.
+
+from flask_securest.acl_handlers.abstract_acl_handler import AbstractACLHandler
+
+SECURITY_CTX_USER = 'user'
+
+
+class CloudifyACLHandler(AbstractACLHandler):
```
dankilman: I assume this is the "Default" or "Simple" ACL handler so I think it should be called as such. No need to prefix with "Cloudify" (also the module name)

rest-service/manager_rest/server.py
```diff
@@ -181,11 +188,85 @@ def create_logger(logger_name,
...
+                                                        auth_provider)
+
+    def register_role_loader(role_loader):
+        secure_app.app.logger.debug('registering role loader {0}'.
+                                    format(role_loader))
+        loader = utils.get_class_instance(role_loader.get(IMPLEMENTATION_KEY),
+                                          role_loader.get(PROPERTIES_KEY))
+        secure_app.set_role_loader(loader)
+
+    def register_authorization_provider(authorization_provider):
```
dankilman: apply simple algorithm to find impls?

```
impl: ...
properties:
  roles_loader:
     impl:
     properties:

```

rest-service/manager_rest/server.py
```diff
@@ -182,10 +186,77 @@ def create_logger(logger_name,
...
+            secure_app.app.logger.debug(
+                'registering authentication provider {0}'.format(provider))
+            secure_app.register_authentication_provider(
+                provider['name'], create_instance(provider))
+
+    def register_authorization_provider(authorization_provider):
+        secure_app.app.logger.debug('registering authorization provider {0}'
+                                    .format(authorization_provider))
+        # validate the roles_config.yaml file exists
+        roles_config_file_path = \
```
dankilman: just verifying that we know that currently the authorization provider is not really plugabble as we are explicitly setting a `roles_config_file_path` which is something very specific to the roles based authorization provider

noakup: True, and we need to discuss that. I just left it like this because of issues with a better implementation I am working on.

PR #331, opened 2015-11-03
Cfy 3682 add implementation for authrization based on roles

noakup:


dankilman:
please squash the two commits

PR #348, opened 2015-11-12
CFY-4067 Add --force flag to 'snapshots restore'.

xwombat:


workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -385,12 +385,61 @@ def _update_es_node(es_node):
             source['main_file_name'] = ''
 
 
+def _assert_clean_elasticsearch(es):
+    exception = NonRecoverableError('Manager is not clean')
```
idanmo: Perhaps create a method for raising this error where necessary?

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -385,12 +385,61 @@ def _update_es_node(es_node):
...
+        raise exception
+
+    # check storage
+    storage_scan = elasticsearch.helpers.scan(es, index=_STORAGE_INDEX_NAME)
+    storage_scan = _except_types(storage_scan, 'provider_context', 'snapshot')
+    if next(storage_scan, False):
+        raise exception
+
+
+def _check_conflicts(es, restored_data):
```
idanmo: Please write a short docstring about the conflicts semantics

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -571,13 +622,19 @@ def recreate_deployments_environments():
...
     mappings = {
         '3.3': _restore_snapshot_format_3_3,
         '3.2': _restore_snapshot_format_3_2
     }
 
     config = _DictToAttributes(config)
+
+    es = _create_es_client(config)
+
+    if not force:
```
idanmo: I think it would be a good idea to check whether the manager is clean anyway, and have a warning log regardless of whether force=true was passed or not. The force=true will skip the validation but the warning will still be there.

workflows/cloudify_system_workflows/snapshot.py
simon-bar: Do you mean something like:
def raise(msg):
    raise NonRecoverableError(msg)
?

idanmo: Yes - with a better name for the method. and perhaps write a more informative message that it is only allowed to restore a snapshot on a clean manager.

workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -385,12 +385,86 @@ def _update_es_node(es_node):
...
+    """
+    Check if manager ElasticSearch is clean and raise error (or just
+    log warning) if it isn't.
+
+    :param es: ElasticSearch proxy object
+    :param log_warning: instead raising error just log warning
+    """
+
+    def _raise_or_log():
+        message = ('Manager is not clean, snapshot is allowed to be '
```
idanmo: The message should be different for both cases:
1. force=true: "Forcing snapshot restoration on dirty manager"
2. default: "Snapshot restoration on on a dirty manager is not permitted"

Feel free to adjust the messages.

idanmo: Will be taken care of in a separate PR.

simon-bar: @xwombat please take care of this

PR #350, opened 2015-11-17
CFY-4057 optionally disable v1 insecure endpoints

ChenRoth:


rest-service/manager_rest/resources.py
```diff
@@ -424,6 +436,7 @@ class BlueprintsId(SecuredResource):
     )
     @exceptions_handled
     @marshal_with(responses.BlueprintState)
+    @insecure_rest_method
```
dankilman: is this by accident?

ChenRoth: yes, i'll remove it.. good catch

PR #356, opened 2015-11-22
CFY-4213 fix multi value filter for event context fields

ChenRoth:


tests/workflow_tests/test_events.py
```diff
@@ -63,17 +63,30 @@ def test_sorted_events(self):
...
+
+        # filter a subset of the deployments
+        expected_deployment_ids = deployment_ids[:2]
+        filters = {'deployment_id': expected_deployment_ids}
         events = self.client.events.list(**filters)
+
         self.assertGreater(len(events), 0, 'No events')
+
+        # count events for each requested deployment id
+        events_per_deployment = \
```
ran-z: how about instead:
deployments_with_events = {event['context']['deployment_id'] for event in events}
self.assertEquals(deployments_with_events, set(expected_deployment_ids))

seems shorter
also, if at some point we break the filter and somehow the third deployment also returns an event, the current code will break on line 83

tests/workflow_tests/test_events.py
```diff
@@ -63,17 +63,25 @@ def test_sorted_events(self):
...
+        filters = {'deployment_id': expected_deployment_ids}
         events = self.client.events.list(**filters)
+
         self.assertGreater(len(events), 0, 'No events')
-        for event in events:
-            self.assertEqual(event['context']['deployment_id'],
-                             deployment_id,
-                             'Expected only events related to'
-                             ' deployment id {0}'.format(deployment_id))
+        deployments_with_events = \
```
ran-z: rather than doing set([x for x in y]) you can just do set comprehension using { x for x in y }

tests/workflow_tests/test_events.py
```diff
@@ -63,17 +63,25 @@ def test_sorted_events(self):
...
+            set([event['context']['deployment_id'] for event in events])
+        self.assertEquals(sorted(deployments_with_events),
+                          sorted(expected_deployment_ids),
+                          'Expected only events of deployment ids {0}, '
```
ran-z: error message assumes there might be additional deployment ids, though some might be missing out rather than have additional ones
i'd change the phrasing

PR #357, opened 2015-11-23
CFY-4236 current date as logstash index in es during snapshot restore

konradxyz:


workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -477,24 +477,13 @@ def logstash_to_cloudify_events():
...
-        return 'logstash-' + d.strftime('%Y.%m.%d')
-
-    def get_event_date(e):
-        return e['_source']['@timestamp']
-
     def cloudify_events_to_logstash():
-        event_indices = []
+        d = datetime.now()
+        index = 'logstash-{0}'.format(d.strftime('%Y.%m.%d'))
+        if not es.indices.exists(index=index):
```
simon-bar: This is probably not needed (?). The index should get created automatically I think..

nir0s:
LGTM

PR #358, opened 2015-11-23
CFY-4022 Merge plugins from snapshots with the existing ones instea…

pawelcodilime:
…d of replacing.

### DIFF
--- a/workflows/cloudify_system_workflows/snapshot.py
+++ b/workflows/cloudify_system_workflows/snapshot.py
@@ -133,13 +133,21 @@ def _copy_data(archive_root, config, to_archive=True):
         if not os.path.exists(p1):
             continue
 
-        # copy data (and override if target exists)
+        # copy data
         if os.path.isfile(p1):
             shutil.copy(p1, p2)
         else:
-            if os.path.exists(p2):
-                shutil.rmtree(p2, ignore_errors=True)
-            shutil.copytree(p1, p2)
+            if not os.path.exists(p2):
+                os.makedirs(p2)
+
+            for item in os.listdir(p1):
+                s = os.path.join(p1, item)
+                d = os.path.join(p2, item)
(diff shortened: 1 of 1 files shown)


workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -111,14 +111,16 @@ def _copy_data(archive_root, config, to_archive=True):
...
+    # absolute then should point to proper data in manager/snapshot archive.
+    # The last element denotes whether the directory should be overwritten.
     data_to_copy = [
-        (config.file_server_blueprints_folder, 'blueprints'),
-        (config.file_server_uploaded_blueprints_folder, 'uploaded-blueprints'),
-        (config.file_server_uploaded_plugins_folder, 'plugins')
+        (config.file_server_blueprints_folder, 'blueprints', True),
+        (config.file_server_uploaded_blueprints_folder, 'uploaded-blueprints',
+         True),
+        (config.file_server_uploaded_plugins_folder, 'plugins', False)
```
idanmo: Isn't it true for blueprints and uploaded blueprints as well?
If my manager contains blueprints, it will just erase them if I use force... same issue.. right?

simon-bar: Hmm in this case the True/False flag is not needed at all. We should simply always add things incrementally when --force flag is supplied.

idanmo: Sounds like it.

simon-bar: Moreover, we should probably use:

''' python
os.system("cp -r src/* target")

Because this shutil thing explodes if you want to copy a directory onto an existing one. This will result in a failure when restoring a snapshot with plugins on a manager on which we have just created this snapshot..

PR #359, opened 2015-11-24
CFY-4224 Restrict checking whether the manager is clean only to check…

pawelcodilime:
…ing whether there are any blueprints.

### DIFF
--- a/workflows/cloudify_system_workflows/snapshot.py
+++ b/workflows/cloudify_system_workflows/snapshot.py
@@ -385,31 +385,25 @@ def _update_es_node(es_node):
             source['main_file_name'] = ''
 
 
-def _assert_clean_elasticsearch(es, log_warning=False):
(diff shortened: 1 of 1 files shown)


workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -385,31 +385,25 @@ def _update_es_node(es_node):
...
-    # check storage
-    storage_scan = elasticsearch.helpers.scan(es, index=_STORAGE_INDEX_NAME)
-    storage_scan = _except_types(storage_scan,
-                                 'provider_context',
-                                 'snapshot',
-                                 'execution')
-    if next(storage_scan, False):
-        _raise_or_log()
+            raise NonRecoverableError(
+                "Snapshot restoration on on a dirty manager is not permitted.")
```
idanmo: ..on on..

PR #364, opened 2015-12-01
CFY-4314: Recreating deployment environments only for deployments tha…

simon-bar:
…t are being restored


workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -422,7 +422,7 @@ def _check_conflicts(es, restored_data):
     """
 
     old_data = elasticsearch.helpers.scan(es, index=_STORAGE_INDEX_NAME,
-                                          doc_type='blueprint, deployment')
+                                          doc_type='blueprint,deployment')
```
simon-bar: To the reviewer: This cosmetic change fixes our detection of deployment IDs conflicts

PR #365, opened 2015-12-04
CFY-4381 changed format of broker data stored in snapshot

konradxyz:


pleszcz:
@dankilman Please remote it. It is included in #375

PR #367, opened 2015-12-13
CFY-4287 ignore failures on botched deployment deletion

ChenRoth:


workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -147,27 +148,34 @@ def delete(ctx, **kwargs):
...
+        ctx.send_event('Stopping deployment operations worker'),
+        tasks['stop_operations_worker'],
+        ctx.send_event('Deleting deployment operations worker'),
+        tasks['delete_operations_worker'],
+        ctx.send_event('Stopping deployment workflows worker'),
+        tasks['stop_workflows_worker'],
+        ctx.send_event('Deleting deployment workflows worker'),
+        tasks['delete_workflows_worker'],
         ctx.send_event('Stopping deployment policy engine core'),
         ctx.execute_task('riemann_controller.tasks.delete'))
```
ran-z: forgot to use the task you created from before..

workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -147,27 +148,34 @@ def delete(ctx, **kwargs):
...
-        ctx.execute_task(
+        'delete_workflows_worker': ctx.execute_task(
             task_name='cloudify_agent.installer.operations.delete',
-            kwargs=WORKFLOWS_WORKER_PAYLOAD))
+            kwargs=WORKFLOWS_WORKER_PAYLOAD),
+        'stop_deployment_policy_engine_core':
+            ctx.execute_task('riemann_controller.tasks.delete')
+    }
+
+    # quick fix for unblocking botched deployments in 3.3.x
```
ran-z: i'd say the comment should have the word "delete"/"deletion" somewhere in there
it sounds like the deployment's stuck, but it's the deletion that fails

tests/mock_plugins/cloudify_agent/installer/operations.py
```diff
@@ -71,16 +71,19 @@ def stop(cloudify_agent=None, **_):
 
 @operation
 def delete(cloudify_agent=None, **_):
+
     installer = get_backend(cloudify_agent)
     installer.delete()
     worker_name = installer.agent_name
     with update_storage(ctx) as data:
+        if 'raise_exception_on_delete' in data:
+            raise Exception("Exception thrown intentionally")
```
ran-z: "exception thrown" --> "error raised"
you damn java addict
:p

tests/mock_plugins/cloudify_agent/installer/operations.py
```diff
@@ -71,16 +71,19 @@ def stop(cloudify_agent=None, **_):
...
         data[worker_name] = data.get(worker_name, {})
         data[worker_name]['states'] = data[worker_name].get('states', [])
         data[worker_name]['states'].append('deleted')
 
-    if ctx.type == context.NODE_INSTANCE:
-        del ctx.instance.runtime_properties['cloudify_agent']
+    # if ctx.type == context.NODE_INSTANCE:
```
ran-z: why is this commented out?

tests/workflow_tests/test_deployment_workflows.py
```diff
@@ -82,3 +85,44 @@ def test_get_workflow_parameters(self):
...
+    def test_delete_botched_deployment(self):
+
+        from testenv import testenv_instance
+        storage_file_path = os.path.join(
+            testenv_instance.plugins_storage_dir,
+            'agent.json'
+        )
+
+        dsl_path = resource('dsl/botched_deployment.yaml')
+        _id = uuid.uuid1()
```
ran-z: we use uuid4 usually - to be honest, i have no idea why. i just read about it for the first time, doesn't really sound like it should matter. just mentioning in case you'd like to be consistent :p

tests/workflow_tests/test_deployment_workflows.py
```diff
@@ -82,3 +85,44 @@ def test_get_workflow_parameters(self):
...
+        self.client.blueprints.upload(dsl_path, blueprint_id)
+        self.client.deployments.create(blueprint_id, deployment_id)
+
+        execution = self.client.executions.list(
+            deployment_id,
+            True,
+            workflow_id='create_deployment_environment'
+        )[0]
+
+        # cancel deployment creation and wait for it to finish
```
ran-z: so i'm not really sure i understood - is this test checking for the deletion of a cancelled deployment-creation workflow, or a failed one? or is this time-based?

tests/resources/dsl/botched_deployment.yaml
```diff
@@ -0,0 +1,11 @@
+tosca_definitions_version: cloudify_dsl_1_2
```
ran-z: i'm thinking there must have already been a blueprint identical to this, isnt there?
if so i'd probably use that one and just have the addition in the code for the flag to raise an error from the operation
not sure the duplication of the yaml is needed

tests/mock_plugins/cloudify_agent/installer/operations.py
ChenRoth: shouldn't be, maybe a mistake with the keyboard

tests/workflow_tests/test_deployment_workflows.py
ChenRoth: dan's code used uuid1.. i'll change to uuid4 anyway

tests/workflow_tests/test_deployment_workflows.py
ChenRoth: the test also cancels the deployment creation to botch it completely
if this is unnecessary, i'll remove it

tests/resources/dsl/botched_deployment.yaml
ChenRoth: on previous tasks, i was advised to create a blueprint for each scenario, so they'll be independent of each other. imo, i can use basic.yaml here it doesn't matter

tests/workflow_tests/test_deployment_workflows.py
ran-z: nah, my mistake then, can keep as is

tests/workflow_tests/test_deployment_workflows.py
ran-z: I think I'd have the test do just one thing so it's clearer what it's doing
maybe I'd have a second test which checks the other scenario as well, and refactor so both tests use most of the same code

tests/workflow_tests/test_deployment_workflows.py
```diff
@@ -82,3 +85,37 @@ def test_get_workflow_parameters(self):
...
+            deployment_id: {'raise_exception_on_delete': True}
+        }
+        with open(storage_file_path, 'w') as f:
+            json.dump(data, f)
+
+        self.client.blueprints.upload(dsl_path, blueprint_id)
+        self.client.deployments.create(blueprint_id, deployment_id)
+
+        # wait for creation to finish
+        from time import sleep
```
ran-z: why use sleep and not check the execution status instead?
i assume that before you used sleep to wait for execution to cancel but it doesnt seem needed now unless im missing something

ChenRoth: you have to use sleep cause the creation has to finish before you can delete it
i'll use the spin lock ```wait_for_execution_to_end```

PR #370, opened 2015-12-14
CFY-4550 test insecure endpoints is enabled by default

ChenRoth:


rest-service/manager_rest/test/test_list_v1.py
```diff
@@ -57,6 +57,12 @@ def instance(*args, **kwargs):
             # restore original function
             config.instance = original_instance_func
 
+    def test_insecure_endpoints_enabled_by_default(self):
+        try:
+            self.client.events.get(execution_id='111')
+        except CloudifyClientError, e:
+            self.assertNotEquals(405, e.status_code)
```
ran-z: i'd also assert the code is of the 200 family, to make sure we didn't get any 404, 400 or 50x before the 405 could even get launched or something like that

ChenRoth: but you do get 500, because elasticsearch doesn't exist in these tests

ran-z: right.. well, i guess its ok then, but not really ideal, as theoretically there could have been an 400 or 50x error prior to the 405 validation check which would make this test pass anyway

PR #371, opened 2015-12-15
CFY-4212 escape reserved characters in elasticsearch

ChenRoth:


rest-service/manager_rest/manager_elasticsearch.py
```diff
@@ -52,12 +54,20 @@ def build_request_body(filters=None, pagination=None, skip_size=False,
                         and values are the range limits of that field
         :return: An elasticsearch Query DSL body.
         """
+
+        def _escape_reserved_es_chars(val):
+            reserved_chars_regex = \
```
ran-z: i'd probably have this as a module global constant

rest-service/manager_rest/manager_elasticsearch.py
```diff
@@ -52,12 +54,20 @@ def build_request_body(filters=None, pagination=None, skip_size=False,
...
+                '([\(\)\{\}\+\-\=\>\<\!\[\]\^\"\~\*\?\:\\/]|&&|\|\|\s)'
+            return re.sub(reserved_chars_regex, r'\\\1', val)
+
         def _build_query_match_condition(k, val_list):
+            val_list = \
+                [_escape_reserved_es_chars(val) for val in val_list]
```
ran-z: how come here we only escape the value but not the key, unlike over at the wildcard condition build?

tests/workflow_tests/test_events.py
```diff
@@ -92,6 +92,10 @@ def test_paginated_events(self):
         self.assertEquals(pagination_info.offset, offset)
         self.assertEquals(pagination_info.size, size)
 
+    def test_query_with_reserved_characters(self):
```
ran-z: maybe also add one more test which checks more than the request merely not crashing, but rather actually try to retrieve a real event which contains one of these characters?

rest-service/manager_rest/manager_elasticsearch.py
ChenRoth: in wildcard, the key is part of the user query, rather than part of the structured query

ran-z: got it

rest-service/manager_rest/manager_elasticsearch.py
```diff
@@ -53,6 +57,12 @@ def build_request_body(filters=None, pagination=None, skip_size=False,
                         and values are the range limits of that field
         :return: An elasticsearch Query DSL body.
         """
+        def _escape_reserved_es_chars(val):
+            return re.sub(RESERVED_CHARS_REGEX, r'\\\1', val)
+
+        def _omit_reserved_es_chars(val):
+            return re.sub(RESERVED_CHARS_REGEX, r'', val)
```
ran-z: should replace with space, not empty char

tests/workflow_tests/test_events.py
```diff
@@ -92,6 +93,27 @@ def test_paginated_events(self):
...
+        message_without_reserved_chars = \
+            re.sub(RESERVED_CHARS_REGEX, '', message)
+
+        keywords = message_without_reserved_chars.lower().split()
+
+        # assert some search results contain requested message
+        for event in searched_events:
+            text = event['message']['text'].lower()
+            self.assertTrue(all([keyword in text for keyword in keywords]))
+
```
ran-z: need another test which checks partial query, e.g. "dum"

PR #373, opened 2015-12-15
CFY-4066 catch TransportError when running events list (3.3.1)

dankilman:


rest-service/manager_rest/resources_v2.py
```diff
@@ -21,12 +21,14 @@
...
 from collections import OrderedDict
 
-from flask_securest.rest_security import SecuredResource
 
+import elasticsearch
 from flask_restful_swagger import swagger
 from flask import request
 from flask.ext.restful import marshal
 
+from flask_securest.rest_security import SecuredResource
```
ran-z: why blank lines both above and below? either we treat it as a 3rd party or as a cloudify library, no?

rest-service/manager_rest/resources_v2.py
```diff
@@ -960,9 +962,18 @@ def list_events(self, query, include=None):
...
+        try:
+            return es.search(index=self._set_index_name(),
+                             body=query,
+                             _source=include or True)
+        except elasticsearch.TransportError as e:
+            # This happens when the events index is in the process
+            # of being created.
+            if (e.status_code == 503 and
+                    'SearchPhaseExecutionException' in e.error):
+                return {'hits': {'hits': [], 'total': 0}}
```
ran-z: for the record i dont really like this ("hits" etc.) being here - totally not the place for it, but seeing as how the whole implementation of _search doesnt really belong in this module we might as well :/

rest-service/manager_rest/resources_v2.py
dankilman: That's a matter of taste. Although not entirely standard, for "3rd party" library imports that actually derive from Cloudify (e.g. commons imports rest client) I usually have a fourth section that comes between the actual 3rd party and the in package imports.

It is up to debate whether we treat flask_securest as external or internal

rest-service/manager_rest/resources_v2.py
dankilman: I would go further and say that the implementation of `list_events` is very leaky as well. Anyway, I wish to avoid the refactoring such change will involve, at the moment.

PR #404, opened 2016-01-11
Added better error messages for failed snapshots that are tried to do…

xwombat:
…wnload/restore

### DIFF
--- a/rest-service/manager_rest/blueprints_manager.py
+++ b/rest-service/manager_rest/blueprints_manager.py
@@ -180,7 +180,12 @@ class BlueprintsManager(object):
         return execution
 
     def restore_snapshot(self, snapshot_id, recreate_deployments_envs, force):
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/blueprints_manager.py
```diff
@@ -180,7 +180,12 @@ def create_snapshot(self, snapshot_id,
         return execution
 
     def restore_snapshot(self, snapshot_id, recreate_deployments_envs, force):
-        self.get_snapshot(snapshot_id)  # Throws error if no snapshot found
+        # Throws error if no snapshot found
+        snap = self.get_snapshot(snapshot_id)
+        if snap.status == models.Snapshot.FAILED:
+            raise manager_exceptions.ManagerException(
```
ran-z: I think so far we've been consistent in not raising a "ManagerException" directly but rather a subclass instance
im thinking maybe one should be created for these two exceptions as well

PR #409, opened 2016-02-08
CFY-4889

gilzellner:


nir0s:
I think there's a reason it's in /tmp. I think we read that log. Suggest consulting with Dan.

gilzellner:
the log has moved to /var/log/cloudify/riemann.log sometime ago, and this only shows up during error messages when something else is wrong with the deployment.

PR #431, opened 2016-03-25
CFY-5082 Creating Riemann core only if policies are actually used in …

pleszcz:
…the blueprint

### DIFF
--- a/tests/resources/dsl/without_policies.yaml
+++ b/tests/resources/dsl/without_policies.yaml
@@ -1,0 +1,26 @@
+# this blueprint is like with_policies1, but without the policies
+# - so that we have two simple blueprints that only differ by the policies
+
+tosca_definitions_version: cloudify_dsl_1_2
+
+imports:
+    - cloudify/types/types.yaml
+    - plugins/testmockoperations.yaml
+    - plugins/mock_workflows.yaml
+    - policies/nop.yaml
+
+node_templates:
+    node:
+        type: cloudify.nodes.Root
+        interfaces:
+            test:
+                op1: testmockoperations.testmockoperations.tasks.mock_operation_from_custom_workflow
+
+workflows:
+    some_workflow:
+        mapping: mock_workflows.mock_workflows.workflows.test_policies_1
+        parameters:
+            key:
+                description: a key to write
+            value:
+                description: a value to write
--- a/workflows/cloudify_system_workflows/deployment_environment.py
+++ b/workflows/cloudify_system_workflows/deployment_environment.py
@@ -23,6 +23,23 @@ from cloudify.workflows import tasks as workflow_tasks
 from cloudify.workflows import workflow_context
 
 
+def _should_create_policy_engine_core(policy_configuration):
+    """Examine the policy_configuration and decide to start a riemann core
+    """
+    groups = policy_configuration['groups']
+    policy_types = policy_configuration['policy_types']
+
+    if not groups or not policy_types:
+        return False
+
+    for group in groups.values():
+        for policy_config in group.get('policies', {}).values():
+            if policy_config['type'] in policy_types:
+                return True
+
+    return False
+
+
 def generate_create_dep_tasks_graph(ctx,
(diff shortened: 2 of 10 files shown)


tehasdf:
Updated the commit to check ['groups'], ['policy_triggers'], ['policy_types'];
also send a message when we decide to skip creating the core.

tests/resources/dsl/without_policies.yaml
```diff
@@ -0,0 +1,26 @@
+# this blueprint is like with_policies1, but without the policies
+# - so that we have two simply blueprints that only differ by the policies
```
ran-z: "simple"? :)

ill also use this comment to also mention that the CI tests are failing

tehasdf: thanks!
CI somehow managed to run an older version of this branch - the failing assert isn't even in the code, lol. Probably because of the merge conflict. I'll rebase it tomorrow.

workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -38,10 +38,17 @@ def generate_create_dep_tasks_graph(ctx,
             ctx.execute_task('cloudify_agent.operations.install_plugins',
                              kwargs={'plugins': plugins_to_install}))
 
-    sequence.add(
-        ctx.send_event('Starting deployment policy engine core'),
-        ctx.execute_task('riemann_controller.tasks.create',
-                         kwargs=policy_configuration or {}))
+    if (policy_configuration['groups'] and
```
ran-z: So here's the problem - what you wrote here works fine right now, however in the very near future we're going to introduce the notion of "scaling groups" in the DSL, and, well, long story short, there may be a blueprint which only has groups which use the specific policy type for scaling groups - in which case, we don't need a riemann core for the deployment after all.

rather than have the type hardcoded here, I think the condition here should be as follows: if there is *any* group which is mapped to a policy type *which is defined in policy_types*, we need to install riemann, otherwise don't.
i know this sounds a bit confusing right now, but basically this hardcoded type i'm referencing is a policy type which is actually not going to be defined in the "policy_types" dictionary - so that's why only if there's a group with an actual "custom" type, does riemann become relevant.

tests/resources/dsl/without_policies.yaml
```diff
@@ -0,0 +1,26 @@
+# this blueprint is like with_policies1, but without the policies
+# - so that we have two simplw blueprints that only differ by the policies
```
dankilman: two simple

workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -39,10 +56,15 @@ def generate_create_dep_tasks_graph(ctx,
...
-        ctx.send_event('Starting deployment policy engine core'),
-        ctx.execute_task('riemann_controller.tasks.create',
-                         kwargs=policy_configuration or {}))
+    if _should_create_policy_engine_core(policy_configuration):
+        sequence.add(
+            ctx.send_event('Starting deployment policy engine core'),
+            ctx.execute_task('riemann_controller.tasks.create',
+                             kwargs=policy_configuration))
+    else:
+        sequence.add(
```
dankilman: maybe update the delete deployment part so that is says something like
`Stopping deployment policy engine core (if applicable)`
or some other wording

workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -23,6 +23,23 @@
...
+    """Examine the policy_configuration and decide to start a riemann core
+    """
+    groups = policy_configuration['groups']
+    policy_types = policy_configuration['policy_types']
+
+    if not groups or not policy_types:
+        return False
+
+    for group in groups.values():
+        for policy_config in group['policies'].values():
```
dankilman: a group may have no policies defined (not true yet, but will be true soon) so do something like
`group.get('policies', {})` or similar

PR #432, opened 2016-03-31
Added CORS headers to response

01000101:
I didn't get a response in chat if anyone was against adding CORS support to the REST service, so here's the proposed patch.  This will allow third party web-based applications (ie. JavaScript) to make calls to the REST service.


geokala:

PR #442, opened 2016-04-22
CFY-5225 separate db snapshot impl  from plugins-common impl 

adamlavie:


workflows/cloudify_system_workflows/snapshot.py
```diff
@@ -593,10 +364,18 @@ def _restore_snapshot(config, tempdir, metadata):
     # elasticsearch
     es = _create_es_client(config)
```
dankilman: check if es creation is actually required here

geokala:

PR #449, opened 2016-05-04
Cfy 5218 manager elasticsearch

tehasdf:
Both "ES Storage Manager" and "Manager Elasticsearch" have a ES
connection. Share the connection between them, and store it on
the app.


tehasdf:
Decided to not share the connection between ESStorageManager and ManagerElasticsearch - unfortunately, ESStorageManager needs to work outside of request context (because it's used in tests - just one time, though), so we can't have it store the conn on current_app.

PR #471, opened 2016-05-23
CFY-5286 Don't hide 404 errors

tehasdf:
A 404 error could result in a 500, which obscured it, leading to an
unhelpful error. Make sure the maintenance mode handler doesn't
break on un-routed requests.

### DIFF
--- a/rest-service/manager_rest/test/test_app.py
+++ b/rest-service/manager_rest/test/test_app.py
@@ -1,0 +1,33 @@
+#########
+# Copyright (c) 2016 GigaSpaces Technologies Ltd. All rights reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/test/test_app.py
```diff
@@ -0,0 +1,33 @@
...
+"""Test the basic HTTP interface, app-wide error handling
+"""
+
+from nose.plugins.attrib import attr
+
+from manager_rest.test import base_test
+
+
+@attr(client_min_version=1, client_max_version=base_test.LATEST_API_VERSION)
+class AppTestCase(base_test.BaseServerTestCase):
```
tehasdf: This didn't really fit anywhere else :( I guess just making a new file is fine, and in the future we can test other app-wide things like errorhandlers etc here.

PR #474, opened 2016-05-24
CFY-4936 edited error text

nirbir:


rest-service/manager_rest/maintenance.py
```diff
@@ -174,7 +177,7 @@ def _maintenance_mode_error():
     traceback.print_exc(file=s_traceback)
 
     response = jsonify(
-            {"message": "Request rejected since maintenance mode is active",
+            {"message": ERROR_MESSAGE,
```
nir0s: consolidate `_maintenance_mode_error` and `_activating_maintenance_mode_error`

rest-service/manager_rest/maintenance.py
```diff
@@ -38,6 +38,9 @@
                      'version']
 LOCAL_ADDRESS = '127.0.0.1'
 
+ERROR_MESSAGE = 'Your request was rejected since Cloudify Manager ' \
```
nir0s: move into function

PR #479, opened 2016-05-29
Cfy 5344 support skipping install/uninstall in update rebase

max-orlov:


rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -329,6 +342,11 @@ def _execute_update_workflow(self,
                 extract_ids(removed_instances.get(NODE_MOD_TYPES.RELATED))
         }
 
+        if not workflow_id:
+            # Whether execute or not install and uninstall
```
max-orlov: Whether or not execute install or uninstall

PR #502, opened 2016-06-18
CFY 4929 deployment update

ran-z:


tests/workflow_tests/test_deployment_update.py
```diff
@@ -2004,3 +2012,104 @@ def test_remove_deployment(self):
         self.assertEquals(len(deployment_update_list), len(undel_depups_ids))
         for i in deployment_update_list.items:
             self.assertIn(i['id'], undel_depups_ids)
+
+
+class TestDeploymentUpdateMisc(DeploymentUpdateBase):
```
max-orlov: So there's this class: TestDeploymentUpdateMixedOperations...

ran-z: I figured mixed operations are about mixing of specific updates, e.g. node and relationship update; this class is about misc update scenarios. do you still think they should be unified?

tests/testenv/es_schema_creator.py
```diff
@@ -32,10 +32,22 @@
 DEPLOYMENT_SCHEMA = {
     'deployment': {
         'properties': {
-            'plan': {
```
max-orlov: so the plan was never really there to put in storage?

ran-z: at the dark old times, we used to have the deployment plan under the deployment object; this was later removed - but apparently it was forgotten in this module :)

tests/workflow_tests/test_deployment_update.py
max-orlov: well, maybe the name is misleading, but the mixoperations basically held any test which is more than one step/skipping install/executing custom workflow etc...

it sounds like it's a good idea to extract any non step related test to your new class...

tests/testenv/es_schema_creator.py
max-orlov: cool beans

tests/workflow_tests/test_deployment_update.py
ran-z: ok, i moved all but two tests to my new class; kinda difficult seeing that on github comparison view but you can try and just view the file itself

PR #514, opened 2016-06-29
CFY-5643 404 when getting node-instances for nonexistent deployment

tehasdf:
This makes node-instances endpoint's behaviour consistent with what
others (executions) do. Also note that the CLI expects the REST service
to throw 404 in that case.


geokala:

PR #523, opened 2016-07-12
CFY-5598 Pass REST credentials in the workflow context

tehasdf:


tehasdf:
This makes sure that the REST credentials (username&password, or token) are in the workflow context so that the agents (and management worker) can use them.

rest-service/manager_rest/workflow_client.py
```diff
@@ -79,6 +99,8 @@ def execute_system_workflow(wf_id, task_id, task_mapping,
             context['deployment_id'] = deployment.id
 
         execution_parameters['__cloudify_context'] = context
+        execution_parameters['__cloudify_context'].update(
```
dankilman: convert to `classmethod` and use `cls`

rest-service/manager_rest/workflow_client.py
```diff
@@ -52,6 +68,10 @@ def execute_workflow(name,
                 'package_version': plugin.get('package_version')
             }
         }
+
+        execution_parameters['__cloudify_context'].update(
+            WorkflowClient._get_rest_credentials())
```
dankilman: convert to `classmethod` and use `cls`

rest-service/manager_rest/workflow_client.py
```diff
@@ -15,11 +15,26 @@
...
 
 
 class WorkflowClient(object):
+    @staticmethod
+    def _get_rest_credentials():
+        if not current_app.config.get(SECURED_MODE):
+            return {}
+
+        if hasattr(current_app, 'auth_token_generator'):
+            return {'rest_token':
```
dankilman: what about token expiration? if it is configurable we are at risk of the token expiring before the workflow execution ended

tehasdf: we decided to increase the default expiration to 24h as this should cover all uses

rest-service/manager_rest/config.py
```diff
@@ -50,6 +50,7 @@ def __init__(self):
         self._security_authentication_providers = []
         self._security_authorization_provider = None
         self._insecure_endpoints_disabled = False
+        self._security_agent_rest_credentials = None
```
dankilman: having one property with username/password is inconsistent with other properties that are top level.
while it makes sense generally, i don't think this is how it should be implement to keep up with current implementation.

rest-service/manager_rest/workflow_client.py
tehasdf: in cloudify-cosmo/cloudify-manager-blueprints#365

rest-service/manager_rest/config.py
tehasdf: OK. Same for the node property? https://github.com/cloudify-cosmo/cloudify-manager-blueprints/pull/365/files#diff-9a67a043795bd3da7b6364194f907257R157

rest-service/manager_rest/workflow_client.py
tehasdf: converted

rest-service/manager_rest/workflow_client.py
tehasdf: in the future we could change flask-securest's API to allow non-expiring, salted (by eg. execution id) tokens instead

rest-service/manager_rest/config.py
dankilman: yes

PR #536, opened 2016-07-18
CFY-5464 sort timestamp test

nirbir:


rest-service/manager_rest/test/test_events_v2.py
```diff
@@ -135,6 +136,13 @@ def _mock_es_search_delete(self, *args, **kwargs):
         }
 
         self.assertIn('body', kwargs)
+        try:
+            list_to_sort = \
```
dankilman: can `expected_list` be renamed to `expected_must_list` and `list_to_sort` to `actual_must_list`?

PR #537, opened 2016-07-18
CFY-5157 Install agents script: dont pass old celery config

tehasdf:


resources/rest-service/cloudify/install_agent.py
```diff
@@ -46,6 +46,12 @@ def __init__(self, logger):
     def run(self, command, execution_env=None):
         self.logger.debug('run: {0}'.format(command))
         command_env = os.environ.copy()
+
+        # we're running on the old agent - don't pass our celery config to the
+        # new one
+        command_env.pop('CELERY_CONFIG_MODULE')
```
tehasdf: we can't just pass an empty env because we do need (at least) the PATH

PR #541, opened 2016-07-21
CFY-5135 Install agent script: SSL fileserver communication

tehasdf:


dankilman:
The logic itself is indeed required but the solution should also support windows based hosts.
This is why urllib is used instead of forking out

resources/rest-service/cloudify/install_agent.py
```diff
@@ -159,6 +173,13 @@ def install(self):
         finally:
             self.runner.rm_dir(path)
 
+    def _get_certificate_file(self):
+        if not self.cloudify_agent.get('verify_rest_certificate'):
+            return None
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            f.write(self.cloudify_agent['rest_cert_content'])
```
dankilman: will `rest_cert_content` always exist?

resources/rest-service/cloudify/install_agent.py
```diff
@@ -68,15 +70,24 @@ def run(self, command, execution_env=None):
             self.logger.error(out)
             self.logger.error('Stderr:')
             self.logger.error(err)
-            raise Exception()
+            raise CommandExecutionException(command, err, out, p.returncode)
 
-    def download(self, url, destination=None):
+    def download(self, url, destination=None, verify_certificate=True,
```
dankilman: you are currently ignoring `destination` if it is provided

resources/rest-service/cloudify/install_agent.py
dankilman: perhaps create the file under the temp dir `path` that gets created and deleted already.
currently this file is created but never gets deleted.

noakup: the file might already exit, maybe check for it?

resources/rest-service/cloudify/install_agent.py
tehasdf: fixed

resources/rest-service/cloudify/install_agent.py
```diff
@@ -159,6 +181,17 @@ def install(self):
         finally:
             self.runner.rm_dir(path)
 
+    def _get_certificate_file(self, path=None):
+        if not self.cloudify_agent.get('verify_rest_certificate'):
+            return None
+
+        if self.cloudify_agent.get('local_rest_cert_file'):
```
tehasdf: TODO check

i think actually at this point we cna't use local_rest_cert_file yet, and instead we should've kept the previous implementation

resources/rest-service/cloudify/install_agent.py
```diff
@@ -68,14 +70,31 @@ def run(self, command, execution_env=None):
...
         self.logger.debug('Retrieving file from {0}'.format(url))
-        if destination is None:
-            fh_num, destination = tempfile.mkstemp()
-            os.close(fh_num)
-        urllib.urlretrieve(url, destination)
+
+        verify = False
+        if verify_certificate:
+            if certificate_file:
+                verify = certificate_file
```
ran-z: verify = certificate_file or True

resources/rest-service/cloudify/install_agent.py
```diff
@@ -68,14 +70,31 @@ def run(self, command, execution_env=None):
...
+        if destination:
+            destination_file = open(destination)
+        else:
+            destination_file = tempfile.NamedTemporaryFile(delete=False)
+            destination = destination_file.name
+
+        with destination_file as f:
+            for chunk in response.bytes_stream(8192):
+                f.write(chunk)
+            f.flush()
```
ran-z: shouldn't f be closed rather than flushed?

resources/rest-service/cloudify/install_agent.py
```diff
@@ -159,6 +181,19 @@ def install(self):
         finally:
             self.runner.rm_dir(path)
 
+    def _get_certificate_file(self, path=None):
+        if not self.cloudify_agent.get('verify_rest_certificate'):
+            return None
+
+        # we were told to verify the cert, but didn't get the cert to verify
```
ran-z: maybe im misunderstanding this, but i think the comment should be inside the if, not above it

resources/rest-service/cloudify/install_agent.py
tehasdf: the `with` statement closes it. `flush` isn't technically needed at this point but rest-client from where i copied this does it

PR #542, opened 2016-07-21
CFY-5639 Allow deleting events by Deployment ID

mcouthon:


rest-service/manager_rest/models.py
```diff
@@ -277,3 +277,18 @@ def __init__(self, **kwargs):
         self.excluded_wheels = kwargs['excluded_wheels']
         self.supported_py_versions = kwargs['supported_py_versions']
         self.uploaded_at = kwargs['uploaded_at']
+
+
+class Event(SerializableObject):
```
ChenRoth: event/logs docs may contain variable fields, and do not comply to a single schema
we'll have to consult and see if this solution is sufficient

mcouthon: I'm not really even using this class. It exists, but it isn't being returned anywhere. I just created it to adhere to how things work in the BlueprintManager class. I'm pretty sure I can leave it with the `id` field only, and it'll be good enough.

ChenRoth: ok ,  change to master.zip and merge

PR #543, opened 2016-07-21
CFY-5464 fixed snapshot restoration bug caused by a different timesta…

nirbir:
…mp format in pre 3.5.0 version. added sorting unit test


rest-service/manager_rest/utils.py
```diff
@@ -196,6 +196,7 @@ def plugin_installable_on_current_platform(plugin):
     ]))
 
 
-def get_formatted_timestamp():
+def get_formatted_timestamp(timestamp=None):
```
nirbir: remove

PR #561, opened 2016-08-10
CFY-5135 install_agent: unescape the cert before writing

tehasdf:


tehasdf:
see #562 for comments

PR #562, opened 2016-08-10
CFY-5135 install_agent: unescape the cert before writing

tehasdf:


resources/rest-service/cloudify/install_agent.py
```diff
@@ -187,7 +187,8 @@ def _get_certificate_file(self, path=None):
             return None
 
         with tempfile.NamedTemporaryFile(delete=False, dir=path) as f:
-            f.write(self.cloudify_agent['rest_cert_content'])
+            f.write(self.cloudify_agent['rest_cert_content']
+                    .decode('string-escape'))
```
tehasdf: this change mirrors the one made here: https://github.com/cloudify-cosmo/cloudify-agent/pull/146#discussion-diff-72792192

still not 100% sure why is it escaped, but it seems we'll just need to deal with it for now :)

tehasdf:
CFY-5831 is about getting rid of this, but requires some research

PR #568, opened 2016-08-16
CFY-5824 Add-postgresql-to-integration-tests-environment

dusking:


tests/dev-requirements.txt
```diff
@@ -15,3 +15,5 @@ pyzmq==15.1.0
 python-dateutil
 wagon
 elasticsearch==1.6.0
+flask_sqlalchemy
```
dusking: add version

tests/postgresql_configuration.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
dusking: move the db creation to the setup per test,
for example:  tests/testenv/__init__.py
include: create + cleanup

tests/testenv/__init__.py
```diff
@@ -319,6 +321,11 @@ def start_elasticsearch(self):
         self.elasticsearch_process = ElasticSearchProcess()
         self.elasticsearch_process.start()
 
+    def start_postgresql(self):
```
dusking: since systemd not exists in all system, don't use it. just tell the developer to make sure that he have running ps.....
no need for postgresql process object

PR #569, opened 2016-08-18
CFY-5824 Add-postgresql-to-integration-tests-environment

dusking:


tests/dev-requirements.txt
```diff
@@ -15,3 +15,8 @@ pyzmq==15.1.0
 python-dateutil
 wagon
 elasticsearch==1.6.0
+psycopg2==2.6.2
```
idanmo: Are you using all of these packages?
I couldn't find a place where webargs is used...

tests/testenv/processes/postgresql.py
```diff
@@ -0,0 +1,54 @@
...
+    def configure(self, db_name, username, password):
+        if not self.is_running():
+            logger.warning('postgresql service is not running')
+            return
+        logger.info('Configuring postgresql db..')
+        ps_config_source = 'tests/postgresql_configuration.sh'
+        ps_config_destination = '/tmp/postgresql_configuration.sh'
+        utils.copy(source=ps_config_source,
+                   destination=ps_config_destination)
+        utils.sudo('chmod +x {0}'.format(ps_config_destination))
```
idanmo: Do you really have to use sudo?
It means that test will have to run with sudo permissions, which is not good...

idanmo:
Please check my comments.

PR #574, opened 2016-09-01
CFY-5904 add pg to build env

dusking:


packaging/restservice/build.spec
```diff
@@ -36,6 +36,8 @@ sudo /tmp/env/bin/pip install wheel==0.24.0 && \
 %build
 %install
 
+os.environ['REST_SERVICE_BUILD'] = "True"
```
nir0s: Um.. python?

dusking: Hmmm.. rpm build doesn't support python?? well it should... :-)

rest-service/setup.py
```diff
@@ -13,10 +13,35 @@
...
+    'PyYAML==3.10',
+    'elasticsearch==1.6.0',
+    'celery==3.1.17',
+    'flask-securest==0.8',
+    'psutil==3.3.0',
+    'jsonpickle==0.9.2',
+    'wagon==0.3.2'
+]
+
+
```
nir0s: Please add a comment on why this is here

PR #601, opened 2016-09-28
CFY-5982 Set the password generated during bootstrap in environment

mcouthon:


tests/integration_tests/resources/dev_resource_urls.yaml
```diff
@@ -1,5 +1,5 @@
 cli_source_url: ''
-rest_service_source_url: ''
+rest_service_source_url: https://github.com/cloudify-cosmo/cloudify-manager/archive/CFY-5982-autogenerate-password.tar.gz
```
dankilman: should probably change

tests/integration_tests/tests/agentless_tests/test_deployment_nodes.py
```diff
@@ -153,6 +155,7 @@ def test_update_node_instance_runtime_properties(self):
         # Verifying the node no longer has any runtime properties
         self.assertEquals(0, len(node_instance.runtime_properties))
 
+    @skip('Deprecated since using sqlalchemy locking mechanism')
```
dankilman: why not simply remove?

dusking (approved):

tests/integration_tests/resources/dev_resource_urls.yaml
benson-budiman-cb: No code change. Nil UserFacingError on the success path is intentional: callers short-circuit only when it is non-nil (same contract as the protocol-quote maturity mapping). Forcing a non-nil empty Error would break that sentinel.

PR #609, opened 2016-10-14
CFY-6009 Add HA cluster endpoint

tehasdf:


tests/integration_tests/framework/docl.py
```diff
@@ -198,6 +198,7 @@ def run_manager(label=None, tag=None):
     _wait_for_services()
     logger.info(
         'Container start took {0} seconds'.format(time.time() - start))
+    return container_details
```
tehasdf: In `cloudify_premium`'s HA tests, we use this test framework's interface to docl to run an additional manager container (because testing a cluster necessarily involves running several machines). Returning the details here allows us to actually pass the container's IP around where needed.

PR #641, opened 2016-12-06
CFY-6162 set max results returned from api

dusking:


rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -293,16 +293,48 @@ def _paginate(query, pagination):
...
         else:
+            total = query.order_by(None).count()
+            SQLStorageManager._validate_returned_size(total)
             results = query.all()
             return results, len(results), 0, 0
 
+    @staticmethod
+    def _validate_pagination(pagination_size):
+        if pagination_size < 0:
+            raise manager_exceptions.IllegalActionError(
```
idanmo: I think the BadParametersError error is more appropriate here.

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -293,16 +293,48 @@ def _paginate(query, pagination):
...
+                'Invalid pagination size: {0}.'.format(
+                    pagination_size
+                )
+            )
+
+        if pagination_size > config.instance.max_results:
+            raise manager_exceptions.IllegalActionError(
```
idanmo: I think the BadParametersError error is more appropriate here.

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -293,16 +293,48 @@ def _paginate(query, pagination):
...
+                'Invalid pagination size: {0}. Max allowed: {1}'.format(
+                    pagination_size,
+                    config.instance.max_results
+                )
+            )
+
+    @staticmethod
+    def _validate_returned_size(size):
+        if size > config.instance.max_results:
```
idanmo: Why would you get more results than you requested?

rest-service/manager_rest/utils.py
```diff
@@ -222,7 +222,7 @@ def setup_flask_app(db, user_datastore, manager_ip='localhost', driver=''):
 def get_postgres_db_uri(manager_ip, driver):
     """Get a valid SQLA DB URI
     """
-    dialect = 'postgres+{0}'.format(driver) if driver else 'postgres'
+    dialect = 'postgresql+{0}'.format(driver) if driver else 'postgres'
```
idanmo: ?

dusking: this is a fix for our tests env, it should be postgresql, not postgres

rest-service/manager_rest/storage/storage_manager.py
dusking: _validate_returned_size is being used when you're not using pagination

idanmo: So why not setting pagination to 100 by default?

PR #645, opened 2016-12-08
CFY-6162 inforce max result

dusking:


rest-service/manager_rest/es_storage_manager.py
```diff
@@ -50,6 +51,15 @@
     'refresh': True
 }
 
+logging.basicConfig(filename='/var/log/cloudify/es_manager.log',
+                    level=logging.DEBUG,
+                    format='%(asctime)s - %(name)s - %(message)s')
+logger = logging.getLogger(__name__)
+logging.getLogger('manager-rest').setLevel(logging.ERROR)
```
idanmo: What is the current level of these loggers?
Will users be able to change the level from ERROR without changing code?

rest-service/manager_rest/manager_elasticsearch.py
```diff
@@ -104,12 +103,15 @@ def _build_wildcard_condition(k, v):
...
+            pagination_size = pagination.get('size',
+                                             config.instance().max_results)
+            ManagerElasticsearch._validate_pagination(pagination_size)
             if not skip_size:
-                body['size'] = pagination.get('size', DEFAULT_SEARCH_SIZE)
+                body['size'] = pagination_size
             if 'offset' in pagination:
                 body['from'] = pagination['offset']
         elif not skip_size:
-            body['size'] = DEFAULT_SEARCH_SIZE
```
idanmo: Validation should probably occur here if the user provided it..

rest-service/setup.py
```diff
@@ -38,7 +38,6 @@
         'requests==2.7.0',
         'gunicorn==18.0',
         'PyYAML==3.10',
-        'elasticsearch==1.6.0',
```
idanmo: You should probably leave this line..
In order to be more explicit and safe perhaps the modified client version should be changed to something like 1.6.0.1 (because elasticsearch is still required by this project).

The github dependency of the modified es client should probably also be added to dev-requirements.txt (for unit tests and development).

rest-service/manager_rest/manager_elasticsearch.py
dusking: if user provided size we'll be in "if pagination" scope. we'll get here if user didn't provided it..

PR #652, opened 2016-12-19
CFY-6219-add-min-memory-validation

dusking:


rest-service/manager_rest/manager_exceptions.py
```diff
@@ -24,6 +24,15 @@ def __init__(self, http_code, error_code, *args, **kwargs):
         self.error_code = error_code
 
 
+class InsufficientMemoryError(ManagerException):
+    INSUFFICIENT_MEMORY_ERROR_CODE = 'insufficient_memory_error'
+
+    def __init__(self, *args, **kwargs):
+        super(InsufficientMemoryError, self).__init__(
+            400, InsufficientMemoryError.INSUFFICIENT_MEMORY_ERROR_CODE,
```
idanmo: I think error code 503 would be more suitable here and 4xx are usually for client side errors.

PR #654, opened 2016-12-20
CFY-6145 Prevent deletion/modification of bootstrap admin

mcouthon:
And make its ID to always be 0


rest-service/manager_rest/storage/management_models.py
```diff
@@ -147,5 +150,9 @@ def is_admin(self):
     def is_suspended(self):
         return self.role == SUSPENDED_ROLE
 
+    @property
+    def is_bootstrap_admin(self):
+        return self.role == BOOTSTRAP_ADMIN_ID
```
dusking: id..

PR #665, opened 2017-01-01
CFY-6263 - Added all tenants flag to list all resources in tenant

adamlavie:


rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -121,7 +121,8 @@ def _filter_query(self, query, model_class, filters):
         of such values)
         :return: An SQLAlchemy AppenderQuery object
         """
-        query = self._add_tenant_filter(query, model_class)
+        if not all_tenants:
```
mcouthon: If not admin filter by all the user's tenants

rest-service/manager_rest/test/base_test.py
```diff
@@ -138,6 +140,15 @@ def create_client(self, headers=None):
 
         return client
 
+    @contextmanager
+    def client_using_tenant(self, client, tenant_name):
```
mcouthon: Maybe use `self.client` by default? :)

rest-service/manager_rest/test/security/test_base.py
```diff
@@ -40,6 +41,8 @@ def use_secured_client(self, headers=None, **kwargs):
 
     def get_secured_client(self, headers=None, **kwargs):
         headers = headers or create_auth_header(**kwargs)
+        if not headers.get(CLOUDIFY_TENANT_HEADER):
```
mcouthon: You can use setdefault :D

PR #668, opened 2017-01-04
CFY-6184 Add proper password + token handling

mcouthon:


rest-service/manager_rest/security/authentication.py
```diff
@@ -71,25 +106,13 @@ def _basic_http_authenticate(user, hashed_pass):
         :param user: A valid user object, or None
         :param hashed_pass: md5 hashed password, or None
         """
-        if not user or md5(user.password) != hashed_pass:
+        if not user or user.password != hashed_pass:
```
dusking: is it hashed_pass or encrypted_pass?

rest-service/manager_rest/security/user_handler.py
```diff
@@ -31,47 +31,28 @@ def user_loader(request):
...
-        user, hashed_pass = _get_user_from_token(token)
+    return request.headers.get(token_auth_header)
+
 
-    return user, hashed_pass
+def get_token_status(token):
+    return flask_get_token_status(
+            token=token,
+            serializer='remember_token',
+            max_age='LOGIN',
```
dusking: need to verify max_age.. do we have user login.. wouldn't it be simpler to validate the token ourselves?

rest-service/manager_rest/server.py
```diff
@@ -72,6 +72,11 @@ def _set_flask_security(self):
         # Make sure that it's possible to get users from the datastore
         # by username and not just by email (the default behavior)
         self.config['SECURITY_USER_IDENTITY_ATTRIBUTES'] = 'username, email'
+        self.config['SECURITY_PASSWORD_HASH'] = 'pbkdf2_sha256'
+        self.config['SECURITY_PASSWORD_SALT'] = 'abckjshd0-dsi;dlksP0980!*'
```
dusking: we have it also in the utils. should we have one var for it?

PR #669, opened 2017-01-04
CFY 6290. Split resources into modules

jcollado:
In this PR, the `manager_rest.rest.resources` module is split into a package with multiple modules that should be easier to maintain. The code itself hasn't been updated and the same imports as before this PR still work.


jcollado:
[Issue](https://cloudifysource.atlassian.net/browse/CFY-6290) created and package renamed to use the `v1` suffix.

jcollado:
Closing this PR to rename the branch and create a new one that takes advantage of the integration with JIRA

PR #676, opened 2017-01-09
CFY-6290. Split resources_v2 into submodules

jcollado:
In this PR, the long `resources_v2.py` module is converted into a package with submodules. The implementation stays the same.


jcollado:
@mcouthon In the case of imports I prefer to do so because it's easier to sort them in an editor when needed and also because when a dependency is added/removed the diff is clearer.

In any case, I have to admit that I haven't seen too many references about the subject except this one: 
https://dev.launchpad.net/PythonStyleGuide#Multiline_imports

PR #678, opened 2017-01-09
CFY-6284. Remove search endpoint implementation

jcollado:
In this PR, the implementation of the `search` endpoint is removed since it will no longer be supported.

The reason for this is that is an endpoint that supports sending a query directly to Elasticsearch, which is not going to be used as an events backend anymore.


rest-service/manager_rest/rest/resources_v1/search.py
```diff
@@ -43,10 +42,6 @@ class Search(SecuredResource):
...
-        """
-        Search using an Elasticsearch query
-        """
-        request_dict = get_json_and_verify_params()
-        return ManagerElasticsearch.search(
-            index='cloudify_storage',
-            body=request_dict)
+        """Search using an Elasticsearch query."""
+        raise manager_exceptions.MethodNotAllowedError(
+            'Event search using elasticsearch queries is not longer supported')
```
idanmo: "is not longer" -> "is no longer"

jcollado: Fixed in 79f7de8

jcollado:
The tests failed because there is a constant (`DEFAULT_SEARCH_SIZE`) that is now used in the `events` module. That can be fixed easily, but I have also realized that part of the `manager_elasticsearch` module will be used still to restore snapshots with elasticsearch data. Hence, it's not a good idea to completely remove the module. I'm going to re-add the parts that are going to be used and then merge the PR.

PR #680, opened 2017-01-10
CFY-6253 allow user to list tenants

adamlavie:


rest-service/manager_rest/test/base_test.py
```diff
@@ -102,6 +103,15 @@ class BaseServerTestCase(unittest.TestCase):
     def __init__(self, *args, **kwargs):
         super(BaseServerTestCase, self).__init__(*args, **kwargs)
 
+    def create_client_with_creds(self, username, password, tenant=None):
```
dusking: why you're adding this function here?

PR #686, opened 2017-01-17
CFY-6317 Protect default tenant

mcouthon:
- Make the default tenant ID constant (set to 0)
 - Prevent deletion of the default tenant
 - Streamline the creation of the default tenant
 - Make more use of the default tenant constants (instead of strings)

### DIFF
--- a/rest-service/manager_rest/flask_utils.py
+++ b/rest-service/manager_rest/flask_utils.py
@@ -1,0 +1,80 @@
+#########
+# Copyright (c) 2013 GigaSpaces Technologies Ltd. All rights reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+#  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  * See the License for the specific language governing permissions and
+#  * limitations under the License.
+
+from collections import namedtuple
+
+from flask import Flask
+from flask_security import Security
+
+from manager_rest.storage import user_datastore, db
+
+
+def setup_flask_app(manager_ip='localhost', driver=''):
+    """Setup a functioning flask app, when working outside the rest-service
+
+    :param manager_ip: The IP of the manager
+    :param driver: SQLA driver for postgres (e.g. pg8000)
+    :return: A Flask app
+    """
+    app = Flask(__name__)
+    db_uri = _get_postgres_db_uri(manager_ip, driver)
+    app.config['SQLALCHEMY_DATABASE_URI'] = db_uri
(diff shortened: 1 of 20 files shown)


rest-service/manager_rest/flask_utils.py
```diff
@@ -0,0 +1,80 @@
+#########
```
mcouthon: Moved it here to avoid circular dependency (was previously located in `manager_rest/utils.py`)

rest-service/manager_rest/storage/utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def _create_roles():
+    for role in constants.ALL_ROLES:
+        user_datastore.create_role(name=role)
+    return user_datastore.find_role(constants.ADMIN_ROLE)
+
+
+def _create_default_tenant():
+    default_tenant = Tenant(
```
mcouthon: Now creating the default tenant with a specific ID (like the BS admin)

rest-service/manager_rest/test/base_test.py
```diff
@@ -211,23 +213,19 @@ def _handle_flask_app_and_db(self, server):
         self._handle_default_db_config(server)
         self._setup_anonymous_user(server.app, server.user_datastore)
 
-    def _init_default_tenant(self, db, app):
```
mcouthon: No longer needed - this is being taken care of in `create_default_user_tenant_and_roles`

rest-service/manager_rest/utils.py
```diff
@@ -173,87 +164,5 @@ def create_auth_header(username=None, password=None, token=None, tenant=None):
     elif token:
         headers = {CLOUDIFY_AUTH_TOKEN_HEADER: token}
     if tenant:
-        headers[CLOUDIFY_TENANT_HEADER] = tenant
+        headers[constants.CLOUDIFY_TENANT_HEADER] = tenant
     return headers
-
-
-def create_security_roles_and_admin_user(user_datastore,
```
mcouthon: Moved to a separate module

rest-service/manager_rest/utils.py
```diff
@@ -173,87 +164,5 @@ def create_auth_header(username=None, password=None, token=None, tenant=None):
...
-        id=BOOTSTRAP_ADMIN_ID,
-        username=admin_username,
-        password=encrypt_password(admin_password),
-        roles=[admin_role]
-    )
-    user_obj.tenants.append(default_tenant)
-    user_datastore.commit()
-
-
-def setup_flask_app(db, user_datastore, manager_ip='localhost', driver=''):
```
mcouthon: Moved to a separate module

tests/integration_tests/tests/utils.py
```diff
@@ -199,12 +198,5 @@ def patch_yaml(yaml_path, is_json=False, default_flow_style=True):
         yield patch
 
 
-def get_remote_storage_manager():
```
mcouthon: Was never being used, and would pose a problem anyway (because of the open SQL session).

rest-service/manager_rest/storage/utils.py
```diff
@@ -0,0 +1,54 @@
+#########
```
mcouthon: rename to storage_utils

PR #693, opened 2017-01-24
Remove six dependency

jcollado:
Test cases have started failing all of sudden apparently because `six==1.10.0` is required in some dependency while the rest service itself required `six==1.8.0`. However, I haven't seen any place in which the rest service code makes use `six`, so it makes sense to remove the dependency from `setup.py` to fix the conflict.

### DIFF
--- a/rest-service/setup.py
+++ b/rest-service/setup.py
@@ -19,7 +19,6 @@ from setuptools import setup
 
 
 install_requires = [
-    'six==1.8.0',
     'Flask==0.10.1',
     'flask-restful==0.2.5',
     'flask-restful-swagger==0.12',


jcollado:
@mcouthon helped me to figure out that the new dependency comes from `setuptools` whose latest version is installed in the test environment as it can be seen [here|https://github.com/pypa/setuptools/commit/089cdeb489a0fa94d11b7307b54210ef9aa40511#diff-2eeaed663bd0d25b7e608891384b7298R167].

Despite this won't be very common, maybe it makes sense to install a specific version of the tools used for testing to avoid this kind of problems. What do you think?

funkyHat:
@jcollado pinning the version of setuptools (etc.) in the test environment would have hidden this issue in the tests so a user could find it for us instead, so we probably shouldn't do that

jcollado:
@funkyHat That's a good point. Thanks for the comment.

PR #694, opened 2017-01-24
CFY-6340 Allow working with the bootstrap admin even when LDAP is configured

mcouthon:


rest-service/manager_rest/security/authentication.py
```diff
@@ -56,26 +56,55 @@ def authenticate(self, request):
...
+        :param user: The DB user object
+        :param username: The username from the request
+        :param password: The password from the request
+        :return: The DB user object
+        """
+        self.logger.debug('Running basic HTTP authentication')
+        if not user:
+            raise_unauthorized_user_error(
+                'Authentication failed for '
+                '<User username=`{0}`>'.format(username)
```
dusking: I think that the syntax <User username='''>  gives the impression that it's a print of the User object, but we don't really have User object for that username..

PR #700, opened 2017-01-30
CFY 6332. Fix events integration tests cases

jcollado:
In this PR code has been updated to support parameters that were not part of the initial implementation like `_include` to apply projections and `_range` to filter by timestamp. This means that the integration test cases that depend on this behavior have been fixed.

Note that there are still a couple of test cases that need to be fixed that depend on a `message.text` parameter that was used to match messages using full-text search. This will be addressed in a separate PR.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v1/events.py
+++ b/rest-service/manager_rest/rest/resources_v1/events.py
@@ -253,7 +253,8 @@ class Events(SecuredResource):
                 event[key] = '{}Z'.format(value.isoformat()[:-3])
 
         # Keep only keys passed in the _include request argument
-        event = dicttoolz.keyfilter(lambda key: key in _include, event)
+        if _include is not None:
+            event = dicttoolz.keyfilter(lambda key: key in _include, event)
 
         return event
 
--- a/rest-service/manager_rest/rest/resources_v1/events.py
+++ b/rest-service/manager_rest/rest/resources_v1/events.py
@@ -374,7 +374,7 @@ class Events(SecuredResource):
             'offset': pagination['offset'],
         }
 
-        count_query = self._build_count_query(filters)
+        count_query = self._build_count_query(filters, range_filters)
         total = count_query.params(**params).scalar()
 
         select_query = self._build_select_query(
--- a/rest-service/manager_rest/rest/resources_v1/events.py
+++ b/rest-service/manager_rest/rest/resources_v1/events.py
@@ -227,6 +227,9 @@ class Events(SecuredResource):
 
         if 'execution_id' in filters:
             events_query.filter(Execution.id == bindparam('execution_id'))
+        if 'deployment_id' in filters:
+            events_query = events_query.filter(
+                Deployment.id.in_(filters['deployment_id']))
 
         if 'cloudify_log' in filters['type']:
             logs_query = (
@@ -239,6 +242,9 @@ class Events(SecuredResource):
             if 'execution_id' in filters:
                 logs_query = logs_query.filter(
                     Execution.id == bindparam('execution_id'))
+            if 'deployment_id' in filters:
+                logs_query = logs_query.filter(
+                    Deployment.id.in_(filters['deployment_id']))
 
             query = db.session.query(
                 events_query.subquery().c.count +
--- a/rest-service/manager_rest/rest/resources_v1/events.py
+++ b/rest-service/manager_rest/rest/resources_v1/events.py
@@ -54,6 +54,39 @@ class Events(SecuredResource):
     DEFAULT_SEARCH_SIZE = 10000
 
     @staticmethod
+    def _apply_sort(query, sort):
+        """Apply sorting criteria.
+
+        Sorting will be rejected if the field doesn't match any of the column
+        names that has been selected for the query. Note that the query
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/resources_v1/events.py
```diff
@@ -51,38 +54,98 @@ class Events(SecuredResource):
...
+        :type range_filters: dict(str, dict(str))
+        :returns: Query with filtering applied
+        :rtype: :class:`sqlalchemy.orm.query.Query`
+
+        """
+        for field, range_filter in range_filters:
+            # Drop `@` prefix for compatibility
+            # with old Elasticsearch based implementation
+            field = field.lstrip('@')
+            if hasattr(model, field):
```
idanmo: Shouldn't an error be raised if the attribute does not exist on the model?

rest-service/manager_rest/rest/resources_v1/events.py
```diff
@@ -51,38 +54,98 @@ class Events(SecuredResource):
...
+            :class:`manager_rest.storage.resource_models.Log`
+        :param field: Field in the model that should be filtered
+        :type field: str
+        :param range_filter: Range filter passed as a request argument
+        :type range_filter: dict(str)
+        :returns: Query with filtering applied
+        :rtype: :class:`sqlalchemy.orm.query.Query`
+
+        """
+        if 'from' in range_filter:
```
idanmo: what's the behavior if 'from' or 'to' are not in range_filter? what if both of them are not there?

rest-service/manager_rest/rest/resources_v1/events.py
```diff
@@ -128,37 +194,61 @@ def _build_select_query(_include, filters, pagination, sort):
...
+            if 'deployment_id' in filters:
+                logs_query = logs_query.filter(
+                    Deployment.id.in_(filters['deployment_id']))
 
-        if '@timestamp' not in sort or sort['@timestamp'] != 'asc':
-            raise manager_exceptions.BadParametersError(
-                'Sorting ascending by `timestamp` is expected')
+            query = query.union(logs_query)
+
+        if 'execution_id' in filters:
```
idanmo: deployment_id cannot be bound as well? instead of per query?

rest-service/manager_rest/rest/resources_v1/events.py
```diff
@@ -128,37 +194,61 @@ def _build_select_query(_include, filters, pagination, sort):
...
+            query = query.filter(Execution.id == bindparam('execution_id'))
+
+        if '@timestamp' in sort:
```
idanmo: should also support 'timestamp' ?

rest-service/manager_rest/rest/resources_v1/events.py
jcollado: At first I though that just ignoring that filtering criteria would be a good enough, but now I think that indeed an error is better to let the user know that there was a problem which was maybe a typo in the request.

rest-service/manager_rest/rest/resources_v1/events.py
jcollado: From this method's point of view:
- when `from` is missing, all records that are oldest than the timestamp in `to` will be retrieved
- when `to` is missing, all records that are newer than the timestamp in `from` will be retrieved
- when both `from` and `to` are missin, all records will be retrieved (no filtering applied)

However, from the [`rangeable`](https://github.com/cloudify-cosmo/cloudify-manager/blob/CFY-6332-fix-events-integration-tests/rest-service/manager_rest/rest/rest_decorators.py#L206) decorator point of view, all three values (field name, from and to) must be present.

What might be missing here is that rangeable doesn't enforce any validation of the values in `from` and `to`, which makes sense because the might be numbers, dates or some other thing. However, the decorator can be probably expanded to pass a validation function to make sure that `from` and `to` have the right format. In any case, this is out of the scope of the current PR, so I'll do that in a separate one.

rest-service/manager_rest/rest/resources_v1/events.py
jcollado: I think that both implementations are possible, but I'm under the impression that filtering in the subqueries is more performant than after the union. I'll have a look at it, but it probably makes sense to do this filtering per subquery.

rest-service/manager_rest/rest/resources_v1/events.py
jcollado: What I've seen in the integration test cases and in the CLI is just the use of `@timesamp`. However, it makes sense to support `timestamp` to allow and deprecate `@timestamp`.

rest-service/manager_rest/rest/resources_v1/events.py
jcollado: Fixed in e037754

rest-service/manager_rest/rest/resources_v1/events.py
jcollado: Fixed in d9e2532

rest-service/manager_rest/rest/resources_v1/events.py
jcollado: Filtering has been refactored to be applied per query in cd66a13

PR #701, opened 2017-01-30
CFY-6207 add endpoint for enabling authentication against LDAP

adamlavie:


rest-service/manager_rest/rest/resources_v3.py
```diff
@@ -387,3 +391,63 @@ def delete(self):
...
+        # be loaded to all flask processes.
+        set_restart_task()
+
+        ldap_config.pop('ldap_password')
+        return ldap_config
+
+
+def _is_manager_clean():
+    """
+    True if no users other than the admin use exists.
```
dusking: admin user.. (forgot the "r")
and should be after the return...

rest-service/manager_rest/rest/resources_v3.py
```diff
@@ -387,3 +391,63 @@ def delete(self):
...
+
+        # Restart the rest service so that each the LDAP configuration
+        # be loaded to all flask processes.
+        set_restart_task()
+
+        ldap_config.pop('ldap_password')
+        return ldap_config
+
+
+def _is_manager_clean():
```
dusking: maybe _only_admin_in_manager or something like that, 
because manager might not be clean (might contain blueprint, deployments, etc.)

rest-service/manager_rest/rest/resources_v3.py
```diff
@@ -387,3 +391,63 @@ def delete(self):
         """
         params = self._get_and_validate_params()
         return PermissionsHandler.remove_permissions(params)
+
+
+class LdapAuthentication(SecuredResource):
+    @rest_decorators.exceptions_handled
+    @rest_decorators.marshal_with(LdapResponse)
+    def post(self):
```
dusking: try to split into functions...

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -98,3 +100,9 @@ def make_streaming_response(res_id, res_path, content_length, archive_type):
     response.headers['X-Accel-Redirect'] = res_path
     response.headers['X-Accel-Buffering'] = 'yes'
     return response
+
+
+def set_restart_task(delay=0.5):
```
dusking: maybe better to have 1sec, to be in the safe side..

PR #727, opened 2017-02-16
CFY-6437 fix snapshots of v4.0

dusking:


workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -203,14 +191,15 @@ def make_zip64_archive(zip_filename, directory):
...
     with zip_context_manager as zip_file:
         path = os.path.normpath(directory)
-        zip_file.write(path)
+        ctx.logger.debug('Creating zip archive of: {0}'.format(path))
+        len_base_dir = len(path)
         for dirpath, dirnames, filenames in os.walk(directory):
             for dirname in sorted(dirnames):
                 path = os.path.normpath(os.path.join(dirpath, dirname))
-                zip_file.write(path)
+                zip_file.write(path, path[len_base_dir:])
```
jcollado: What about using `os.path.relpath(path, base_dir)` instead of `path[len_base_dir:]`. The result should be the same, but I believe the first option is more readable.

jcollado:
@dusking Thanks for the fix. As @idanmo pointed out in a slack conversation, I overlooked that `shutil.make_archive` uses `os.chdir` and that makes a difference with regard to the paths stored in the zip file.

workflows/cloudify_system_workflows/snapshots/utils.py
dusking: nice, agree

PR #735, opened 2017-02-23
CFY-6463 Users are now activated/deactivated instead of suspended

mcouthon:


rest-service/manager_rest/rest/resources_v3.py
```diff
@@ -262,6 +262,24 @@ def delete(self, username, multi_tenancy):
...
+    @rest_decorators.marshal_with(UserResponse)
+    def post(self, username, multi_tenancy):
+        """
+        Activate a user
+        """
+        return multi_tenancy.activate_user(username)
+
+    @rest_decorators.exceptions_handled
+    @rest_decorators.marshal_with(UserResponse)
+    def delete(self, username, multi_tenancy):
```
dusking: use post with data instead of delete..

PR #745, opened 2017-03-01
CFY-6492 Make the file server always secured

mcouthon:
* The file server is now always internally accessed via the REST's port (with a /resources suffix), and is then redirected to  its internal port
 * All the folder names now reside in cloudify-plugins-common.constants (seeing as how they are indeed constants)
 * All the workers (agent + mgmt) now need to have much fewer params regarding the file server


packaging/restservice/build.spec
```diff
@@ -40,7 +40,7 @@ sudo /tmp/env/bin/pip install wheel==0.24.0 && \
 export REST_SERVICE_BUILD=True
 
 destination="/tmp/${RANDOM}.file"
-curl --retry 10 --fail --silent --show-error --location https://github.com/cloudify-cosmo/cloudify-manager/archive/%{CORE_TAG_NAME}.tar.gz --create-dirs --output $destination && \
+curl --retry 10 --fail --silent --show-error --location https://github.com/cloudify-cosmo/cloudify-manager/archive/CFY-6492-secure-file-server.tar.gz --create-dirs --output $destination && \
```
dusking: ..

rest-service/manager_rest/rest/resources_v2/plugins.py
```diff
@@ -39,6 +38,9 @@
 from manager_rest.upload_manager import UploadedPluginsManager
 from manager_rest.utils import create_filter_params_list_description
```
dusking: space..

PR #747, opened 2017-03-02
CFY-6508 remove cloudify-aria-extensions

funkyHat:


tests/integration_tests/tests/utils.py
```diff
@@ -75,7 +75,7 @@ def create_rest_client(**kwargs):
 
 
 def wait_for_deployment_creation_to_complete(
-        deployment_id, timeout_seconds=30):
+        deployment_id, timeout_seconds=60):
```
idanmo: why is the change from 30 to 60?

funkyHat: These tests are failing on my laptop with the timeout at 30 (this doesn't actually make each test take another 30s, as it's checking every 0.1s).

PR #762, opened 2017-03-14
CFY-6530 Add a new DSL function `get_secret`

mcouthon:


rest-service/manager_rest/functions.py
```diff
@@ -0,0 +1,80 @@
+#########
```
dusking: dsl_functions

PR #763, opened 2017-03-14
CFY-6359 Status: replace the old webui with stage

tehasdf:


tehasdf:
todo also check on docl

PR #764, opened 2017-03-14
CFY-6553 Remove the runit/docker part from the status endpoint

tehasdf:
Introduced in CFY-1308, the docker part was obsolete and long not used.


rest-service/manager_rest/rest/resources_v1/status.py
```diff
@@ -25,6 +23,10 @@
 )
 from manager_rest.security import SecuredResource
 from manager_rest.security.authorization import authorize
+try:
+    from manager_rest.systemddbus import get_services
+except ImportError:
+    get_services = None
```
geokala: What situations trigger this, and what is its purpose?

geokala: Specifically: It's not a premium-only thing. We only support running the manager on centos/rhel7, which requires systemd, so... when should this import fail?

tehasdf: in theory, you can install the restservice wherever you want, including on systems that don't have systemd/dbus. We even declare dbus as an optional dependency here:
https://github.com/cloudify-cosmo/cloudify-manager/blob/05b5a75832dc3e1eceab755f9f883959f3a3f966/rest-service/setup.py#L58
so if it's optional, we must handle the case of it not being there
(in this case "handling" means returning a useless noop response instead of a 500, but we can't do better than that)

geokala: OK, I'm not convinced on the base design of that, but given the way it has been designed, this looks as correct as can be, yes.

PR #772, opened 2017-03-16
CFY 6557. Support empty values in rangeable decorator

jcollado:
In this PR, support for empty values has been added back to the `rangeable` rest decorator. This means that instead filtering by `field,from,to`, it's possible not to include one of the `from`/`to` fields: `field,,to` or `field,from,`. This way it's possible to use the filter when only an upper/lower limit is needed.

Datetime parsing has been improved as well by using `dateutil.parser.parse` because from what I've seen the `Datetime` validator from `voluptuous` wasn't accepting iso8601 datetime which should be accepted by the events endpoint.

### DIFF
--- a/rest-service/manager_rest/rest/rest_decorators.py
+++ b/rest-service/manager_rest/rest/rest_decorators.py
@@ -272,7 +272,7 @@ def rangeable(func):
     schema = Schema(
         All(
             ExactSequence([
-                str,
+                basestring,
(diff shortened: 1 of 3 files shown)


jcollado:
There was a problem with validation that has been fixed.

The problem was that headers coming from flask are unicode strings, not regular strings, so the schema needed an update.

PR #776, opened 2017-03-20
CFY-6601 snapshot broken, specify pg port if needed

dusking:


workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -118,6 +125,7 @@ def drop_db(self):
         drop_db_bin = os.path.join(self._bin_dir, 'dropdb')
         command = [drop_db_bin,
                    '--host', self._host,
+                   '--port', self._port,
```
tehasdf: port probably needs to be `str`, if it can be an int

dusking: it is... self._port is a substr of self._host, which is a string

dusking: fixed by changing POSTGRESQL_DEFAULT_PORT to str, because otherwise it's substr of self._host..

PR #778, opened 2017-03-21
CFY-6408. Support more filter fields

jcollado:
In this PR, support for the following filters is added to the events endpoint: `blueprint_id`, `event_type`, `level` and `message`/`message.text`.

Note: For the `message`/`message.text` filter the `ILIKE` operator is used which is not equivalent to full text search. Consequently, even if filtering by message is supported, it is not equivalent in terms of functionality or performance to what it was available when data was indexed through Elasticsearch.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v1/events.py
+++ b/rest-service/manager_rest/rest/resources_v1/events.py
@@ -53,8 +53,10 @@ class Events(SecuredResource):
 
     DEFAULT_SEARCH_SIZE = 10000
     ALLOWED_FILTERS = {
-        'execution_id': Execution,
-        'deployment_id': Deployment,
+        'execution_id': Execution.id,
+        'deployment_id': Deployment.id,
+        'event_type': Event.event_type,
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/rest/resources_v1/events.py
```diff
@@ -46,39 +47,67 @@ class Events(SecuredResource):
...
+                model_field = getattr(model, model_field)
+
+            if filter_type == 'in':
+                query = query.filter(model_field.in_(filter_))
+            elif filter_type == 'ilike':
+                for filter_element in filter_:
+                    query = query.filter(model_field.ilike(filter_element))
+            else:
+                raise ValueError(
+                    'Unknown filter type: {0}'.format(filter_type))
```
idanmo: Perhaps include the allowed filter types in this message?

jcollado: Good suggestion! Fixed in 45dc47c

PR #808, opened 2017-04-12
CFY-6660. Remove deployment permission assertion

jcollado:
In this PR, an assertion that was causing problems in snapshot integration test cases has been removed because it's not longer needed.


jcollado:
Added `deployment.created_by` assertion as suggested by @mcouthon

PR #815, opened 2017-04-22
Fix snapshot permissions

geokala:
Required for https://github.com/cloudify-cosmo/cloudify-manager-blueprints/pull/556

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
+++ b/workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
@@ -19,6 +19,7 @@ import shutil
 import zipfile
 import platform
 import tempfile
+import subprocess
 from contextlib import contextmanager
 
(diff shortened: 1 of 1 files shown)


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -92,6 +93,9 @@ def restore(self):
                 self._restore_credentials(postgres)
                 self._restore_agents()
                 self._restore_deployment_envs(existing_dep_envs)
+                subprocess.check_call([
```
idanmo: Can't we make sure the permissions are correct on bootstrap?

geokala: We do, but when the restore runs it copies a bunch of files around and messes up the permissions.

PR #817, opened 2017-04-24
CFY-6750 Use flask-migrate to recreate the DB in integration tests

mcouthon:


tests/integration_tests/framework/flask_utils.py
```diff
@@ -14,9 +14,13 @@
 # limitations under the License.
 import yaml
 import logging
+from path import path
```
idanmo: from path import `Path`?

PR #821, opened 2017-04-25
CFY-6718 Migration path from 4.0 to 4.0.1

jcollado:
In this PR, the migration scripts directory has been updated to have:
- a script to update from scratch to the 4.0 schema
- another script from 4.0 to 4.0.1 (assuming the schema is not going to be updated again before the release).

The reason to split the original script in two different steps is to have in place what is needed to be able to restore a 4.0 snapshot. The current plan is to downgrade the database to a 4.0 schema, restore the snapshot and upgrade the schema afterwards. However, this will happen in a separate PR, since this one is already big enough.

The scripts have been tested using both PostgreSQL and SQLite with no data and they work fine. When data is in the database they will fail because of database constraints not met for fields such as `_creator_id`, `_tenant_id` and `reported_timestamp`. All these problems will become apparent when the code to restore a 4.0 snapshot is written, so they will be addressed then.


jcollado:
It turns out that data (not schema) migration is not well supported for SQLite because of the lack of a documented API to execute custom SQL code and get the temporary table name used during batch operations.

A new PR will be created that only contains the changes needed to support PostgreSQL in the migration scripts.

PR #824, opened 2017-04-26
CFY-6781 Fix rogue deployment resource test

mcouthon:


mcouthon:
* The join on the daemon thread was getting stuck sometimes. After reading a bit on it, it's probably best to just risk the occasional traceback.
* In the deployment resources test we needed to `chown` the whole directory.
* In the modifications test I was trying to write a dict to file.

PR #833, opened 2017-04-29
CFY-6791 move snapshot agent keys to secret store

funkyHat:


workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -182,8 +149,38 @@ def _create_agent_key_path_dict(self):
...
+        )
+
+    client = CloudifyClient(
+        'localhost',
+        token=ctx.rest_token,
+        tenant=tenant,
+    )
+    secrets = client.secrets
+
+    while True:
```
idanmo: No while True please :|

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -182,8 +149,38 @@ def _create_agent_key_path_dict(self):
...
+
+
+def add_key_secret(tenant, key_path, key_data):
+    key_name = SECRET_STORE_AGENT_KEY_PREFIX + ''.join(
+        char if char in ALLOWED_KEY_CHARS else '_'
+        for char in key_path
+        )
+
+    client = CloudifyClient(
+        'localhost',
```
funkyHat: Not that I found. The secret store is not very well documented but certainly doesn't seem to be available from within a SystemWorkflowContext

workflows/cloudify_system_workflows/snapshots/credentials.py
funkyHat: How would you suggest revising?
We are looping until we find either a matching auth key (line 183) or a secret key name which isn't used yet (this is needed because I'm not allowed to use `/` and a whole bunch of other stuff, so we might get key name collisions for filenames which are distinct)

When I started writing this function I initially had something along the lines of

```python
secret_value = None
while secret_value != key_data and ...:
    ...
```
but I think that's actually much less clear, and the chance of this accidentally being an infinite loop is just the same

(of course I actually *did* forget the bit which mutates the name before we try again, so I do agree it's a concern. Fixed that now :)

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -53,9 +58,10 @@ def dump(self, tempdir):
 
         for deployment_id, n in self._get_hosts():
             props = n.properties
-            if 'cloudify_agent' in props and 'key' in props['cloudify_agent']:
+            agent_config = self._get_agent_config(props)
+            if 'key' in agent_config:
                 node_id = deployment_id + '_' + n.id
-                agent_key_path = props['cloudify_agent']['key']
+                agent_key_path = agent_config['key']
```
funkyHat: Good point, the whole thing will fall over if it hits one of those right now. I'll fix

workflows/cloudify_system_workflows/snapshots/credentials.py
idanmo: I meant that there's probably a way to get a REST client from some helper method which already does that..

funkyHat: Ah I see :). Yes, looks like `cloudify.manager.get_rest_client()` :) thanks

workflows/cloudify_system_workflows/snapshots/credentials.py
funkyHat: (not sure why GH things this is outdated)

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -182,8 +149,35 @@ def _create_agent_key_path_dict(self):
...
+        )
+
+    client = get_rest_client()
+    secrets = client.secrets
+
+    while True:
+        try:
+            secret_value = secrets.get(key_name)
+        except CloudifyClientError as e:
+            if e.status_code == 404 and 'not found' in str(e):
```
idanmo: why do you need the 'not found' check here?

funkyHat: Good question. Removed

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -69,99 +75,67 @@ def _get_hosts():
...
-
-        if os.path.isfile(db_agent_key_path):
-            self._handle_existing_agent_key_path(
-                restored_agent_key_path,
-                db_agent_key_path
-            )
-        else:
-            copy_file(restored_agent_key_path, db_agent_key_path)
+        try:
+            shutil.copy(source, destination)
```
idanmo: Before copying, make sure this is a file with something like:
if os.path.exists(source) and os.path.isfile(source):

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -13,52 +13,53 @@
...
-            if 'cloudify_agent' in props and 'key' in props['cloudify_agent']:
+            agent_config = get_agent_config(props)
+            if 'key' in agent_config:
                 node_id = deployment_id + '_' + n.id
-                agent_key_path = props['cloudify_agent']['key']
+                agent_key = agent_config['key']
                 self._dump_agent_key(
                     node_id,
-                    agent_key_path,
+                    agent_key,
```
idanmo: What is this is not a key path?
Possible values:
1. Private key content.
2. Private key path.
3. get_secret intrinsic function.

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -69,121 +70,136 @@ def _get_hosts():
...
-        with open(restored_key_path) as key_file:
-            content_2 = key_file.read()
-        if content_1 != content_2:
-            raise NonRecoverableError(
-                'Agent key path already taken: {0}'.format(db_key_path)
-            )
-        ctx.logger.debug('Agent key path already exist: '
-                         '{0}'.format(db_key_path))
+        try:
+            shutil.copy(source, destination)
```
idanmo: why not checking if `source` exists before copying?

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -69,121 +70,136 @@ def _get_hosts():
...
+                raise
+
+
+def get_agent_config(node_properties):
+    """cloudify_agent is deprecated, but still might be used in older
+    systems, so we try to gather the agent config from both sources
+    """
+    cloudify_agent = node_properties.get('cloudify_agent', {})
+    agent_config = node_properties.get('agent_config', {})
+    agent_config.update(cloudify_agent)
```
idanmo: Shouldn't `agent_config` take precedence?

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -69,121 +70,136 @@ def _get_hosts():
...
+        return
+
+    client = get_rest_client()
+
+    tenants = [t.name for t in client.tenants.list()]
+
+    credential_dirs = set(os.listdir(dump_cred_dir))
+
+    for tenant in tenants:
+        client._client._set_header('Tenant', tenant)
```
idanmo: Perhaps you should retain and set the "original" tenant value in the client before existing this method.. (try..finally..).

workflows/cloudify_system_workflows/snapshots/credentials.py
```diff
@@ -69,121 +70,136 @@ def _get_hosts():
...
+            if 'key' not in agent_config:
+                continue
+
+            agent_key = agent_config['key']
+            dir_name = deployment_id + '_' + node_id
+
+            if not isinstance(agent_key, basestring):
+                ctx.logger.info('key for {} is not a path'.format(dir_name))
+                continue
+            if re.search('BEGIN .* PRIVATE KEY', agent_key):
```
idanmo: What if `agent_key` = `get_secret` intrinsic function?

workflows/cloudify_system_workflows/snapshots/credentials.py
funkyHat: Good point. I'll fix. Was copied from the existing code like this...

workflows/cloudify_system_workflows/snapshots/credentials.py
funkyHat: Yep, or perhaps it's better to switch back to constructing a client per tenant, rather than messing with underscored vars (there's no way to pass a tenant in per operation)?

workflows/cloudify_system_workflows/snapshots/credentials.py
funkyHat: Then it will not be a string, so it gets caught by line 158

funkyHat: (we're storing the properties as a pickle(?!) in the DB, so it will be a dict)

workflows/cloudify_system_workflows/snapshots/credentials.py
funkyHat: It's against my religion (but seriously BAFP is generally considered more pythonic... and I know at the moment it's not really racey because nothing else *should* be happening at the same time on the manager, but I'd still rather try to set an example, and in any case this isn't *worse*.
This also means there's no need to write an "is it a file path?" function, with potential bugs in it. Just let the OS tell us if it's a file path.

workflows/cloudify_system_workflows/snapshots/credentials.py
funkyHat: See below...

PR #839, opened 2017-05-03
CFY 6718. Restore 4.0 snapshot

jcollado:
In this PR, the ability to restore 4.0 snapshots by running database schema migrations has been implemented.

Note that the migration scripts need to run in the manager virtualenv while the postgres restore itself runs in the management worker virtualenv. When a better design for restoring snapshots is in place, this should be refactored.

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
+++ b/workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
@@ -141,7 +141,6 @@ class SnapshotRestore(object):
         elif self._snapshot_version == V_4_0_0:
             with utils.db_schema('333998bc1627'):
                 postgres.restore(self._tempdir)
-            postgres.restore_stage(self._tempdir)
         else:
             if self._should_clean_old_db_for_3_x_snapshot():
                 postgres.clean_db()
--- a/workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
+++ b/workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
@@ -139,26 +139,9 @@ class SnapshotRestore(object):
             postgres.restore(self._tempdir)
             postgres.restore_stage(self._tempdir)
         elif self._snapshot_version == V_4_0_0:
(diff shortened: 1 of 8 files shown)


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -135,8 +135,28 @@ def _restore_files_to_manager(self):
 
     def _restore_db(self, postgres):
         ctx.logger.info('Restoring database')
-        if self._snapshot_version >= V_4_0_0:
+        if self._snapshot_version > V_4_0_0:
+            postgres.restore(self._tempdir)
+        elif self._snapshot_version == V_4_0_0:
+            base_script = (
+                'import flask_migrate;'
```
idanmo: Perhaps it would be better to add this script as a resource to the project which will accept arguments and behave according to them?

jcollado: I've just committed a refactoring that basically moves the ugly code to the utils module. Please let me know if it looks better now.

workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -218,3 +219,55 @@ def compare_cert_metadata(path1, path2):
...
+
+    Used before restoring a snapshot to make sure that the schema matches the
+    one that was used when the snapshot was created.
+
+    :param revision: Revision to downgrade to.
+    :type revision: str
+
+    """
+    subprocess.check_call([
+        '/opt/manager/env/bin/python',
```
idanmo: Perhaps have a constant for Python path?

jcollado: Fixed in e52f343

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
jcollado: Fixed in 63b1ca0

PR #843, opened 2017-05-04
Cfy 6791 snapshot secrets tests

geokala:


tests/integration_tests/tests/agentless_tests/test_snapshot_restore_secrets.py
```diff
@@ -0,0 +1,315 @@
...
+
+class TestSnapshotSecrets(AgentlessTestCase):
+    maxDiff = None
+    run_storage_reset = False
+
+    def test_restore_3_4_2_secrets(self):
+        # Yes, there are six private keys in this file
+        # This will present a grave risk to the no-longer-existing VMs that
+        # the snapshots were created on.
+        key1 = '''
```
idanmo: Perhaps store the pem files in the resources folder, file per key?

tests/integration_tests/tests/utils.py
```diff
@@ -200,3 +204,110 @@ def patch_yaml(yaml_path, is_json=False, default_flow_style=True):
...
+    validate_snapshot(snapshot_id, path_to_snapshot)
+
+    execution = restore_snapshot(snapshot_id, tenant)
+
+    # If checking the workflow status immediately, we will be likely to cause
+    # a deadlock on the database and make the restore fail
+    # This may be possible to do using the cfy utility, but I've not managed
+    # to be fast enough with it to trigger the issue.
+    # NOTE: Jira should be opened for this
+    time.sleep(3)
```
idanmo: You can probably change this to: `wait_for_restore_workflow_execution_to_appear`

tests/integration_tests/tests/utils.py
```diff
@@ -200,3 +204,110 @@ def patch_yaml(yaml_path, is_json=False, default_flow_style=True):
...
+
+    snapshot = client.snapshots.get(snapshot_id)
+    assert snapshot['status'] == 'uploaded', (
+        'Snapshot {name} did not upload correctly.'.format(
+            name=original_name,
+        )
+    )
+
+
+def wait_for_execution_to_end(execution,
```
idanmo: Use the `retrying.retry` decorator instead of this while loop.

tests/integration_tests/tests/agentless_tests/test_snapshot_restore_secrets.py
geokala: That's doable- I'm not sure it gains us a huge amount though?

tests/integration_tests/tests/utils.py
geokala: Does that handle the manager not being up? There's a brief moment during the restore when you get 500s from the server because of the DB being messed about with.

Also, I can't find a function definition for that or anything super-similar, aside from wait_for_execution_to_end

tests/integration_tests/tests/utils.py
geokala: Okay, will convert to that tomorrow AM- sorry, this was just a minor tweak to the existing logic.

geokala: On looking at this, if I make this change I basically remove the ability to specify the timeout for the wait. That'll change the way this function works beyond just extending its functionality to support the snapshots.

Are we happy to just have one timeout specified for this?

tests/integration_tests/tests/utils.py
idanmo: Well, you can write your own function and use the `retrying.retry` decorator.
It can be retried for any error until the execution's status == terminated.

tests/integration_tests/tests/utils.py
idanmo: Sounds fine.

geokala:
With thanks to @tehasdf for the race condition fix on the downgrade.

geokala:
S3 uploads have been requested so I'll test with them as soon as they're uploaded.

resources/rest-service/cloudify/migrations/versions/9aa6f74c9653_4_0_1.py
```diff
@@ -397,6 +397,7 @@ def upgrade():
 
 def downgrade():
     # ### commands auto generated by Alembic - please adjust! ###
+    op.execute('LOCK TABLE users IN ACCESS EXCLUSIVE MODE')
```
tehasdf: The issue is that during the test, while the migration runs and tries to grab all kinds of locks on the users table (because of adding FK's etc), the test framework (or the CLI, actually) is going to hammer the server by polling the execution status - which runs the auth logic - which updates `last_login_at` - which needs to grab a writer lock on the user row.

This was causing a deadlock in cases when the `last_login_at` update happened somewhere in the middle of this downgrade transaction block.

Btw, to easier reproduce this deadlock (which seems to be very dependent on the environment, and luck - like most race conditions), one way is to edit the test framework code that polls execution status (in `wait_for_execution_to_end`), to poll eg. every 10 msec, not 500.

This change just grabs a lock on the whole table, so all requests will simply have to wait for the downgrade to finish, but should fix the deadlocks.

geokala:

PR #844, opened 2017-05-07
CFY-6828 support ui db migration

yarivgiga:


workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -292,3 +292,19 @@ def db_schema_get_current_revision():
     ])
     revision = output.split(' ', 1)[0]
     return revision
+
+
+def stage_db_schema_get_current_revision():
```
idanmo: Perhaps name the function: `get_stage_current_db_schema_revision` ?

workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -292,3 +292,19 @@ def db_schema_get_current_revision():
...
+    """Get stage database schema revision.
+
+    :returns: Current revision
+    :rtype: str
+
+    """
+    output = subprocess.check_output([
```
idanmo: I don't know what's the right handling here, but for community edition this will not work..

workflows/cloudify_system_workflows/snapshots/utils.py
yarivgiga: makes sense, but i wanted it to be compatible with the other funcs (db_schema_...)

workflows/cloudify_system_workflows/snapshots/utils.py
yarivgiga: why not?
and maybe we can use `try`?

PR #847, opened 2017-05-07
Fix executions sort test

inbala:


tests/integration_tests/tests/agentless_tests/test_rest_service_sort.py
```diff
@@ -50,8 +50,19 @@ def test_executions_sort(self):
         for i in range(5):
             self.execute_workflow('install', deployment.id)
             self.execute_workflow('uninstall', deployment.id)
-        self._test_sort('executions',
-                        ['deployment_id', '-status'])
+
+        api = getattr(self.client, 'executions')
```
idanmo: Why don't you use: `self.client.executions.list` ?

tests/integration_tests/tests/agentless_tests/test_rest_service_sort.py
```diff
@@ -50,8 +50,19 @@ def test_executions_sort(self):
...
+        sorted_list = api.list(_sort=['-workflow_id', 'status'])
+        self.assertGreater(len(sorted_list), 0)
```
idanmo: If you expecting the list to have at least 11 items, then perhaps change the assertion to at least 11..

PR #848, opened 2017-05-07
CFY-6828 verify premium edition before storing stage revision in snap…

yarivgiga:
…shot

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/utils.py
+++ b/workflows/cloudify_system_workflows/snapshots/utils.py
@@ -21,7 +21,7 @@ import shutil
 import subprocess
 import zipfile
 
-from cloudify import constants
+from cloudify import constants, manager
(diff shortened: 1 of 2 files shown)


workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -301,6 +302,10 @@ def stage_db_schema_get_current_revision():
     :rtype: str
 
     """
+    headers = {constants.CLOUDIFY_TOKEN_AUTHENTICATION_HEADER: ctx.rest_token}
```
idanmo: Change to:

from cloudify import manager
client = manager.get_rest_client()
version = client.manager.get_version()
if version['edition'] != 'premium' ...

If not relevant the method should return `None` and where you pass the return value to the metadata creation, if the value is `None` don't add it to the snapshot metadata.

PR #852, opened 2017-05-08
CFY 6835. Use schema revision on snapshot restore

jcollado:
In this PR, the code to restore snapshots is updated to take into account the schema revision that is included in the metadata starting from 4.0.1.

Notes:
- this change is expected to be merged in 4.1.
- a local snapshot was used for testing. When 4.0.1 is released, a new snapshot needs to be created and uploaded to the s3 bucket used to store testing snapshots.

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
+++ b/workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
@@ -53,6 +53,9 @@ V_4_0_0 = ManagerVersion('4.0.0')
 
 
 class SnapshotRestore(object):
+
+    SCHEMA_REVISION_4_0 = '333998bc1627'
+
     def __init__(self,
                  config,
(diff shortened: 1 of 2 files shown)


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -79,14 +84,16 @@ def restore(self):
         try:
             metadata = self._extract_snapshot_archive(snapshot_path)
             self._snapshot_version = ManagerVersion(metadata[M_VERSION])
+            # Assume 4.0 schema if metadata is missing
```
dusking: Instead of comment, can add const: first_schema_revision = '333998bc1627'

jcollado: Fixed in bd503dc

PR #855, opened 2017-05-12
CFY-6652. Keep using timestamps in GMT

jcollado:
In this PR, the changes to use the manager timezone for timestamp are reverted and now the events and logs timestamp are now formatted again as GMT.


idanmo:
Just to make sure, when providing a timestamp range to REST API, the range should be UTC as well?

jcollado:
@idanmo That's not a requirement. It might be whatever `dateutils.parser.parse` can parse as can be seen [here|https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/rest/rest_decorators.py#L253].

PR #856, opened 2017-05-13
CFY-6869 4.0.1 DB migration: remove drop_index commands

tehasdf:
All the drop_index commands issued by alembic turned out to be wrong:
the 4.0.1 DB schema contains those indexes, so they should be in the
migration that upgrades the database to 4.0.1.


tehasdf:
In the future, we'll likely need to pay closer attention to commands generated by alembic

tehasdf:
TODO verify that every create has a corresponding drop

tehasdf:
@jcollado I used a docl image, but any bootstrap would be the same, since bootstrapping creates the db from the migrations :) Hence, missing indexes.

Not sure why would alembic get confused and emit wrong drop commands, if I were to guess, I suppose perhaps the migrations were created using a db that was made from a `.create_all()`?

tehasdf:
verified that every `drop_index` in upgrade has a corresponding `create_index` in downgrade

PR #857, opened 2017-05-13
Add index=True to the Tenant FK

tehasdf:
Pretty much all of our resources are queried with a WHERE on tenants,
so let's make that indexable by default.

This is mostly required for the events and logs tables, which can grow 
very large, so we really don't want seq scans on those tables.


tehasdf:
Here's an analysis for the `get events and logs` query: https://gist.github.com/tehasdf/804991c7a560f43eb7fe5226af3a799a

PR #862, opened 2017-05-16
CFY-6652 Fix datetime type in migration scripts

jcollado:
As reported by @mcouthon, the migration scripts where broken when #855 was merged because they still use the old `LodalDateTime` type.

In this PR, that is replaced by `UTCDatetime` which was used in the past. Note that that is SQLAlchemy type that changes serialization, but uses the same underlying database type. Hence, there's no need to create a new migration script.


idanmo:
Are there any implications on 4.0.X to 4.1 migration?

jcollado:
@idanmo The only implication I'm aware of is that the tests that were failing before of the missing data type in the migration script should work now.

PR #865, opened 2017-05-17
REST service config: change default db port from 9200 to 5432

tehasdf:
We're now using postgresql instead of elasticsearch, so we can
default the db port to postgresql's default, instead of
elasticsearch's default


geokala:

PR #866, opened 2017-05-17
CFY-6864 schema.py: use the same postgresql_host as restservice

tehasdf:
Postgresql port isn't always 5432 - when running a cluster, it's by default 15432.

This change makes the port parametrized, by passing the whole `postgresql_host` from restservice - which contains both the host and the port, eg. "localhost:15432"

The `self._config` for the snapshot restore contains that value (it is created here https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/resource_manager.py#L97 ), so we use that - pass the whole config to the utils `db_schema` function, which passes it to the `schema.py` script via a cmdline arg, where it is used for creating a flask app that is used by flask_migrate

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/utils.py
+++ b/workflows/cloudify_system_workflows/snapshots/utils.py
@@ -228,7 +228,7 @@ def compare_cert_metadata(path1, path2):
 
 
 @contextlib.contextmanager
-def db_schema(revision):
+def db_schema(revision, config=None):
     """Downgrade schema to desired revision to perform operation and upgrade.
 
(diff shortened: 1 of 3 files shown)


workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -254,15 +254,10 @@ def db_schema_downgrade(revision='-1'):
     :type revision: str
 
     """
-    subprocess.check_call([
```
tehasdf: factored out the common logic (the subprocess call, passing the --postgresql-host parameter) to a separate function

PR #867, opened 2017-05-18
Cfy 6871 remove source url from validation params

adigrabow:


rest-service/manager_rest/resource_manager.py
```diff
@@ -666,6 +666,7 @@ def create_deployment(self,
         blueprint = self.sm.get(models.Blueprint, blueprint_id)
         plan = blueprint.plan
         try:
+
```
idanmo: Remove blank line

PR #872, opened 2017-05-22
cfy-6904-wip

dusking:


workflows/cloudify_system_workflows/snapshots/npm.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self):
+        self.npm_bin = os.path.join(NODEJS_DIR, 'bin', 'npm')
+
+    @staticmethod
+    def clear_db():
+        """ Clear the Stage DB """
+        ctx.logger.info('Clearing Stage DB')
+        command = ['/opt/nodejs/bin/npm', 'run', 'db-migrate-clear']
+        run_shell(command, cwd='/opt/cloudify-stage/backend')
```
idanmo: Since we are passing this `cwd` in several places (not only this module) - perhaps it would be better to have a utility method for `node/npm` ?

PR #880, opened 2017-06-13
CFY-6966- Snapshot restore improvements (this does not solve all of the jira issue)

geokala:
Makes 3.x snapshot restores only go to the existing tenant, which works around the issue of plugins not being in the tenant after the restore.

Makes plugin wagons be restored to correct tenants for 4.x snapshots.

Disables restoring 3.x snapshots to the not-current tenant.

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/estopg.py
+++ b/workflows/cloudify_system_workflows/snapshots/estopg.py
@@ -49,7 +49,7 @@ class EsToPg(object):
         app = setup_flask_app()
         admin = self._set_current_user(app)
         storage_manager = get_storage_manager()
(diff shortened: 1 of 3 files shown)


geokala:
These address parts of https://cloudifysource.atlassian.net/browse/CFY-6966

However, they don't completely solve it, as there are still issues with plugin installs from source not happening for other tenants (when wagons are missing, or if they don't exist (e.g. a user created plugin)) and with plugin installs from source that depend on other OS packages (e.g. gcc and python-devel) hanging without any information for 25 minutes before failing.

geokala:
Outdated, we've merged this with the full package.

PR #885, opened 2017-06-14
CFY-6882 Delete vhosts and users from unknown tenants

jcollado:
If a service is stopped and restarted later and between those two events
a tenant has been removed, then its related vhost and user should be
removed on restart. This is what the code in this change does.

### DIFF
--- a/rest-service/manager_rest/amqp_manager.py
+++ b/rest-service/manager_rest/amqp_manager.py
@@ -70,6 +70,40 @@ class AMQPManager(object):
 
         return tenant
 
+    def sync_metadata(self, tenants):
+        """Synchronize database tenants with rabbitmq metadata.
+
+        :param tenants: Tenants currently available in the database
+        :type tentans: list(manager_rest.storage.management_models.Tenant)
+
+        """
+        # Remove vhosts in rabbitmq not present in the database
+        expected_vhosts = set(tenant.rabbitmq_vhost for tenant in tenants)
+        current_vhosts = set(
+            vhost
+            for vhost in self._client.get_vhost_names()
+            if vhost.startswith(self.VHOST_NAME_PATTERN[:-3])
+        )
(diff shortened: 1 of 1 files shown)


rest-service/manager_rest/amqp_manager.py
```diff
@@ -70,6 +70,40 @@ def create_tenant_vhost_and_user(self, tenant):
...
+            for user in self._client.get_users()
+            if user['name'].startswith(self.USERNAME_PATTERN[:-3])
+        )
+        extra_usernames = current_usernames - expected_usernames
+        for username in extra_usernames:
+            self._client.delete_user(username)
+
+        # Create vhosts and users present in the database
+        for tenant in tenants:
+            self.create_tenant_vhost_and_user(tenant)
```
tehasdf: is it OK to run this for tenants that already had vhosts and users?

we already looked up existing users and vhosts, so if necessary we can only create the missing ones

mcouthon: I think that it's supposed to be harmless. One more API call shouldn't be an issue

jcollado: @tehasdf I've checked that it's not an error to try to create a vhost or a user that doesn't exist. However, let me see if I can improve this.

jcollado: @tehasdf What I've seen is that is very easy to know when a vhost or a user should be deleted because the tenant it should be linked to is no longer in the database. However, for each tenant a vhost and a user are created. Hence, if for some reason only one of them is missing, an attempt to create the other one will happen anyway with the current code, so I think I prefer to keep it simple given that trying to create something that already exists doesn't result in a failure.

tehasdf: cool :+1:

PR #889, opened 2017-06-16
CFY-6966 Fix recreate deployment envs

geokala:


geokala:
Requires:
https://github.com/cloudify-cosmo/cloudify-cli/pull/599
cloudify-cosmo/cloudify-rest-client#179

This approach was taken because:

    Trying to fake the context into acting as a different tenant is now considerably harder and every attempt simply resulted in deployment env recreation being run as the user who was running the snapshot.
    Trying to make this workflow run as a sub-workflow of the restore workflow failed because system wide workflows cannot be run when there are other workflows running (this error was given even when both workflows were specified not to be system wide).

It's likely that the best solution of this would be to make the create+restore snapshot work only on the current tenant, and have the CLI do the heavy lifting for taking snapshots for multiple tenants. This could have other benefits aside from making reasoning about and implementing security measures simpler.

geokala:
Approach changed, closing.

PR #898, opened 2017-06-22
CFY-6966 Fix deployment environment recreation

geokala:


geokala:
Depends on:
https://github.com/cloudify-cosmo/cloudify-agent/pull/265
https://github.com/cloudify-cosmo/cloudify-plugins-common/pull/338

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -250,11 +277,19 @@ def _get_snapshot_path(self):
...
-
-    def _get_existing_dep_envs(self):
-        return [dep.id for dep in self._client.deployments.list()]
+        return [self._get_plugin_archive_path_id_and_tenant(p)
+                for p in existing_plugins]
+
+    def _get_plugin_archive_path_id_and_tenant(self, plugin):
+        return {
+            'path': os.path.join(
+                '/opt/manager/resources/plugins',
```
idanmo: Isn't this path hard-coded somewhere you can import?

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -264,17 +299,26 @@ def _get_plugins_to_install(self, existing_plugins):
...
                          .format(len(all_plugins)))
-        plugins_to_install = [p for p in all_plugins if should_install(p)]
-        ctx.logger.debug('Found {0} plugins to install'
-                         .format(len(plugins_to_install)))
+        plugins_to_install = {}
+        for plugin in installable_plugins:
+            if should_install(plugin):
+                tenant = plugin['tenant']
+                plugins_to_install.setdefault(tenant, []).append(plugin)
+        ctx.logger.debug('Found plugins to install by tenant: {p}'
```
idanmo: Consider changing to info...

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -284,16 +328,23 @@ def _restore_plugins(self, existing_plugins):
...
+                    'Installing plugin {plugin} for {tenant}'.format(
+                        plugin=wagon_name,
+                        tenant=tenant,
+                    )
+                )
+                temp_plugin = os.path.join(plugins_tmp, wagon_name)
+                shutil.copyfile(plugin['path'], temp_plugin)
+                client.plugins.delete(plugin['id'], force=True)
+                client.plugins.upload(temp_plugin)
+                os.remove(temp_plugin)
```
idanmo: try..finally?

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -117,14 +110,55 @@ def restore(self):
...
         finally:
             ctx.logger.debug('Removing temp dir: {0}'.format(self._tempdir))
             shutil.rmtree(self._tempdir)
 
+    def _restore_deployment_envs(self):
+        deps = {}
+        for tenant in self._client.tenants.list():
+            ctx.logger.info(tenant)
+            tenant_name = tenant['name']
+            ctx.logger.info(tenant_name)
```
geokala: Sorry, that was output while I was debugging, not actually that useful as output now. I'm tearing it out.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -117,14 +110,55 @@ def restore(self):
...
+            ctx._context['tenant']['original_name'] = ctx.tenant_name
+            ctx._context['tenant']['name'] = tenant_name
```
geokala: I'm adding some more comments to explain what's going on here.

A context manager is an option but doesn't gain us a huge amount here since it's only used in this one place now, and the only way it can avoid having the context reset is if it raises an exception, which will mean the end of the execution anyway. Given that, I didn't feel like a context manager was worth using for this.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -117,14 +110,55 @@ def restore(self):
...
+            ctx._dep_contexts = None
+            deps[tenant_name] = ctx.deployments_contexts
```
geokala: Comments should explain this in the next commit.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -117,14 +110,55 @@ def restore(self):
...
+            ctx._context['tenant']['name'] = (
+                ctx._context['tenant']['original_name']
+            )
+            ctx._context['tenant'].pop('original_name')
+
+        for tenant, deployments in deps.iteritems():
+            ctx.logger.debug(
```
geokala: Info makes sense for this, I think, yes. Especially as it only appears in /var/log/mgmtworker/cloudify/logs/__system__.log

(which is something I'd like to fix but I don't think there's time for this release)

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -117,14 +110,55 @@ def restore(self):
...
+                'Restoring deployment environments for {tenant}'.format(
+                    tenant=tenant,
+                )
+            )
+            tenant_client = get_rest_client(tenant=tenant)
+            for deployment_id, dep_ctx in deployments.iteritems():
+                with dep_ctx:
+                    dep = tenant_client.deployments.get(deployment_id)
```
geokala: Coming soon to a commit near you.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
geokala: There appears to be a utils function I can use to generate it in the manager actually, yes- I'll switch to that.

geokala: Oh... no, that won't work, we don't have that util in mgmtworker.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
geokala: Done.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
geokala: Will do- I wasn't so concerned with stuff like that in this because a failed snapshot restore basically results in an inconsistent manager so it almost certainly needs burning down and starting again if anything about the restore fails.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
geokala: To add to this: The constants are in cloudify_manager too, so the only way to use them for this is to copy them into mgmtworker (only used in this one place), or to make it part of the mgmtworker env. I feel like neither of those are brilliant... maybe they should be in plugins-common?

geokala: (or maybe we should have manager-common since a bunch of stuff in plugins-common is manager related but not plugin related)

idanmo: Sounds like this should be a configuration propagated from manager blueprint doesn't it?

geokala: Possibly... though I'm not convinced that it's really configurable at the moment (meanig that changing it would just cause breakages), given that things are contained within the directories that are deployed (I think) by RPM, so it's likely that we'd find unexpected assumptions based on that.

So it's doable, but not if we want to be able to merge this today, and I'm not convinced it'll bring a great deal of value without more work across the manager and probably agent repos.

PR #900, opened 2017-06-27
CFY-7005 Parse timestamps as UTC values and remove timezone info

jcollado:
Passing timezone aware timestamp to PosgreSQL through SQLAlchemy doesn't
seem to work well in manual tests


jcollado:
@idanmo Sure, let me do it in the branch where I'm working on the integration tests.

jcollado:
The test case that reproduces the problem is in the commit that references this PR that can be seen above.

PR #904, opened 2017-07-02
CFY-6966 fix snapshots integration tests

idanmo:


tests/integration_tests/tests/agentless_tests/test_snapshot.py
```diff
@@ -306,32 +306,36 @@ def _get_snapshot(self, name):
...
+        execution = self._wait_for_execution_to_end(execution, rest_client)
         if execution.status == Execution.FAILED:
             self.logger.error('Execution error: {0}'.format(execution.error))
         self.assertEqual(Execution.TERMINATED, execution.status)
 
-    def _upload_and_validate_snapshot(self, snapshot_path, snapshot_id):
+    def _upload_and_validate_snapshot(self,
+                                      snapshot_path,
+                                      snapshot_id,
+                                      rest_client=None):
```
yarivgiga: The default value "None" is not necessary (because the function is only called once, with this parameter supplied) - and if it was necessary the case when it is None should be treated.

PR #905, opened 2017-07-02
fix validation for old snapshot restore

yarivgiga:


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -121,11 +121,12 @@ def restore(self):
 
     def _restore_deployment_envs(self):
         deps = {}
-        for tenant in self._client.tenants.list():
+        tenants = [get_tenant_name()] if self._snapshot_version < V_4_0_0 \
```
idanmo: For readability perhaps have this line set using a standard if statement.. you can do this after this PR is merged.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -383,11 +384,8 @@ def _restore_credentials(self, postgres):
 
     def _restore_agents(self):
         ctx.logger.info('Restoring cloudify agent data')
-        Agents().restore(
-            self._tempdir,
-            self._tenant_name,
-            is_older_than_4_1_0=self._snapshot_version < V_4_1_0,
-        )
+        Agents().restore(self._tempdir,
+                         is_older_than_4_1_0=self._snapshot_version < V_4_1_0)
```
idanmo: Since you have an if statement for checking whether there are tenants and whether the version is earlier than 410 then perhaps have these as methods with a meaningful name, for example:
if `tenants_are_available` etc..

PR #916, opened 2017-07-18
Agent install link endpoint

jcollado:
In this PR, the `agent-install-link` endpoint is added to the API.


jcollado:
Plugins are expected to use a function provided in the cloudify context. Hence, a new endpoint is not needed for this.

PR #929, opened 2017-08-08
CFY-7141 Fix community snapshot creds restore

geokala:


geokala:
System tests are passing.

PR #931, opened 2017-08-14
Cfy 7126 support okta

dusking:


rest-service/manager_rest/security/user_handler.py
```diff
@@ -47,6 +48,9 @@ def user_loader(request):
     if api_token:
         user, user_token_key = extract_api_token(api_token)
         return user
+    if current_app.okta.okta_saml_inside(request):
```
tehasdf: might need a check that `current_app.okta` isn't None in case of the community edition?

PR #932, opened 2017-08-15
CFY-7153 add api to retrieve current user info

yarivgiga:


rest-service/manager_rest/rest/resources_v3.py
```diff
@@ -283,6 +293,8 @@ def get(self, username, multi_tenancy):
         Get details for a single user
         """
         rest_utils.validate_inputs({'username': username})
+        if username == current_user.username:
```
dusking: comment - avoid auth flow for my info..

PR #934, opened 2017-08-16
CFY-7163 handle external auth

dusking:


rest-service/manager_rest/security/user_handler.py
```diff
@@ -48,7 +48,7 @@ def user_loader(request):
     if api_token:
         user, user_token_key = extract_api_token(api_token)
         return user
-    if current_app.okta and \
+    if current_app.okta and current_app.okta.is_configured and \
```
tehasdf: call()

mcouthon: Good catch, I missed it

PR #941, opened 2017-08-30
CFY-7200 Add a `networks` property to AgentConfig

mcouthon:


resources/rest-service/cloudify/types/types.yaml
```diff
@@ -537,6 +537,13 @@ data_types:
           'cloudify.nodes.Compute.cloudify_agent.name'.
         type: string
         required: false
+      networks:
+        description: >
+          A list of network names and IP addresses associated with them.
+          By default, there is only a "default" network, with the manager's
+          private IP associated with it. This network can be overwritten
+          (type: list)
```
tehasdf: why not use the `type: list` declaration for the property type, instead of putting this in the description ;P

PR #950, opened 2017-09-11
CFY-7188 Use a script to set manager SSL state

tehasdf:
Instead of calling `sed` from the restservice code, let's add
a script that will do the necessary config changes.
This way, we can also run that script from within cluster code,
and also we won't need to enable `sudo` for `sed`, only for this
specific script.


tehasdf:
The cluster part is used from cloudify-cosmo/cloudify-premium#260
The script is added in cloudify-cosmo/cloudify-manager-blueprints#658

PR #959, opened 2017-09-21
CFY-7265 Restore CA certs on cert restore

geokala:


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -163,26 +165,39 @@ def _restore_certificate(self):
...
-              'mv {new_p12} {old_p12}'.format(
-                time=time_to_wait_for_workflow_to_finish,
-                old_cert=old_cert,
-                old_key=old_key,
-                old_p12=old_p12,
-                new_cert=new_cert,
-                new_key=new_key,
-                new_p12=new_p12
-                )
+        cmd = (
```
tehasdf: maybe time to use a loop here? :P

geokala:
Agreed, that would be good. For reference of anyone else who might read this:
1. Not rebooting /should/ be straightforward, since we just need to figure out which order to restart services to minimise logged errors (ideally to avoid them entirely).

2. Not doing this as a bash script is a problem, because one of the things that needs restarting is the management worker, which is running the script. I think that's the primary problem, but it might also have issues if nginx restarts while it's trying to note what's happening. Ultimately, this probably requires an external script, maybe in combination with any work to provide the ability for users to update certs.

PR #961, opened 2017-09-25
CFY-7238 support customizable system-wide roles

yarivgiga:


rest-service/manager_rest/storage/storage_utils.py
```diff
@@ -39,14 +41,15 @@ def get_node(deployment_id, node_id):
 
 def create_default_user_tenant_and_roles(admin_username,
                                          admin_password,
-                                         amqp_manager):
+                                         amqp_manager,
+                                         authorization_file_path=None):
```
dusking: Seems that need to be without the "=None"

rest-service/manager_rest/storage/storage_utils.py
```diff
@@ -61,8 +64,14 @@ def create_default_user_tenant_and_roles(admin_username,
...
 
-def _create_roles():
-    for role in constants.ALL_ROLES:
+def _create_roles(authorization_file_path):
+    if authorization_file_path:
+        with open(authorization_file_path) as f:
+            content = load(f)
+        roles = [role['name'] for role in content['roles']]
+    else:
+        roles = constants.ALL_ROLES
```
dusking: Need to remove the ALL_ROLES const, and usage.. is not true anymore..

PR #964, opened 2017-09-26
CFY-7260 expose role conf

dusking:


rest-service/manager_rest/rest/resources_v3_1/manager_config.py
```diff
@@ -0,0 +1,50 @@
...
+                result['authorization'] = self._roles_config()
+        if not filter_by:
+            current_app.logger.debug('Retrieving roles without a filter')
+            result['authorization'] = self._roles_config()
+        return json.dumps(result)
+
+    @staticmethod
+    def _roles_config():
+        cfy_config = config.instance
+        roles = {
```
yarivgiga: maybe change that name because it's not only roles

PR #973, opened 2017-10-01
CFY-7257. Add roles column

jcollado:
Note: this is a work in progress

In this PR, the `users_tenant` secondary table is replaced with an association object that generates the same table with the following details:
- a new column `role_id` is added to the table
- a primary key constraint is generated to include `user_id` and `tenant_id`. Note that `role_id` is not part of the primary key because for now only one role is allowed per user and tenant.


jcollado:
Closing in favor of #978

PR #975, opened 2017-10-02
CFY-7328 Config: add the load_from_file method

tehasdf:
This allows loading the configuration passing a filename directly,
without the need to go through env vars. This will make it easier to
use the config from parts of the code that aren't run inside the
restservice.


tehasdf:
direct cause for this is that i want to use the storage_manager from cloudify-premium. Handling only the filenames is enough suffering, no need to require env vars on top of it as well.

PR #977, opened 2017-10-02
CFY-7328 Instantiate the app in a separate wsgi.py module

tehasdf:
This will allow importing CloudifyFlaskApp without invoking side
effects at import time.

gevent must now use manager_rest.wsgi:app
instead of manager_rest.server:app.


tehasdf:
This is a somewhat common pattern with flask, eg. http://flask.pocoo.org/docs/0.12/patterns/appfactories/#using-applications
Except that instead of having a function for an app factory, we use the app's init

PR #978, opened 2017-10-02
CFY-7257. Add roles column to secondary table

jcollado:
In this PR, a custom secondary table is created to add the new `role_id` column to the `users_tenants` table.

Related #973

### DIFF
--- a/rest-service/manager_rest/storage/management_models.py
+++ b/rest-service/manager_rest/storage/management_models.py
@@ -163,24 +163,16 @@ class User(SQLModelBase, UserMixin):
 
     @declared_attr
     def tenants(cls):
-        secondary_table = db.Table(
-            'users_tenants',
+        table_names = ['users', 'tenants', 'roles']
+        columns = [
             db.Column(
-                'user_id',
+                '{0}_id'.format(table_name[:-1]),
                 db.Integer,
-                db.ForeignKey('users.id')
-            ),
-            db.Column(
-                'tenant_id',
-                db.Integer,
-                db.ForeignKey('tenants.id')
-            ),
-            db.Column(
-                'role_id',
-                db.Integer,
-                db.ForeignKey('roles.id')
-            ),
-        )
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/storage/management_models.py
```diff
@@ -163,24 +163,16 @@ def groups(cls):
 
     @declared_attr
     def tenants(cls):
-        secondary_table = db.Table(
-            'users_tenants',
+        table_names = ['users', 'tenants', 'roles']
+        columns = [
             db.Column(
-                'user_id',
+                '{0}_id'.format(table_name[:-1]),
```
mcouthon: If we're doing this, why not just do `column_names = ['user', 'tenant', 'role']`?

rest-service/manager_rest/storage/management_models.py
```diff
@@ -174,12 +174,8 @@ def tenants(cls):
...
         constraint = db.PrimaryKeyConstraint(
             'user_id', 'tenant_id', name='users_tenants_pkey')
-        secondary_table = db.Table(
-            'users_tenants',
-            constraint,
-            *columns
-
-        )
+        args = columns + [constraint]
+        secondary_table = db.Table('users_tenants', *args)
```
mcouthon: Maybe we can do something like this?
```py
secondary_table = db.Table('users_tenants', *columns, constraint=constraint)
```
That way we won't need the awkward `args`

rest-service/manager_rest/storage/management_models.py
jcollado: The table name is also needed to generate the foreign key, so, despite I don't like much this, the other options are:
- generate the foreign key names from the column_names (similar to this change)
- have two different hardcoded lists (a little bit of duplication)

For now, I think I'm going to stick to the current solution since it's similar to the `many_to_many_relationship` implementation and it will make easier a future refactoring if needed.

PR #982, opened 2017-10-03
CFY-7257 Use association object instead of secondary table

jcollado:
In this PR, the original secondary table design is replaced with an association object.

This is because the secondary table doesn't provide a way to access the `role_id` column
at the ORM level and that is something that will be needed to add/remove tenants to a user. 

Note that back references are set to new attributes (`user_associations` and `tenant_associations`)
while association proxies are used to keep the old attributes (`users` and `tenants`).
This way, old code that uses those attributes don't need to be updated.

Also, there's a cascade parameter in the relationships that will be used in a separate PR
when removing users from a tenant.

### DIFF
--- a/rest-service/manager_rest/storage/management_models.py
+++ b/rest-service/manager_rest/storage/management_models.py
@@ -60,6 +60,12 @@ class Tenant(SQLModelBase):
     rabbitmq_username = db.Column(db.Text)
     rabbitmq_password = db.Column(db.Text)
 
+    user_associations = db.relationship(
+        'UserTenant',
+        back_populates='tenant',
+        cascade='all, delete-orphan',
+    )
+    users = association_proxy('user_associations', 'user')
     group_associations = db.relationship(
         'GroupTenantAssoc',
         back_populates='tenant',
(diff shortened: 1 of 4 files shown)


rest-service/manager_rest/storage/management_models.py
```diff
@@ -141,6 +149,13 @@ class User(SQLModelBase, UserMixin):
     password = db.Column(db.String(255))
     api_token_key = db.Column(db.String(100))
 
+    tenant_associations = db.relationship(
+        'UserTenant',
+        back_populates='user',
```
mcouthon: Should this be `user` or `users`?

jcollado: The back_populates value should match the attribute value defined in `management_models.UserTenant` which is called `user` because from the association object point of view this is a one-to-many relationship.

mcouthon: Ah, ok. This is the secondary table. Got it.

rest-service/manager_rest/storage/storage_utils.py
```diff
@@ -58,7 +58,12 @@ def create_default_user_tenant_and_roles(admin_username,
         password=encrypt_password(admin_password),
         roles=[admin_role]
     )
-    admin_user.tenants.append(default_tenant)
+    user_tenant = UserTenant(
```
mcouthon: Is this the only way to do it? I think you'll agree that the previous way was much more readable. Won't SQLA just take care of it?

jcollado: Since we want to set the tenant and the role in one shot, the association proxy won't be helpul in this case.

I can maybe rename `UserTenant` to `UserTenantAssoc` to make clearer that we are setting an association, not just a tenant. Would that be better?

PR #1001, opened 2017-10-11
CFY-7258. Set "user" role to "admin" user in "default_tenant"

jcollado:
In this PR, the `user` role is set to the `admin` user as a member of the `default_tenant`.

This is not really needed since the `sys_admin` system role will be effective anyway but provides more consistency since `user` is the default role for any user added to a tenant.


rest-service/manager_rest/storage/storage_utils.py
```diff
@@ -58,9 +58,15 @@ def create_default_user_tenant_and_roles(admin_username,
         password=encrypt_password(admin_password),
         roles=[admin_role]
     )
+
+    # User role assigned to admin user as a member of default tenant
+    # This is the default role when adding a user is added to a tenant.
+    # Anyway, `sys_admin` will be the effective role since is the system role.
+    user_role = user_datastore.find_role('user')
```
yarivgiga: i think we better have a const for DEFAULT_TENANT_ROLE = 'user'.
then in the future we may need some other stuff like DEFAULT_SYSTEM_ROLE = 'default' (like i did in cloudify-premium).

jcollado: Fixed in b8cae0c

PR #1002, opened 2017-10-11
CFY-7331. Set role description in database

jcollado:
In this PR, the role description is inserted into the database, if available, when roles are created from the  `authorization.conf` file.


jcollado:
As discussed offline, there's no need to have the description in the database if it's in the config file already.

PR #1005, opened 2017-10-15
CFY-7296 Add api to set a resource as global

inbala:


rest-service/manager_rest/resource_manager.py
```diff
@@ -1258,6 +1260,27 @@ def get_resource_availability(self, private_resource):
         return AvailabilityState.PRIVATE if private_resource \
             else AvailabilityState.TENANT
 
+    def set_global_availability(self, model_class, element_id):
+        # Check that there is no resource with the same name
+        if self.sm.count(model_class, element_id) > 1:
+            raise manager_exceptions.IllegalActionError(
+                "Can't set the availability of `{0}` to global because "
+                "there are resources with the same name".format(element_id))
+        else:
```
dusking: no need for else..

rest-service/manager_rest/resource_manager.py
```diff
@@ -1258,6 +1260,27 @@ def get_resource_availability(self, private_resource):
...
+            # Set the resource_availability to global
+            resource = self.sm.get(model_class, element_id)
+            resource.resource_availability = AvailabilityState.GLOBAL
+            resource.updated_at = utils.get_formatted_timestamp()
+            return self.sm.update(resource)
+
+    def is_modification_permitted(self, resource):
```
dusking: function name imply Boolean result.. need to be verify_ or something...

rest-service/manager_rest/rest/endpoint_mapper.py
```diff
@@ -28,6 +28,7 @@ def setup_resources(api):
         'Blueprints': 'blueprints',
         'BlueprintsId': 'blueprints/<string:blueprint_id>',
         'BlueprintsIdArchive': 'blueprints/<string:blueprint_id>/archive',
+        'BlueprintsGlobal': 'blueprints/<string:blueprint_id>/global',
```
dusking: if we only allow set global, and not unset, etc, maybe we should call it setglobal.. :)

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -0,0 +1,33 @@
...
+from manager_rest.rest import rest_decorators
+from manager_rest.security import SecuredResource
+from manager_rest.security.authorization import authorize
+from manager_rest.resource_manager import get_resource_manager
+
+
+class PluginsGlobal(SecuredResource):
+
+    @rest_decorators.exceptions_handled
+    @authorize('plugin_set_global')
```
dusking: maybe resource_set_global is sufficient?

PR #1020, opened 2017-10-18
CFY-7383 Restore admin user token correctly

geokala:


geokala:
This is to make the API token for admin users be restored correctly when a snapshot with the old hash salt is provided.

PR #1026, opened 2017-10-19
CFY-7350. Add endpoint to update user-tenant relationship

jcollado:
In this PR, an endpoint to update the user-tenant relationship is added.

Requires:
- cloudify-cosmo/cloudify-manager-blueprints#698
- cloudify-cosmo/cloudify-premium#293

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3/tenants.py
+++ b/rest-service/manager_rest/rest/resources_v3/tenants.py
@@ -146,7 +146,7 @@ class TenantUsers(SecuredMultiTenancyResource):
         )
         rest_utils.validate_inputs(request_dict)
         role_name = request_dict['role']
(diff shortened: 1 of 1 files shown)


rest-service/manager_rest/rest/resources_v3/tenants.py
```diff
@@ -128,6 +128,34 @@ def put(self, multi_tenancy):
...
+                    'type': unicode,
+                },
+                'role': {
+                    'type': unicode,
+                },
+            },
+        )
+        rest_utils.validate_inputs(request_dict)
+        role_name = request_dict['role']
+        rest_utils.validate_role_name(role_name)
```
dusking: Yariv changed the "validate_role_name" with "verify_role"
seems to update is function, since it should get both system and tenant roles..

PR #1029, opened 2017-10-20
CFY-7373. Add endpoint to update group-tenant association

jcollado:
In this PR, the endpoint to update the group-tenant association is added.

Requires:
- cloudify-cosmo/cloudify-manager-blueprints#699
- cloudify-cosmo/cloudify-premium#294

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3/tenants.py
+++ b/rest-service/manager_rest/rest/resources_v3/tenants.py
@@ -177,6 +177,34 @@ class TenantGroups(SecuredMultiTenancyResource):
         )
 
     @rest_decorators.exceptions_handled
+    @authorize('tenant_update_group', get_tenant_from='data')
+    @rest_decorators.marshal_with(TenantResponse)
+    @rest_decorators.no_external_authenticator('update group in tenant')
+    def patch(self, multi_tenancy):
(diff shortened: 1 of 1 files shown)


rest-service/manager_rest/rest/resources_v3/tenants.py
```diff
@@ -177,6 +177,34 @@ def put(self, multi_tenancy):
...
+                    'type': unicode,
+                },
+                'role': {
+                    'type': unicode,
+                },
+            },
+        )
+        rest_utils.validate_inputs(request_dict)
+        role_name = request_dict['role']
+        rest_utils.verify_role(role_name)
```
dusking: the default of verify_role is is_system_role=False
and we allow both system_wide and tenant_wide roles, so need to update the function..

jcollado: Why do we allow both system and tenant roles to update the role in the group tenants relationship? My understanding is that system roles are only allowed in the `users_roles` table and that for relationship tables (`users_tenants` and `groups_tenants`) only tenant roles are allowed. Is this correct?

Also, note that setting `is_system_role=True` here would be inconsistent with the `put` method a few lines above in which `is_sistem_role=False`.

PR #1031, opened 2017-10-22
CFY-7410 fix is administrator

dusking:


rest-service/manager_rest/storage/management_models.py
```diff
@@ -247,6 +247,21 @@ def _get_identifier_dict(self):
     def roles(cls):
         return many_to_many_relationship(cls, Role)
 
+    def have_role(self, tenant, list_of_roles):
```
tehasdf: maybe `has_role` ;P

rest-service/manager_rest/utils.py
```diff
@@ -182,9 +182,10 @@ def all_tenants_authorization():
     )
 
 
-def is_administrator():
+def is_administrator(tenant):
+    administrators_roles = \
+        config.instance.authorization_permissions['administrators']
```
tehasdf: wil `administrators` always be there?

dusking: in the conf file? yes

PR #1038, opened 2017-10-24
CFY-7419 fix download global blueprint

dusking:


mcouthon:
@dusking Be sure to fix flake error.

PR #1063, opened 2017-11-09
CFY-7509 If CA cert isn't in snapshot, use internal cert

tehasdf:


geokala:
Looks good, assuming it works as expected.

PR #1072, opened 2017-11-23
CFY-7562. Secret create update_if_exists

jcollado:
In this PR, the API endpoint to create a secret is updated to allow for an `upsert` flag to be passed in the json data. When the flag is set, if the secret already exists, it is updated.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3/secrets.py
+++ b/rest-service/manager_rest/rest/resources_v3/secrets.py
@@ -58,12 +58,13 @@ class SecretsKey(SecuredResource):
         rest_utils.validate_inputs({'key': key})
 
         sm = get_storage_manager()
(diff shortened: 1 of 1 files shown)


rest-service/manager_rest/rest/resources_v3/secrets.py
```diff
@@ -41,14 +42,39 @@ def put(self, key):
         """
         Create a new secret
         """
-        key, value = self._validate_secret_inputs(key)
+        request_dict = rest_utils.get_json_and_verify_params({
+            'value': {
+                'type': unicode,
+            },
+            'upsert': {
```
dusking: please consider changing `upsert` with `update_if_exists` or something..

jcollado: Fixed in 9560882

PR #1074, opened 2017-12-03
CFY-7584. Add basic way to get aria model storage

jcollado:
In this PR, function that creates an aria model storage that uses the same database as cloudify is added.

In an endpoint method that needs to exercise the aria mapi, the way to use it would be as follows:
```
from manager_rest.storage import aria_model
model_storage = aria_model.model_storage()

model_storage.<object>.<mapi_method>(*args, **kwargs)
```


rest-service/manager_rest/storage/aria_model.py
```diff
@@ -0,0 +1,40 @@
+import aria
+from aria.storage.sql_mapi import SQLAlchemyModelAPI
+
+from .models_base import db
+
+_model_storage = None
+
+
+def initiator():
```
mxmrlv: The kind of defeats the purpose - the initiator should be serializable. However in our case we don't really need it to be. However we could just pass the engine and session as part of the api_kwargs

PR #1077, opened 2017-12-06
Merge cloudify-manager-install

funkyHat:
It belongs in this repo


funkyHat:
This should still happen but there's no way this branch will apply cleanly now, so someone will have to redo it, I reckon.

PR #1078, opened 2017-12-06
CFY-7615 improve cli output regarding roles from groups

yarivgiga:


rest-service/manager_rest/storage/management_models.py
```diff
@@ -279,10 +314,39 @@ def roles_in_tenant(self, tenant):
...
+        """
+        return {
+            tenant_association.tenant.name: tenant_association.role.name
+            for tenant_association in self.tenant_associations
+        }
+
+    @property
+    def group_tenants(self):
+        """
+        Return all tenants associated with a user via groups,
```
dusking: Return all tenants user associated to..

PR #1079, opened 2017-12-06
New build_rpm.py script

funkyHat:
This script will be used to build the new style RPMs from spec files.

### DIFF
--- a/packaging/Vagrantfile
+++ b/packaging/Vagrantfile
@@ -27,6 +27,13 @@ GITHUB_PASSWORD = ENV['GITHUB_PASSWORD']
 CENTOS7_OFFICIAL_BOX_URL = 'http://cloud.centos.org/centos/7/vagrant/x86_64/images/CentOS-7.box'
 
 Vagrant.configure('2') do |config|
+  config.vm.define "builder" do |builder|
+    #dummy box, will be overriden
+    builder.vm.box = "centos/7"
+    builder.vm.provider :libvirt
+    builder.vm.provider :virtualbox
+  end
+
   config.vm.define "restservice" do |rest|
     #dummy box, will be overriden
     rest.vm.box = "dummy"
--- a/packaging/build_rpm.py
+++ b/packaging/build_rpm.py
@@ -141,49 +141,51 @@ def build_in_vagrant(source, spec_file_name):
     run this script inside a Vagrant box
     where `mock` is installed and configured
     """
-    check_call(['vagrant', 'up', 'builder'])
+    try:
+        check_call(['vagrant', 'up', 'builder'])
 
-    ssh_config = check_output(
-            ['vagrant', 'ssh-config', 'builder'])
-    with open(SSH_CONFIG_FILE, 'w') as f:
-        f.write(ssh_config)
+        ssh_config = check_output(
+                ['vagrant', 'ssh-config', 'builder'])
+        with open(SSH_CONFIG_FILE, 'w') as f:
+            f.write(ssh_config)
 
-    install_mock()
-    # sync the code
-    for local, remote in (
-            (source, 'source'),
-            (SCRIPT_DIR + '/', 'build_script')):
+        install_mock()
+        # sync the code
+        for local, remote in (
+                (source, 'source'),
+                (SCRIPT_DIR + '/', 'build_script')):
+            check_call(
+                    ['rsync', '-avz',
+                     '-e', 'ssh -F {}'.format(SSH_CONFIG_FILE),
+                     '--exclude', '.git',
+                     '--exclude', '*.sw[op]',  # vim swap files
+                     '--delete', '--delete-excluded',
+                     local,
+                     'builder:' + remote,
+                     ])
+
+        source = path_join('source', basename(source))
+
+        run_vagrant(
+                'python "{this_script}" "{spec_file}"'.format(
+                    this_script=path_join('build_script', SCRIPT_NAME),
+                    spec_file=path_join(source, 'packaging', spec_file_name),
+                ))
+
+        # Copy files back
+        final_rpm = run_vagrant(
+            "find /var/lib/mock/epel-7-x86_64/result "
+            "-name '*.noarch.rpm' -o -name '*.x86_64.rpm'",
+            _func=check_output,
+            ).strip()
         check_call(
-                ['rsync', '-avz',
-                 '-e', 'ssh -F {}'.format(SSH_CONFIG_FILE),
-                 '--exclude', '.git',
-                 '--exclude', '*.sw[op]',  # vim swap files
-                 '--delete', '--delete-excluded',
-                 local,
-                 'builder:' + remote,
-                 ])
+                ['scp', '-F', SSH_CONFIG_FILE,
+                 'builder:{rpm}'.format(rpm=final_rpm),
+                 '.'])
 
-    source = path_join('source', basename(source))
-
-    run_vagrant(
-            'python "{this_script}" "{spec_file}"'.format(
-                this_script=path_join('build_script', SCRIPT_NAME),
-                spec_file=path_join(source, 'packaging', spec_file_name),
-            ))
-
-    # Copy files back
-    final_rpm = run_vagrant(
-        "find /var/lib/mock/epel-7-x86_64/result "
-        "-name '*.noarch.rpm' -o -name '*.x86_64.rpm'",
-        _func=check_output,
-        ).strip()
-    check_call(
-            ['scp', '-F', SSH_CONFIG_FILE,
-             'builder:{rpm}'.format(rpm=final_rpm),
-             '.'])
-
-    # Power down the vagrant box
-    check_call(['vagrant', 'halt', 'builder'])
+    finally:
+        # Power down the vagrant box
+        check_call(['vagrant', 'halt', 'builder'])
 
 
 def get_rpmbuild_defines():
--- a/packaging/build_rpm.py
+++ b/packaging/build_rpm.py
@@ -47,11 +47,12 @@ logger = logging.getLogger(SCRIPT_NAME)
 
 def main(args):
     parser = argparse.ArgumentParser(
-            description="Build RPM using mock",
+            description="Build RPM using mock "
+            "(in a Vagrant box if you don't have mock installed locally)",
             )
     parser.add_argument(
             'spec_file', type=argparse.FileType('r'),
-            help="which RPM to build",
+            help="which RPM spec file to build",
             )
 
     args = parser.parse_args(args)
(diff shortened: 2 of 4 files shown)


packaging/Vagrantfile
```diff
@@ -23,31 +23,42 @@ PACKAGER_BRANCH = ENV['PACKAGER_BRANCH']
...
+-   path: /etc/sudoers.d/999-vagrant-cloud-init-requiretty
+    permissions: 440
+    content: |
+        Defaults:centos !requiretty
+)
 
 Vagrant.configure('2') do |config|
-  config.vm.define "restservice" do |rest|
+  config.vm.define "builder" do |builder|
     #dummy box, will be overriden
```
mcouthon: Is this still relevant?

tehasdf: that's the vm that'll be used by build_rpm if you want it to run vagrant (it has a mode of running a vagrant box, and a mode of running mock locally)

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+
+RPMS_DIR = SCRIPT_DIR
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(SCRIPT_NAME)
+
+
+def main(args):
+    parser = argparse.ArgumentParser(
+            description="Build RPM using mock",
```
mcouthon: I guess it can be vagrant as well, right?
Maybe we should even make it explicit?

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+            else:
+                raise
+        with open(outfile, 'wb') as out:
+            print('writing', outfile)
+            out.write(f.read())
+    finally:
+        f.close()
+
+
+def get_sources(spec_file):
```
mcouthon: Could use a bit of a docstring here. Not exactly sure what are those sources we're getting

mcouthon: Ok, I understand _what_ we're doing, but I'm not sure why. Why define something in a spec file if we're going to parse it like this then?

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+    spec_file_out = spec_file.replace('.spec', '.rendered.spec')
+    with open(spec_file_out, 'w') as f:
+        f.write(spec_content)
+
+    return spec_file_out
+
+
+def run_vagrant(cmd, *args, **kwargs):
+    """Run command on Vagrant box"""
+    func = kwargs.pop('_func', check_call)
```
mcouthon: Why do it that way? As far as I can see, you're always using `check_call` except one time where you do `check_output`. Couldn't you have just passed an explicit flag (e.g. show/don't show output).
Unless I'm missing uses of this.

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+                ['vagrant', 'ssh-config', 'builder'])
+        with open(SSH_CONFIG_FILE, 'w') as f:
+            f.write(ssh_config)
+
+        install_mock()
+        # sync the code
+        for local, remote in (
+                (source, 'source'),
+                (SCRIPT_DIR + '/', 'build_script')):
+            check_call(
```
mcouthon: This might've used its own function

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+                ))
+
+        # Copy files back
+        final_rpm = run_vagrant(
+            "find /var/lib/mock/epel-7-x86_64/result "
+            "-name '*.noarch.rpm' -o -name '*.x86_64.rpm'",
+            _func=check_output,
+            ).strip()
+        check_call(
+                ['scp', '-F', SSH_CONFIG_FILE,
```
mcouthon: Do we assume that the box has scp/rsync? Does the base image promises this to us?

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+                 'builder:{rpm}'.format(rpm=final_rpm),
+                 '.'])
+
+        # Power down the vagrant box
+        check_call(['vagrant', 'halt', 'builder'])
```
mcouthon: Maybe we need a big try except here, with a finally that halts the box? Just so we don't leave VMs running if something fails.

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+    for url in sources:
+        download_if_newer(url)
+
+    chdir(SCRIPT_DIR)
+
+    if has_cmd('mock'):
+        # run locally
+        rpm = build(source, spec_file_name)
+        shutil.copy(rpm, SCRIPT_DIR)
+    elif has_cmd('vagrant'):
```
mcouthon: I'd put all of this in a `build_in_vagrant` function (which I would also split into 3-5 parts) - it's a pretty large and self contained piece of logic

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+    Apply the defines to the spec file. We don't use `--define` because
+    `mock`'s `--rpmbuild-opts` doesn't apply to `--buildsrpm`
+
+    Returns the path to the rendered spec file
+    """
+    with open(spec_file) as f:
+        spec_content = f.read()
+
+    for k, v in defines.items():
+        spec_content = spec_content.replace('%{' + k + '}', v)
```
mcouthon: Why is this necessary? Why don't we just put the version info inside the spec files initially? I mean `CLOUDIFY_PACKAGE_RELEASE` doesn't have to be the same for all of the packages, right?

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+        return
+
+    logger.info(
+            'installing dependencies from <{file}>. RPMs dir: {dir}'.format(
+                file=DEPENDENCIES_FILE.format(spec_file=spec_file),
+                dir=RPMS_DIR,
+                ))
+
+    rpms = {}
+    for f in listdir(RPMS_DIR):
```
mcouthon: I'm not entirely sure what each of the two for loops does. The first one goes over the current folder and looking for rpms to install, and the second loop looks for dependencies from the .spec.dependencies file?
Maybe add some more comments?

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+        f = f.strip()
+        if any(f.endswith(y) for y in ('.x86_64.rpm', '.noarch.rpm')):
+            name = check_output([
+                'rpm', '-qp', '--queryformat', '%{NAME}', f,
+                ])
+            rpms[name] = f
+
+    logger.info(('found RPMs:', rpms))
```
mcouthon: Why the tuple in here? :)

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+            package_name = dep
+            install_package = rpms[dep]
+
+        check_call([MOCK, '--yum-cmd', 'remove', package_name])
+        check_call([
+            MOCK, '--yum-cmd', 'install', install_package,
+            ])
+
+
+def get_dependency_urls(spec_file):
```
mcouthon: Let's add a docstring to explain here that we're only getting here dependencies that need to be downloaded

packaging/build_rpm.py
mcouthon: Why don't we just go ahead and download all the necessary dependencies in advance and put them all in the same folder, and just install everything in there?

packaging/build_rpm.py
funkyHat: During testing I've actually found it useful that the box is left up if the build fails, but I suppose always halting is probably better (there will only ever be one builder VM)

packaging/build_rpm.py
funkyHat: The first loop finds the RPMs and extracts the package names from them (i.e. `cloudfy-rest-service` from `cloudify-rest-service-4.3.blahblah.rpm`)
The second loop installs the RPMs it needs. Because the RPMs directory is the build script directory, there will be multiple RPMs sitting here on local runs, and we don't want to install all of them
(actually this should probably be extended to install the correct version if there are multiple with the same name)

packaging/build_rpm.py
funkyHat: logger wasn't behaving how I thought it would with just 2 arguments. Perhaps this should be updated to present the list of RPMs more neatly anyway

packaging/build_rpm.py
funkyHat: Because defining `Source*:` in the spec file is the "right" way to do it. I'm probably twisting the definition of "right" by also dumping the repo tree in `SOURCES`, but if at some point we decide to ship proper `.src.rpm`s this will be one less thing to fix, at least

packaging/build_rpm.py
tehasdf: it used to be an explicit `--local` for non-vagrant. Hmm. Personally i think i'd prefer the vagrant part to be a separate script

packaging/build_rpm.py
tehasdf: (note that vagrant is mostly only useful for developers' machines, jenkins etc all use mock directly)
we do assume that it does, same as we assume it has vagrant. These reqs should be documented (when this script grows docs)

packaging/build_rpm.py
tehasdf: it doesn't need to be the same, but until we actually have different versions in each package, i think it makes sense to have this, as opposed to updating every single spec file

packaging/build_rpm.py
```diff
@@ -0,0 +1,375 @@
...
+
+    Returns the path to the rendered spec file
+    """
+    with open(spec_file) as f:
+        spec_content = f.read()
+
+    for k, v in defines.items():
+        spec_content = spec_content.replace('%{' + k + '}', v)
+
+    spec_file_out = spec_file.replace('.spec', '.rendered.spec')
```
tehasdf: let's make this a tempfile, because writing next to the .spec just clutters `git status`, and (so far) it's never useful to examine the rendered spec anyway

packaging/build_rpm.py
tehasdf: `logger.info('found RPMs: %s', rpms)` is the right syntax

packaging/build_rpm.py
funkyHat: You're right we only use `_func=` once, but I'm not sure that makes it better or worse than a flag?

packaging/build_rpm.py
funkyHat: Replaced with a log line for each found RPM instead. The printed dict looked ugly anyway

packaging/build_rpm.py
funkyHat: I'm not super keen on hiding the rendered file in a tmpdir somewhere.
Added an entry to .gitignore. Shout if you object

packaging/build_rpm.py
funkyHat: @tehasdf so that I have to run a different command to you on my dev machine vs your dev VM?
My aim is to have a system test runner script (or plugin, whatever) (when the system tests get merged into `cloudify-manager`) which can do the RPM builds as necessary and then run the tests. I don't want users of that script (or even the script itself) to need to worry about whether vagrant is a requirement or not.
Also everything that happens in `main()` before the mock/vagrant condition makes more sense to run on the host if there is a host.

packaging/build_rpm.py
funkyHat: We don't download all in advance because some of what we want to install should come from another one of these spec files.
Doing it this way means I can do a local build of `-rest-service` and then the RPM I need is saved ready for a `-premium` build, no upload/download required. In Jenkins we achieve the same by copying artifacts as you know

packaging/build_rpm.py
funkyHat: made the `build_in_vagrant()` function but I'm not convinced splitting that out further will make it more readable.
It comfortably fits on a page (but OK, my font size is small...), and all of it that's not already factored out seems fairly self-explanatory to me.
LMK if this doesn't convince you and I'll have a go

packaging/build_rpm.py
funkyHat: We don't, but the failure mode is it'll fail with a message like `command not found: rsync` before doing anything. This is intended for developer use only so I'm not worried too much about tracebacks when things go wrong.

PR #1082, opened 2017-12-12
CFY-7593 listing all tenants should not require tenant for authorization

yarivgiga:


rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -210,3 +210,8 @@ def verify_role(role_name, is_system_role=False):
             'Role `{0}` is a {1} and cannot be assigned as a {2}'
             .format(role_name, role['type'], expected_role_type)
         )
+
+
+def check_all_tenants():
```
dusking: request_use_all_tenants

PR #1086, opened 2017-12-13
CFY-7650. Get aria resource storage in a synced directory

jcollado:
In this PR, the ability to create an aria resource storage is added.

Usage example:
```
>>> from manager_rest.storage import aria_resource
>>> aria_resource.resource_storage()
ResourceStorage(api=<class 'aria.storage.filesystem_rapi.FileSystemResourceAPI'>)
```

Requires: cloudify-cosmo/cloudify-plugins-common#371

### DIFF
--- a/rest-service/manager_rest/storage/aria_resource.py
+++ b/rest-service/manager_rest/storage/aria_resource.py
@@ -1,0 +1,27 @@
+import aria
+from aria.storage.filesystem_rapi import FileSystemResourceAPI
+
+# Use a synced directory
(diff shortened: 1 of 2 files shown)


tehasdf:
this directory also needs to be added to snapshots, yes?

jcollado:
@tehasdf Good catch! I've added the aria directory to the list of directories to include in the snapshot and tried to create/restore a snapshot with a file under the aria directory and worked as expected.

PR #1088, opened 2017-12-14
CFY-7660 all_tenants filter should consider roles

yarivgiga:


rest-service/manager_rest/utils.py
```diff
@@ -184,6 +187,14 @@ def all_tenants_authorization():
     )
 
 
+def tenant_specific_authorization(tenant, resource_name, action='list'):
```
dusking: comment.. verify user have permission to exec this action in current tenant on that resource..

PR #1089, opened 2017-12-16
build_rpm.py: only attempt downloading http sources

tehasdf:
There's no point trying to download things that aren't URLs. Those
will simply be assumed to already be there. (eg. moved by jenkins)

### DIFF
--- a/packaging/build_rpm.py
+++ b/packaging/build_rpm.py
@@ -81,7 +81,8 @@ def main(args):
     sources += get_dependency_urls(spec_file)
     logger.info(sources)
     for url in sources:
-        download_if_newer(url)
+        if url.startswith('http')
+            download_if_newer(url)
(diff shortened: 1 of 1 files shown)


tehasdf:
indeed it does, but only for dependencies, not for sources

mcouthon:
In that case, would that make `get_dependency_urls` unnecessary?

packaging/build_rpm.py
```diff
@@ -81,7 +81,8 @@ def main(args):
     sources += get_dependency_urls(spec_file)
     logger.info(sources)
     for url in sources:
-        download_if_newer(url)
+        if url.startswith('http')
```
mcouthon: Is there no better way to determine this?

geokala:

PR #1090, opened 2017-12-17
CFY-7545 new cloudify-rest-service RPM

funkyHat:


rest-service/setup.py
```diff
@@ -73,4 +73,8 @@
     license='LICENSE',
     description='Cloudify manager rest service',
     zip_safe=False,
-    install_requires=install_requires)
+    install_requires=install_requires,
+    extras_require={
+        'dbus': ['dbus-python==1.2.4'],
```
mcouthon: What's that for?

.circleci/config.yml
```diff
@@ -0,0 +1,149 @@
+version: 2
```
mcouthon: Does this new file remove the need for `run-tests.sh`?

packaging/cloudify-rest-service.spec
```diff
@@ -84,7 +84,7 @@ systemd-tmpfiles --create
 /etc/sudoers.d/cloudify-restservice
 /opt/restservice/NOTICE.txt
 /opt/restservice/set-manager-ssl.py*
-/usr/lib/systemd/system/cloudify-rest-service.service
+/usr/lib/systemd/system/cloudify-restservice.service
```
mcouthon: Probably applies to the logrotate file as well?

rest-service/setup.py
tehasdf: dbus-python is used for `cfy status`; currently we do a couple `ln` calls to use the systemwide install of it, but it is nicer to simply install it into the venv instead
(current: https://github.com/cloudify-cosmo/cloudify-manager-install/blob/09472e7b91f98873925ba88ef9590cbca5ace540/cfy_manager/components/restservice/restservice.py#L102 )

.circleci/config.yml
funkyHat: yes

packaging/cloudify-rest-service.spec
funkyHat: tbh I'd rather have them matching the package name as much as possible. This one has to be like this or else requires a bunch of changes elsewhere (cfy status etc.)
The name of the logrotate file is essentially arbitrary

tehasdf:
The `upload_to_s3.py` stuff was moved out to #1098

tehasdf:
the upload_to_s3 stuff was moved back in here after all ;p

PR #1107, opened 2017-12-26
CFY-7726: add start, stop and restart workflows

isaac-s:


rest-service/manager_rest/resource_manager.py
```diff
@@ -722,10 +722,6 @@ def validate_plugin_is_installed(self, plugin):
         if not - raises an appropriate exception.
         :param plugin: A plugin from the blueprint
         """
-        # if the plugin is 'diamond' we don't want it to be installed
```
mcouthon: I agree that this needs to happen, but probably not in this PR

isaac-s: I don't know how this change made it here.
It belongs in another JIRA that I'm sure I had already merged. I'll check.

isaac-s: @mcouthon looks like this hasn't been approved/merged yet. Can you approve? https://github.com/cloudify-cosmo/cloudify-manager/pull/1064

isaac-s:
@mcouthon @tehasdf how about now?

isaac-s:
@tehasdf they're needed because they're all optional (assuming you're talking about `type_names`, `node_ids` and `node_instance_ids`). User may specify either of these three filters, or a subset of them. Or they can specify none at all.

tehasdf:
Reverted in #1111

PR #1111, opened 2017-12-27
Revert "CFY-7726: add start, stop and restart workflows"

isaac-s:
Reverts cloudify-cosmo/cloudify-manager#1107


tehasdf:
Reverting until the original PR can be properly tested

PR #1113, opened 2017-12-28
CFY-7630 new cloudify-rabbitmq RPM

funkyHat:


packaging/cloudify-rabbitmq.spec
```diff
@@ -0,0 +1,46 @@
+
+Name:           cloudify-rabbitmq
+Version:        %{CLOUDIFY_VERSION}
+Release:        %{CLOUDIFY_PACKAGE_RELEASE}%{?dist}
+BuildArch:      noarch
+Summary:        Cloudify's RabbitMQ configuration
+Group:          Applications/Multimedia
+License:        Apache 2.0
+URL:            https://github.com/cloudify-cosmo/cloudify-amqp-influxdb
```
mcouthon: @funkyHat That's probably not the right URL. I wonder if it's the same in the other specs?

funkyHat: whoops https://github.com/cloudify-cosmo/cloudify-manager/pull/1135

PR #1114, opened 2017-12-29
CFY-7726: added start, stop and restart workflows

isaac-s:


tehasdf:
please make the tests passing:
https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/test/endpoints/test_deployments.py#L191
this assertion now gets 11, instead of 8, with the additional workflows

PR #1116, opened 2017-12-31
CFY-7735 Get edition from the config file

tehasdf:
Instead of using the VERSION file to encode the version/edition
at build time, we'll be using the config file for the edition
and we'll query the actually installed version.


tehasdf:
depends on cloudify-cosmo/cloudify-manager-install#91

PR #1118, opened 2018-01-02
CFY-7620 Locations of stage scripts from the RPM

tehasdf:


tehasdf:
Depends on cloudify-cosmo/cloudify-manager-install#92

workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -152,7 +152,8 @@ def restore_stage_files(archive_root, override=False):
             user=snapshot_constants.STAGE_USER,
         )
         restore_command = [snapshot_constants.STAGE_RESTORE_SCRIPT,
-                           stage_tempdir]
+                           stage_tempdir,
+                           '--homedir', snapshot_constants.STAGE_BASE_FOLDER]
```
tehasdf: passing homedir here allows us to not need to render the script

PR #1120, opened 2018-01-02
CFY-7700 Translate plugin yaml urls to local urls

tehasdf:


tehasdf:
This makes so that a plugin can be used only if the uploading user can use it. It doesn't take into account the blueprint visibility settings (so, it is possible to use a tenanted plugin yaml with a global blueprint)

tehasdf:
Usage example is:
```
imports:
  - plugin:cloudify-openstack-plugin:2.0.1
```

rest-service/manager_rest/app_context.py
```diff
@@ -19,23 +43,66 @@ def get_parser_context(sm=None):
...
+        'resolver': resolver,
+        'validate_version': cloudify_section.get(
+            constants.VALIDATE_DEFINITIONS_VERSION, True)
     }
+
+
+class ResolverWithPlugins(DefaultImportResolver):
+    """A resolver which translates plugin-style urls to file:// urls.
+
+    The URL: `file://plugins/cloudify-openstack-plugin/2.0.1` will be
```
yarivgiga: are u sure about the `file://` prefix?
i was under the impression it was gonna be just:
/plugins/cloudify-openstack-plugin/2.0.1

rest-service/manager_rest/app_context.py
```diff
@@ -19,23 +43,66 @@ def get_parser_context(sm=None):
...
+    translated to: `file:///opt/manager/resources/plugins/<id>`, where <id>
```
yarivgiga: u should also add a suffix of the yaml file (not necessarily plugin.yaml, it can be any yanl file)

rest-service/manager_rest/app_context.py
```diff
@@ -19,23 +43,66 @@ def get_parser_context(sm=None):
...
+    is the id of the plugin looked up for the current tenant.
+
+    Both the version and the filename are optional.
+    """
+    PREFIX = 'file://plugins/'
```
yarivgiga: again, we should check the design, maybe the import just starts with "/plugins"

rest-service/manager_rest/app_context.py
```diff
@@ -19,23 +43,66 @@ def get_parser_context(sm=None):
...
+        return sm.get(Plugin, element_id=None, filters=filters)
+
+    def _make_plugin_url(self, plugin_id):
+        plugin_path = os.path.join(
+            config.instance.file_server_root,
+            FILE_SERVER_PLUGINS_FOLDER,
+            plugin_id)
+        yaml_files = glob.glob(os.path.join(plugin_path, '*.yaml'))
+        if len(yaml_files) != 1:
+            return None
```
yarivgiga: maybe raising exception?

tehasdf: yep now that we changed the `file://` to `plugin:`, it doesn't make sense to return None because there's no way that could be an existing file, thanks

rest-service/manager_rest/app_context.py
```diff
@@ -19,23 +43,66 @@ def get_parser_context(sm=None):
...
+            resolved = self._resolve_plugin_url(import_url)
+            if resolved:
+                import_url = resolved
+        return super(ResolverWithPlugins, self).fetch_import(import_url)
+
+    def _is_plugin_url(self, import_url):
+        return import_url.startswith(self.PREFIX)
+
+    def _resolve_plugin_url(self, import_url):
+        parts = import_url.replace(self.PREFIX, '').split(':')
```
dusking: shouldn't it be replace(self.PREFIX, '', 1)?
otherwise, "plugin:cloudify-openstack-plugin:2.0.1" will become: "cloudify-openstack-2.0.1"..

rest-service/manager_rest/app_context.py
```diff
@@ -19,23 +43,66 @@ def get_parser_context(sm=None):
...
+        return self._make_plugin_url(plugin.id)
+
+    def _find_plugin(self, name, version=None):
+        filters = {'package_name': name}
+        if version is not None:
+            filters['package_version'] = version
+        sm = get_storage_manager()
+        return sm.get(Plugin, element_id=None, filters=filters)
+
+    def _make_plugin_url(self, plugin_id):
```
dusking: maybe we should indicate that it returns the plugin_yaml_url...

rest-service/manager_rest/app_context.py
```diff
@@ -19,23 +43,66 @@ def get_parser_context(sm=None):
...
+        if self._is_plugin_url(import_url):
+            resolved = self._resolve_plugin_url(import_url)
+            if resolved:
+                import_url = resolved
+        return super(ResolverWithPlugins, self).fetch_import(import_url)
+
+    def _is_plugin_url(self, import_url):
+        return import_url.startswith(self.PREFIX)
+
+    def _resolve_plugin_url(self, import_url):
```
dusking: maybe _resolve_plugin_yaml_url ?

rest-service/manager_rest/app_context.py
```diff
@@ -19,23 +43,66 @@ def get_parser_context(sm=None):
...
+    is the id of the plugin looked up for the current tenant.
+
+    Both the version and the filename are optional.
+    """
+    PREFIX = 'plugin:'
+
+    def fetch_import(self, import_url):
+        if self._is_plugin_url(import_url):
+            resolved = self._resolve_plugin_url(import_url)
+            if resolved:
```
dusking: what should happen it it's a plugin url, but we failed to resolve it? (missing yaml file)...
we probably should fail it..

PR #1121, opened 2018-01-02
CFY-7514 override dist macro remove .centos

funkyHat:


packaging/build_rpm.py
```diff
@@ -336,22 +338,46 @@ def install_mock():
...
+
+    config_opts = {}
+    config = {'config_opts': config_opts}
+    exec mock_config in config, config
+
+    options = [
+        "config_opts['rpmbuild_networking'] = True",
+        "config_opts['macros']['%dist'] = '.el7'",
+    ]
+    for option in reversed(options):
```
mcouthon: Why not just change the order of options in the first place, instead of using `reversed`?

packaging/build_rpm.py
```diff
@@ -336,22 +338,46 @@ def install_mock():
...
+    (we have to download packages from pypi)
+    """
+    mock_config = check_output(MOCK + ['--debug-config', '-qq'])
+
+    config_opts = {}
+    config = {'config_opts': config_opts}
+    exec mock_config in config, config
+
+    options = [
+        "config_opts['rpmbuild_networking'] = True",
```
mcouthon: Why are we parsing those as strings, instead of having a dict? E.g.
```
options = {
  "config_opts['rpmbuild_networking']": "True",
  "config_opts['macros']['%dist']": "'.el7'",
}
```

packaging/build_rpm.py
```diff
@@ -336,22 +338,46 @@ def install_mock():
...
+        "config_opts['macros']['%dist'] = '.el7'",
+    ]
+    for option in reversed(options):
+        lhs, eq, rhs = option.partition(' = ')
```
mcouthon: See above comment about using dicts here

packaging/build_rpm.py
mcouthon: If it's just a matter of writing them to file later on, a simple `{opt} = {value}` template will do

packaging/build_rpm.py
```diff
@@ -336,22 +338,46 @@ def install_mock():
...
-                )
+    """
+    allow network access during the build and set the %dist macro
+    (we have to download packages from pypi)
+    """
+    mock_config = check_output(MOCK + ['--debug-config', '-qq'])
+
+    config_opts = {}
+    config = {'config_opts': config_opts}
+    exec mock_config in config, config
```
mcouthon: Can you explain again the need for the `exec` statement? I’m not sure I’m following exactly what/why you’re doing.

tehasdf: this is because mock config files are python code, so the way to parse mock config files is to use `exec`. This is unfortunate but that's a design choice by mock, and they are doing the same as well: https://github.com/rpm-software-management/mock/blob/7f363b1c15d5cc8e893fdf495713515df94876c7/mock/py/mockbuild/util.py#L1182

packaging/build_rpm.py
```diff
@@ -336,22 +338,46 @@ def install_mock():
...
+        if not eq:
+            raise ValueError('malformed option', option)
+
+        try:
+            exists = eval('{} == {}'.format(lhs, rhs), config, config)
+        except KeyError:
+            exists = False
+
+        if exists:
+            options.remove(option)
```
tehasdf: please make a new list instead and add filtered options to it. This is too easy to accidentally make removing from the list we're iterating over (which would happen if we eg. removed the `reversed`)

packaging/build_rpm.py
```diff
@@ -336,22 +338,46 @@ def install_mock():
...
+        "config_opts['rpmbuild_networking'] = True",
+        "config_opts['macros']['%dist'] = '.el7'",
+    ]
+    for option in reversed(options):
+        lhs, eq, rhs = option.partition(' = ')
+        if not eq:
+            raise ValueError('malformed option', option)
+
+        try:
+            exists = eval('{} == {}'.format(lhs, rhs), config, config)
```
tehasdf: do we actually need to check? i think we can simply append our config values at the end of the config file, not conditionally, without any checking. I do realize it's not as nice, but in the realistic case, it should be good enough? and it's much simpler

packaging/build_rpm.py
funkyHat: The order isn't important, using reversed allows us to safely delete elements from the list we're iterating over

packaging/build_rpm.py
funkyHat: see above. Could rework this to put the missing ones in a list rather than using one list if you prefer

packaging/build_rpm.py
funkyHat: That means on the mock builder this file will continuously grow, until eventually in a year's time the disk is full and it breaks

packaging/build_rpm.py
funkyHat: yeah ok :(

funkyHat: (I'm unconvinced by the "it might break if we break it" argument, but using a separate list is kind of neater anyway I guess :)

PR #1125, opened 2018-01-03
CFY-7755 Fix diamond plugin use

geokala:


rest-service/manager_rest/resource_manager.py
```diff
@@ -726,6 +726,13 @@ def validate_plugin_is_installed(self, plugin):
         :param plugin: A plugin from the blueprint
         """
 
+        if plugin['package_name'] == 'cloudify-diamond-plugin':
```
isaac-s: I am against hard-coding this.
It'd be best to add a field to the `plugin` DSL schema, something like `agent_only: True/False`.

tehasdf: I prefer we avoided that, because that suggests we have some kind of agent plugins feature, and we essentially don't. And the diamond plugin is a weird exception.
I think having it hardcoded here for now, and getting rid of the plugin altogether in the future (by eg. making the diamond plugin part of agent proper) is a better solution.

Earlier, I suggested that @geokala puts this in a `_is_agent_plugin`-like function, but they didn't want to do it mostly because of the same reason.

@isaac-s does that make sense, or do you think we should go through with a schema change indeed?

isaac-s: >  because that suggests we have some kind of agent plugins feature, and we essentially don't

Well, not necessarily so.
First, very few people actually read the `plugin.yaml` file.
Second, you can claim this about any open-source API we publish, including internal API's.

I'm not the one making the decision... but as far as I'm concerned, hard-coding artifact names is a big no-no. Maybe the flag name should change (something other than `agent_only`. Maybe `agent_bundled`?).

geokala: Two things on this:
1. Yes, if we're going to support agent plugins being installed on the agents then we should probably provide something in the schema that supports this. Until we actually have a plan for how to properly support it I don't think it's really worth us implementing it.

2. Even if we do come up with some nice way of doing it, we still need to hard code the name for backwards compatibility, otherwise we'll fail noisily and with unhelpful error messages for any older blueprints that users try to use (including ones that they bring with them on an upgrade.

PR #1127, opened 2018-01-03
CFY-7757 fix zip file detection

dusking:


rest-service/manager_rest/test/endpoints/test_plugins.py
```diff
@@ -233,6 +233,23 @@ def test_plugin_upload_progress(self):
         finally:
             self.quiet_delete(tmp_file_path)
 
+    @attr(client_min_version=3,
+          client_max_version=base_test.LATEST_API_VERSION)
+    def test_plugin_upload_without_yaml(self):
+        tmp_file_path = self.create_wheel('wagon', '0.6.1')
+        total_size = os.path.getsize(tmp_file_path)
+
+        progress_func = generate_progress_func(
```
tehasdf: the progress_func can likely just be remoed, because it is only used in this file for tests that test progress..

PR #1128, opened 2018-01-03
CFY-7699 fix the yaml url path prefix from /plugins to plugin:

yarivgiga:


tehasdf:
This is related to cloudify-cosmo/cloudify-manager#1120

PR #1136, opened 2018-01-06
CFY-7619 Agents RPM spec

tehasdf:


tehasdf:
There should be a better way instead of curl in the spec, but i'm not yet sure what is it

packaging/cloudify-agents.spec
```diff
@@ -0,0 +1,45 @@
...
+BuildRequires:  python >= 2.7
+Requires(pre):  shadow-utils
+
+
+%description
+Cloudify Agent packages
+
+
+%prep
+curl https://raw.githubusercontent.com/cloudify-cosmo/cloudify-versions/master/packages-urls/agent-packages.yaml > /tmp/agents-list.txt
```
funkyHat: If you don't like `curl` here it should be possible to specify this URL as `Source0:` and have `build_rpm.py` download it to ${RPM_SOURCE_DIR} for you

funkyHat: & you can then refer to it as `%{S:0}` (see cloudify-premium's spec for an example)

funkyHat: I PRd your PR https://github.com/cloudify-cosmo/cloudify-manager/pull/1146

PR #1139, opened 2018-01-08
fix integration test: add workflows to test_deployment_workflows

yarivgiga:


tehasdf:
related to cloudify-cosmo/cloudify-manager#1114

PR #1141, opened 2018-01-09
CFY-7753 Fix the file server validation of a global blueprint

inbala:


rest-service/manager_rest/rest/resources_v3/manager.py
```diff
@@ -53,8 +53,7 @@ def _verify_tenant(uri):
         # the header
         for resource in tenanted_resources:
             if uri.startswith(resource):
-                uri = uri.replace(resource, '', 1).strip('/')
-                uri_tenant, _ = uri.split('/', 1)
+                _, uri_tenant = uri.split('/', 2)[:2]
```
dusking: comment for uri sample.

PR #1147, opened 2018-01-10
CFY-7618 new cloudify logstash RPM

funkyHat:


packaging/cloudify-logstash.spec
```diff
@@ -0,0 +1,99 @@
...
+Requires:       jre, postgresql94-jdbc
+Conflicts:      logstash
+
+Source0:        http://repository.cloudifysource.org/cloudify/components/logstash-output-jdbc-0.2.10.gem
+Source1:        http://repository.cloudifysource.org/cloudify/components/logstash-filter-json_encode-0.1.5.gem
+
+%define _user logstash
+
+# Work around lurking stale hashbangs from logstash's build process
+%define __requires_exclude ^/home/jenkins/.*
```
tehasdf: shouldn't this be `__requires_exclude_from`? Just `__requires_exclude` makes it so that files in there won't be requirements; `_from` makes it so that files in there won't be scanned

packaging/cloudify-logstash.spec
```diff
@@ -0,0 +1,99 @@
+
+Name:           cloudify-logstash
+Version:        %{CLOUDIFY_VERSION}
+Release:        %{CLOUDIFY_PACKAGE_RELEASE}%{?dist}
+Summary:        Cloudify's Logstash
+Group:          Applications/Multimedia
+License:        Apache 2.0
+URL:            https://github.com/cloudify-cosmo/cloudify-amqp-influxdb
```
tehasdf: url

packaging/cloudify-logstash.spec
```diff
@@ -0,0 +1,99 @@
...
+
+%pre
+
+# create logstash group
+if ! getent group logstash >/dev/null; then
+  groupadd -r logstash
+fi
+
+# create logstash user
+if ! getent passwd logstash >/dev/null; then
```
tehasdf: %_user (multiple times in this file)

packaging/cloudify-logstash.spec
```diff
@@ -0,0 +1,99 @@
...
+# Copy static files into place. In order to have files in /packaging/files
+# actually included in the RPM, they must have an entry in the %files
+# section of this spec file.
+cp -R ${RPM_SOURCE_DIR}/packaging/logstash/files/* %{buildroot}
+
+
+%pre
+
+# create logstash group
+if ! getent group logstash >/dev/null; then
```
tehasdf: eh if we have %_user maybe w can have %_group as well

packaging/logstash/files/usr/lib/systemd/system/logstash.service.d/restart.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+Restart=always
```
tehasdf: not adding restartsec or similar?

packaging/cloudify-logstash.spec
funkyHat: No. Some of the repackaged files from the original RPM have hashbangs which point to `/home/jenkins/...`

packaging/cloudify-logstash.spec
funkyHat: These sections are copied from the logstash RPM, because we're bundling the whole application. Hopefully we can drop logstash rather than needing to upgrade it, but in case things need to be merged again I'd rather make minimal changes in copied sections

packaging/logstash/files/usr/lib/systemd/system/logstash.service.d/restart.conf
funkyHat: ahh... this one slipped through the net. I'm going to now say we shouldn't do this change in the same PR as the RPM implementation, because that sounds like a good reason 👍

PR #1152, opened 2018-01-16
CFY-7795 new manager ip setter RPM

funkyHat:


packaging/cloudify-manager-ip-setter.spec
```diff
@@ -0,0 +1,36 @@
...
+# actually included in the RPM, they must have an entry in the %files
+# section of this spec file.
+cp -R ${RPM_SOURCE_DIR}/packaging/ip_setter/files/* %{buildroot}
+
+
+%files
+
+/etc/sudoers.d/cloudify-manager-ip-setter
+
+/opt/cloudify/manager-ip-setter/manager-ip-setter.sh
```
tehasdf: this doesn't need a +x %attr, right?

funkyHat: buildrpm seems to be happily picking up the mode straight from the files

PR #1176, opened 2018-02-02
Update RPM build release info to "ga"

funkyHat:


funkyHat:
Does this need to be based on a 4.3-build branch instead of master?

funkyHat:
already changed

PR #1186, opened 2018-02-08
CFY-7872 handle plugins with plugin.yaml file during snapshot restore

yarivgiga:


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -473,6 +475,34 @@ def should_install(plugin):
...
+        wagon_name = os.path.basename(plugin['path'])
+        ctx.logger.info(
+            'Installing plugin {plugin} for {tenant}'.format(
+                plugin=wagon_name,
+                tenant=tenant,
+            )
+        )
+
+        temp_plugin = os.path.join(plugins_tmp, wagon_name)
+        if len(os.listdir(plugin_dir)) > 1:
```
dusking: comment: means that we have wagon + yaml

PR #1188, opened 2018-02-09
CFY-7792 Make logstash init wait until it starts

geokala:


geokala:
Can be tested with:
sudo sed -i "67i\ \ done" /etc/init.d/logstash
sudo sed -i "67i\ \ \ \ sleep 2" /etc/init.d/logstash
sudo sed -i '67i\ \ \ \ echo "${name} starting..."' /etc/init.d/logstash
sudo sed -i "67i\ \ while ! ps aux | grep 'Xbootclasspath/a:/opt/logstash/vendor/jruby/lib/jruby.jar' | grep 'f /etc/logstash/conf.d' >/dev/null; do" /etc/init.d/logstash
sudo sed -i "67i\ " /etc/init.d/logstash

PR #1227, opened 2018-03-13
Snapshot tests: allow for ConnectionError

tehasdf:
While restoring the database, 500 errors other than CloudifyClientError
can be thrown as well.


tehasdf:
This will be able to be reverted when CY-192 is done

PR #1234, opened 2018-03-21
CY-197 Fix get_secret intrinsic function

inbala:
* Add the value decryption to get_secret intrinsic function

* Add the value encryption to secrets update

* Prevent the update of the DB with the decrypted value in get secret.
  When updating a model in a session of sqlalchemy in the next commit the changes will be updated in the DB

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3/secrets.py
+++ b/rest-service/manager_rest/rest/resources_v3/secrets.py
@@ -38,9 +38,11 @@ class SecretsKey(SecuredResource):
         rest_utils.validate_inputs({'key': key})
         secret = get_storage_manager().get(models.Secret, key)
         encryption_key = config.instance.security_encryption_key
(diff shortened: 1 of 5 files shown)


rest-service/manager_rest/rest/resources_v3/secrets.py
```diff
@@ -91,12 +93,12 @@ def patch(self, key):
...
         """
         request_dict = rest_utils.get_json_and_verify_params({'value'})
-        value = request_dict['value']
+        encrypted_value = self._encrypt_secret_value(request_dict['value'])
         rest_utils.validate_inputs({'key': key})
 
         secret = get_storage_manager().get(models.Secret, key)
         get_resource_manager().validate_modification_permitted(secret)
-        secret.value = value
+        secret.value = encrypted_value
```
mcouthon: Seems like we could just do it one line
```
secret.value = self._encrypt_secret_value(request_dict['value'])
```

PR #1236, opened 2018-03-26
CY-202 Encrypt the secrets values in snapshot restore

inbala:


tests/integration_tests/tests/agentless_tests/test_snapshot.py
```diff
@@ -253,6 +253,28 @@ def test_v_4_2_restore_snapshot_with_resource_availability(self):
...
+        secret_string = self.client.secrets.get('sec1')
+        secret_file = self.client.secrets.get('sec3')
+        assert secret_string.value == 'top_secret'
+        assert 'test_mail' in secret_file.value
+
+        # Validate update secret, the return value is encrypted
+        updated_secret = self.client.secrets.update('sec2', 'top_top_secret')
+        assert updated_secret.value != 'top_top_secret'
+        updated_secret = self.client.secrets.get('sec2')
+        assert updated_secret.value == 'top_top_secret'
```
geokala: This doesn't seem to actually check that the data is encrypted anywhere- e.g. by checking that the particular database field is not set to the secret_string.value?

workflows/cloudify_system_workflows/snapshots/snapshot_create.py
```diff
@@ -173,24 +172,3 @@ def _get_and_create_snapshots_dir(self):
         if not os.path.exists(snapshots_dir):
             os.makedirs(snapshots_dir)
         return snapshots_dir
-
-    def _save_hash_salt(self):
```
geokala: I think removing this will mean that users who are restored from managers with different hash_salt values in rest_security.conf will no longer be usable.

inbala: I'm not sure I understand, in the new snapshot I don't save the hash_salt separately but I save the whole rest-security.conf and then the previous hash_salt will be restored.

tests/integration_tests/tests/agentless_tests/test_snapshot.py
inbala: Is there a better way to validate the value is encrypted than to check that it's value is different?
I can add this 
``` secret_string = get_storage_manager().get(models.Secret, 'sec1')``` 
``` assert secret_string.value != 'top_secret'```
But still it's by checking that the particular database field is not set to the secret_string.value as you said.

geokala: As you're retrieving the string again afterwards, you're confirming that it's encrypted and decrypted successfully, which is reasonable.

You could also generate the encrypted string in the test and check that's in the DB, but I think that's going to add more complexity and fragility to the test than it is worth.

However, checking that what's stored in the database is not exactly the secret that was generated does feel necessary- otherwise, the old behaviour (just storing it plaintext) would also pass this test.

workflows/cloudify_system_workflows/snapshots/snapshot_create.py
geokala: Oh, whoops- yes, good point, for some reason my brain interpreted this method as saving the actual users' password hashes. It doesn't. Sorry, this is fine!

PR #1259, opened 2018-04-10
Remove riemann config directories when deployment is deleted

cswindle:
This fixes an issue where riemann config is not removed and this can lead to a new deployment having VMs healing when first deployed.

### DIFF
--- a/plugins/riemann-controller/riemann_controller/tasks.py
+++ b/plugins/riemann-controller/riemann_controller/tasks.py
@@ -18,6 +18,7 @@ import os
 import time
 import errno
 import json
+import shutil
 import subprocess
 from os import path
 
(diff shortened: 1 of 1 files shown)


geokala:
This looks good. I'll give it a test (hopefully today or tomorrow) before approving though.

cswindle:
Ok, thanks. Regarding testing performed I am currently using a script to continuously redeploy 8 deployments at a time and have done that for the last 4 hours without issue.

Note that I believe that the Cloudify Jira task that this fixes is CY-267 (at least that is what I was told, although I can't see it in Jira myself).

geokala:
You are indeed well informed. <smiles> I have that Jira assigned to me.

geokala:
This appears to behave exactly as intended. I'll approve and merge it.

@cswindle I'm not sure what the method of recognising contributors is at the moment, but please let me know if you want some recognition (other than in the git history) for this and I'll find out what approach we're taking.

PR #1262, opened 2018-04-11
CY-266 update after unify external auth

dusking:


rest-service/manager_rest/security/authentication.py
```diff
@@ -39,10 +39,15 @@ def logger(self):
...
+    @property
+    def external_auth_configured(self):
+        return current_app.external_auth \
+               and current_app.external_auth.configured()
+
     def authenticate(self, request):
         user = self._internal_auth(request)
-        is_bootstrap_admin = user and user.is_bootstrap_admin
-        if self.external_auth \
+        is_bootstrap_admin = bool(user and user.is_bootstrap_admin)
```
tehasdf: the `bool()` is odd, any reason for it? but if it's here, it should probably also be in `_authenticate_password` the same (or factor out a method)

dusking: historical reasons :)

PR #1266, opened 2018-04-12
test_resources_available: check fileserver is not exposed

tehasdf:
The test used to pass the wrong URL, so it appeared as if the resources
were not available, while in reality they are.
Instead, we should test that the fileserver is not exposed at all.


geokala:

PR #1271, opened 2018-04-16
Make riemann stop correctly

geokala:


geokala:
This makes riemann stop correctly (rather than entering a failed state due to exit code 143).

This happens because riemann gets stopped by a sigterm, which apparently causes java processes to exit with status code 143: https://serverfault.com/questions/695849/services-remain-in-failed-state-after-stopped-with-systemctl

Therefore, adding this causes it to enter an inactive state, rather than failed, when the service is stopped.

PR #1283, opened 2018-04-29
CY-247 Add hidden value and visibility to secrets update

inbala:


rest-service/manager_rest/rest/resources_v3/secrets.py
```diff
@@ -118,7 +124,7 @@ def delete(self, key):
         rest_utils.validate_inputs({'key': key})
         storage_manager = get_storage_manager()
         secret = storage_manager.get(models.Secret, key)
-        get_resource_manager().validate_modification_permitted(secret)
+        self._validate_secret_modification_permitted(secret, key)
```
idob-cloudify: key is redundant and can be taken from secret

PR #1293, opened 2018-05-06
CY-318 Encrypt the passwords of RabbitMQ vhosts in the DB

inbala:


rest-service/manager_rest/amqp_manager.py
```diff
@@ -121,13 +122,14 @@ def create_tenant_vhost_and_user(self, tenant):
             self.VHOST_NAME_PATTERN.format(tenant.name)
         username = tenant.rabbitmq_username or \
             self.USERNAME_PATTERN.format(tenant.name)
+
+        # The password is being stored encrypted in the DB
         password = (
             tenant.rabbitmq_password or
-            AMQPManager._generate_user_password()
+            encrypt(AMQPManager._generate_user_password())
```
mcouthon: Eh, I don't like this too much - encrypting just to decrypt a second later.
I understand that the encrypted version is necessary later on, but this doesn't look too elegant

rest-service/manager_rest/constants.py
```diff
@@ -49,3 +49,5 @@
 FILE_SERVER_PLUGINS_FOLDER = 'plugins'
 
 BROKER_SSL_PORT = 5671
+
+SECURITY_FILE_LOCATION = '/opt/manager/rest-security.conf'
```
mcouthon: We already have this string in integration tests. Make sure to remove it there, and use the same constant.

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -375,25 +375,27 @@ def get_deployment_creator_ids_and_tokens(self):
...
-            encrypted_secrets.append((secret[0], encrypted_value))
-
-        update_query = """UPDATE secrets
-                          SET value = encrypted_secrets.value
-                          FROM (VALUES %s) AS encrypted_secrets (id, value)
-                          WHERE secrets.id = encrypted_secrets.id"""
-
-        self.run_query(update_query, vars=encrypted_secrets, bulk_query=True)
+        for value in values['all']:
+            encrypted_value = fernet.encrypt(bytes(value[1]))
```
mcouthon: We are using fernet directly here. Is there a lot of code duplication?

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -405,11 +406,25 @@ def _encrypt_secrets(self, postgres):
...
+    def _encrypt_rabbitmq_passwords(self, postgres):
+        # The passwords are encrypted
+        if self._snapshot_version >= V_4_4_0:
+            return
+
+        with open(SECURITY_FILE_LOCATION) as security_conf_file:
+            rest_security_conf = json.load(security_conf_file)
+
+        ctx.logger.info('Encrypting the passwords of RabbitMQ vhosts')
+        postgres.encrypt_values(str(rest_security_conf['encryption_key']),
```
mcouthon: You're using it twice (line 409) - might as well make it a variable

PR #1308, opened 2018-05-13
CY-326 A secret can be updated to hidden-value only by the admins or …

inbala:
…the secret creator

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3/secrets.py
+++ b/rest-service/manager_rest/rest/resources_v3/secrets.py
@@ -44,7 +44,7 @@ class SecretsKey(SecuredResource):
         secret = get_storage_manager().get(models.Secret, key)
         secret_dict = secret.to_dict()
         if secret_dict['is_hidden_value'] and not \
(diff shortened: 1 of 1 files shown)


rest-service/manager_rest/rest/resources_v3/secrets.py
```diff
@@ -132,28 +132,36 @@ def _encrypt_secret_value(self, value, encryption_key=None):
...
 
     def _update_is_hidden_value(self, secret):
         is_hidden_value = request.json.get('is_hidden_value')
-        if is_hidden_value is not None:
-            is_hidden_value = rest_utils.verify_and_convert_bool(
-                'is_hidden_value',
-                is_hidden_value
+        if is_hidden_value is None:
+            return
+        is_hidden_value = rest_utils.verify_and_convert_bool(
```
idob-cloudify: can do this after _is_hidden_value_permitted

inbala: I think it is better before because it also validate that is_hidden_value is a valid boolean

PR #1341, opened 2018-06-18
CY-405 Replace Logstash with a new, dedicated service

mcouthon:


amqp-postgres/amqp_postgres/tests/tests.py
```diff
@@ -0,0 +1,221 @@
...
+        return {
+            'context': {
+                'blueprint_id': 'bp',
+                'deployment_id': 'dep',
+                'execution_id': execution_id,
+                'node_id': 'vm_7j36my',
+                'node_name': 'vm',
+                'operation': 'cloudify.interfaces.cloudify_agent.create',
+                'plugin': 'agent',
+                'task_id': 'a13973d5-3866-4054-baa1-479e242fff75',
```
tehasdf: this is not asserted on, right? maybe it shouldnt be here at all, or do you prefer to keep for reference?
(ditto "logger")

packaging/amqp-postgres/files/usr/lib/systemd/system/cloudify-amqp-postgres.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Cloudify AMQP PostgreSQL Broker Service
+Wants=postgresql-9.5.service cloudify-rabbitmq.service
+After=postgresql-9.5.service cloudify-rabbitmq.service
+
+[Service]
+TimeoutStartSec=0
+Restart=on-failure
+EnvironmentFile=/etc/sysconfig/cloudify-amqp-postgres
```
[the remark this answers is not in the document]
tehasdf: oh and this must of course have the restservice conf files in the envs, right

amqp-postgres/amqp_postgres/__init__.py
```diff
@@ -0,0 +1,15 @@
+########
+# Copyright (c) 2018 GigaSpaces Technologies Ltd. All rights reserved
```
tehasdf: `Cloudify Platform Ltd.` i guess (same for all license headers)

.circleci/config.yml
```diff
@@ -78,6 +78,27 @@ jobs:
...
+          TEST_DATABASE_URL: postgresql://cloudify:cloudify@localhost/cloudify_db
+      - image: rabbitmq:3.7.4
+      - image: circleci/postgres:9.5-alpine
+        environment:
+          POSTGRES_USER: cloudify
+          POSTGRES_PASSWORD: cloudify
+          POSTGRES_DB: cloudify_db
+    steps:
+      - checkout
+      - run: sudo apt-get install libldap-dev libsasl2-dev
```
tehasdf: we need ldap here?

amqp-postgres/amqp_postgres/amqp_consumer.py
```diff
@@ -0,0 +1,62 @@
...
+
+        # This is here because AMQPConnection expects it
+        self.routing_key = ''
+
+    def register(self, connection):
+        self._connection = connection
+        self._in_channel = connection.channel()
+        self._register_queue(self._in_channel)
+
+    def _register_queue(self, channel):
```
tehasdf: might as well combine the two methods, this and `register`, the split i think does nothing

amqp-postgres/amqp_postgres/amqp_consumer.py
```diff
@@ -0,0 +1,62 @@
...
+        self.queue = str(uuid4())
+        self._connection = None
+        self._in_channel = None
+        self._message_processor = message_processor
+
+        # This is here because AMQPConnection expects it
+        self.routing_key = ''
+
+    def register(self, connection):
+        self._connection = connection
```
tehasdf: no need to even store `_connection` and the channel, is there?

amqp-postgres/amqp_postgres/amqp_consumer.py
```diff
@@ -0,0 +1,62 @@
...
+
+from flask import current_app
+
+
+class AMQPLogsEventsConsumer(object):
+    LOGS_EXCHANGE = 'cloudify-logs'
+    EVENTS_EXCHANGE = 'cloudify-events'
+
+    def __init__(self, message_processor):
+        self.queue = str(uuid4())
```
tehasdf: the queue name should likely be non-random
the idea is that if this program crashes, messages will still be routed to the queue, and after the program starts again, it will be able to consume them
for this, it must use the same queue name, and the queue must be durable, no auto delete, not exclusive (so the queue is set correctly, just the name is random)

packaging/amqp-postgres/files/usr/lib/systemd/system/cloudify-amqp-postgres.service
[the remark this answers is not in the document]
tehasdf: oh yeah, duh, this one goes in the -install PR, sorry

packaging/amqp-postgres/files/usr/lib/systemd/system/cloudify-amqp-postgres.service
```diff
@@ -0,0 +1,15 @@
...
+User=cfyuser
+Group=cfyuser
+ExecStart=/opt/manager/env/bin/python /opt/manager/env/bin/cloudify-amqp-postgres
```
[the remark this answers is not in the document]
mcouthon: Hmm, because it's in the same venv? I'll test this

.circleci/config.yml
mcouthon: Yeah =\ Unfortunately without it you can't install the rest-service venv

tehasdf:
obsoleted by #1392

PR #1342, opened 2018-06-18
CY-413 Validate reinstall list in deployment update

yarivgiga:


rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -339,17 +339,52 @@ def _patch_changes_with_relationship_index(raw_node_instances, raw_nodes):
...
+                                 remove,
+                                 dep_update):
+        """validate node-instances explicitly supplied to reinstall list exist
+        and are not about to be installed or uninstalled in this update"""
+        node_instances = self.sm.list(
+            models.NodeInstance,
+            filters={'deployment_id': dep_update.deployment_id}
+        )
+        node_instances_ids = [n.id for n in node_instances]
+        add_conflict = [n for n in reinstall if n in add]
```
tehasdf: we could use set operations here like
```python
node_instances_ids = {n.id for n in node_instances}
add_conflict = set(reinstall) | set(add)
remove_conflict = set(reinstall) | set(remove)
not_existing = set(reinstall) - node_instances_ids
```

tehasdf: but if you prefer lists, not a big difference either way

yarivgiga: actually i usually do prefer list, though with sets it does seems more elegant.
but it should be &, not |

tehasdf: yes ofc, sorry ;P

PR #1368, opened 2018-07-06
Delete dep-env: don't uninstall managed plugins

tehasdf:


tehasdf:
requires cloudify-cosmo/cloudify-agent#429

PR #1394, opened 2018-07-30
CY-533: gunicorn protection

isaac-s:
- Use `systemd`'s `RuntimeDirectory` instead of `tmpfiles.d`
- Use a unix socket instead of TCP/IP


geokala:

PR #1402, opened 2018-08-02
CY-405 amqppostgres: don't break on missing executions 

tehasdf:


amqp-postgres/amqp_postgres/postgres_publisher.py
```diff
@@ -240,6 +248,9 @@ def _get_log(message, execution):
 
     @staticmethod
     def _get_event(message, execution):
+        task_error_causes = message['context'].get('task_error_causes')
+        if task_error_causes is not None:
+            task_error_causes = json.dumps(task_error_causes)
```
tehasdf: equivalent to:
https://github.com/cloudify-cosmo/cloudify-manager-install/blob/4.4-build/cfy_manager/components/logstash/config/logstash.conf#L70-L72

PR #1412, opened 2018-08-14
Add a REST endpoint for `cfy agents list`

tehasdf:


rest-service/manager_rest/resource_manager.py
```diff
@@ -1057,6 +1061,76 @@ def add_node_instance_from_dict(self, instance_dict):
...
+        if not nodes:
+            pagination = {'total': 0, 'size': 0, 'offset': 0}
+            return ListResult([], metadata={'pagination': pagination})
+
+        instance_filters = {'_node_fk': list(nodes)}
+        if node_instance_ids:
+            instance_filters['id'] = node_instance_ids
+        node_instances = self.sm.list(models.NodeInstance,
+                                      filters=instance_filters,
+                                      get_all_results=True)
```
geokala: What will this do in the case of there being more than max_results node instances?

rest-service/manager_rest/resource_manager.py
```diff
@@ -1057,6 +1061,76 @@ def add_node_instance_from_dict(self, instance_dict):
...
+            return
+        agent = instance.runtime_properties.get('cloudify_agent')
+        if not agent:
+            return
+        if agent.get('windows'):
+            system = 'windows'
+        else:
+            system = agent.get('distro')
+            if agent.get('distro_codename'):
+                system = '{0} {1}'.format(system, agent.get('distro_codename'))
```
geokala: Can (or should) this also run some sort of check of agent health?

tehasdf: no, we have `cfy agents validate` for that

ideally i'd like to have it here as well, but it's unwieldy, since it would require querying the agents, which should be async and not in the restservice... so can't really do that right now

PR #1413, opened 2018-08-15
Cy 538 queued executions

adigrabow:


rest-service/manager_rest/resource_manager.py
```diff
@@ -1236,7 +1408,7 @@ def _create_deployment_environment(self,
         wf_id = 'create_deployment_environment'
         deployment_env_creation_task_name = \
             'cloudify_system_workflows.deployment_environment.create'
-
+        # time.sleep(30)
```
tehasdf: remember to remove commented code before merge

rest-service/manager_rest/resource_manager.py
```diff
@@ -593,6 +761,10 @@ def cancel_execution(self, execution_id, force=False, kill=False):
         if kill:
             force = True
         execution = self.sm.get(models.Execution, execution_id)
+        # When a user cancels queued execution automatically use the kill flag
+        if execution.status == ExecutionState.QUEUED:
+            kill = True
```
tehasdf: hmm, tbh shouldn't be a reason to kill, does it actually change anything for queued executions?

rest-service/manager_rest/resource_manager.py
```diff
@@ -488,31 +631,48 @@ def _check_for_any_active_executions(self):
...
                 .format(executions))
+        return should_queue
 
-    def _check_for_active_system_wide_execution(self):
+    def _check_for_active_system_wide_execution(self, queue):
         filters = {
             'status': ExecutionState.ACTIVE_STATES
         }
+        should_queue = False
         for e in self.list_executions(is_include_system_workflows=True,
```
tehasdf: btw this should likely have `get_all_results=True`, otherwise we'll miss executions when there's more than 1000 of them?

rest-service/manager_rest/resource_manager.py
```diff
@@ -488,31 +631,48 @@ def _check_for_any_active_executions(self):
             for e in self.list_executions(is_include_system_workflows=True,
```
tehasdf: same as below

rest-service/manager_rest/resource_manager.py
```diff
@@ -479,7 +507,122 @@ def execute_workflow(self,
...
+    @staticmethod
+    def _verify_workflow_in_deployment(wf_id, deployment, dep_id):
+        if wf_id not in deployment.workflows:
+            raise manager_exceptions.NonexistentWorkflowError(
+                'Workflow {0} does not exist in deployment {1}'.format(
+                    wf_id, dep_id))
+
+    def _check_for_executions(self, deployment_id, force, queue):
+        """
+        This function returns True if one of the following is running:
```
tehasdf: it also depends on the `queue` param so maybe the docstring should reflect that

rest-service/manager_rest/resource_manager.py
```diff
@@ -479,7 +507,122 @@ def execute_workflow(self,
...
+        deployment = execution.deployment
+        deployment_id = None
+        if deployment:
+            deployment_id = deployment.id
+        workflow_id = execution.workflow_id
+        execution_parameters = execution.parameters
+
+        if self._should_use_system_workflow_executor(execution):
+            # Use `execute_system_workflow`
+            verify_no_executions = True
```
tehasdf: this flag is never changed?

rest-service/manager_rest/resource_manager.py
```diff
@@ -78,9 +79,32 @@ def update_execution_status(self, execution_id, status, error):
...
+
+        for e in queued_executions:
+            self.execute_queued_workflow(e)
+            if e.is_system_workflow:  # To avoid starvation of system workflows
+                break
+
+    def _get_queued_executions(self):
+        filters = {'status': ExecutionState.QUEUED_STATE}
+        queued_executions = [
+            e for e in self.list_executions(is_include_system_workflows=True,
```
tehasdf: do we really need a list here? maybe we can just `return self.list_executions().items`..

rest-service/manager_rest/resource_manager.py
```diff
@@ -78,9 +79,32 @@ def update_execution_status(self, execution_id, status, error):
...
+                                            filters=filters,
+                                            sort={'created_at': 'asc'}).items
```
tehasdf: can we also sort that system execs come first?

rest-service/manager_rest/resource_manager.py
```diff
@@ -479,7 +507,122 @@ def execute_workflow(self,
...
+
+            return execution
+
+        else:  # User `execute_workflow`
+
+            blueprint = self.sm.get(models.Blueprint, deployment.blueprint_id)
+            self._verify_workflow_in_deployment(workflow_id, deployment,
+                                                deployment_id)
+            workflow = deployment.workflows[workflow_id]
+            self._verify_deployment_environment_created_successfully(
```
tehasdf: hmm this will throw a RuntimeError if not, aborting the whole `start_queued_executions` loop, perhaps not what we want...

tehasdf: and it would mean that the previous execution status update, which called the `start_queued_executions` would return a 500 as well, certainly not what we want ;P

rest-service/manager_rest/resource_manager.py
adigrabow: It won't be executed if it's status isn't `queued`. So if a user wants to cancel a queued execution he will use `cfy exec cancel` with no flags (and we will add them in the rest).

rest-service/manager_rest/resource_manager.py
```diff
@@ -479,7 +507,122 @@ def execute_workflow(self,
...
+        queue = True
+        force = False
+        deployment = execution.deployment
+        deployment_id = None
+        if deployment:
+            deployment_id = deployment.id
+        workflow_id = execution.workflow_id
+        execution_parameters = execution.parameters
+
+        if self._should_use_system_workflow_executor(execution):
```
idob-cloudify: check if you can reuse execute_system_workflow

rest-service/manager_rest/resource_manager.py
```diff
@@ -479,7 +507,122 @@ def execute_workflow(self,
...
+            workflow_executor.execute_system_workflow(
+                wf_id=workflow_id,
+                task_id=execution.id,
+                task_mapping=self.task_mapping.get(workflow_id),
+                deployment=deployment,
+                execution_parameters=execution_parameters)
+
+            return execution
+
+        else:  # User `execute_workflow`
```
idob-cloudify: check if you can reuse execute_workflow

rest-service/manager_rest/resource_manager.py
tehasdf: @idob-cloudify in the loop calling this, we break on the first system workflow found "to avoid starving system executions"; but it's still possible that there's many non-system execs and the system exec will have to wait

if we sorted this so that system execs come first, then the queueing would prioritize system execs, maybe that's what we want?

idob-cloudify: no, we want to execute in the order of creation.

PR #1426, opened 2018-08-27
CY 586 Add documentation for log saving

davidginzbourg:
- Also adds the option not to remove the `tar.gz` log files for the logs (more organized).

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -97,13 +97,12 @@ This project runs tests on a Cloudify Manager container created by [`docl`](http
 All done!
 
 Here'es all the commands together:
-     ```
+
      sudo brctl addbr cfy0
      sudo ip addr add 172.20.0.1/24 dev cfy0
      sudo ip link set dev cfy0 up
      sudo dockerd --bridge cfy0 --host 172.20.0.1
      export DOCKER_HOST=172.20.0.1
-     ```
 
 ## Running Tests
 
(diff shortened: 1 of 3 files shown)


tests/integration_tests/tests/test_cases.py
```diff
@@ -102,6 +102,7 @@ def tearDown(self):
     def _save_manager_logs_after_test(self, purge=True):
         self.logger.debug('_save_manager_logs_after_test started')
         logs_dir = os.environ.get('CFY_LOGS_PATH')
+        self.logger.info("CFY_LOGS_PATH: {}".format(logs_dir))
```
tyacbovi: Add more description, mention that this is the remote and you can add the local (if it's set).

tests/integration_tests/tests/test_cases.py
```diff
@@ -124,8 +125,10 @@ def _save_manager_logs_after_test(self, purge=True):
         self.logger.debug('opening tar.gz: {0}'.format(target))
         with tarfile.open(target) as tar:
             tar.extractall(path=logs_dir)
-        self.logger.debug('removing {0}'.format(target))
-        os.remove(target)
+        if not bool(os.environ.get('DONT_REMOVE_LOG_TAR')):
```
tyacbovi: it's better not to extract to tar for the first place, because it's a waste of time in that case

tests/README.md
```diff
@@ -114,6 +113,17 @@ $ cd <root directory of cloudify-manager repository>
 $ pytest -s tests/integration_tests/tests/agentless_tests/test_workflow.py:BasicWorkflowsTest.test_execute_operation
 ```
 
+### Saving the Cloudify Manager's logs
+In order to save the logs for tests, specify the path via an environment variable as follows:
+
+`export CFY_LOGS_PATH=<YOUR-PATH-HERE>`
```
tyacbovi: Maybe change CFY_LOGS_PATH => to CFY_LOGS_PATH_REMOTE

And add example of how the path should look like when in remote mode

tests/integration_tests/tests/test_cases.py
davidginzbourg: :+1:

tests/integration_tests/tests/test_cases.py
davidginzbourg: Change to `SKIP_LOG_SAVE_ON_SUCCESS`

tests/README.md
davidginzbourg: :+1:

PR #1429, opened 2018-08-28
CY-586 Add log extraction

davidginzbourg:
- Setting `CFY_LOGS_PATH` path will now save the logs on the remote
machine in that path.
- Setting `SKIP_LOGS_EXTRACTION` will not remove the `logs.tar.gz` files and skip extraction.


tests/README.md
```diff
@@ -114,6 +113,17 @@ $ cd <root directory of cloudify-manager repository>
 $ pytest -s tests/integration_tests/tests/agentless_tests/test_workflow.py:BasicWorkflowsTest.test_execute_operation
 ```
 
+### Saving the Cloudify Manager's logs
+In order to save the logs for tests before the test environment is destroyed, specify the path via an environment variable as follows:
+
+`export CFY_LOGS_PATH_REMOTE=<YOUR-PATH-HERE>`
```
tyacbovi: Add example for this

davidginzbourg: Done

PR #1438, opened 2018-09-03
CY-596 Send an event when a workflow is queued

inbala:


rest-service/manager_rest/resource_manager.py
```diff
@@ -1696,6 +1698,34 @@ def _any_running_executions(executions):
...
+            'context': message_context,
+            'message': {
+                'text': "'{0}' workflow execution was queued".format(
+                    execution.workflow_id
+                ),
+                'arguments': None
+            }
+        }
+
+        cloudify_utils.send_event(event, 'hook')
```
tehasdf: remember to fix

PR #1448, opened 2018-09-07
amqp_postgres: pass errors to the main thread

tehasdf:
Currently an error thrown in the db thread won't cause the whole
application to exit with code 1 (and be restarted by systemd).

Let's pass the error to the main thread and raise it there.


amqp-postgres/amqp_postgres/postgres_publisher.py
```diff
@@ -273,7 +276,8 @@ def _insert_logs(self, cursor, logs):
     def on_db_connection_error(self):
         logger.critical('Database down - cannot continue')
         self._amqp_connection.close()
-        raise RuntimeError('Database down')
+        self.error_exit = RuntimeError('Database down')
```
geokala: Do we really want to mask the original error? Can we pass through the specific DB error?

PR #1455, opened 2018-09-12
CY-624 prefer to use _storage_id over id while encrypting data in sna…

yarivgiga:
…pshot restore, in order to avoid mixture between resources with the same id in different tenants

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/postgres.py
+++ b/workflows/cloudify_system_workflows/snapshots/postgres.py
@@ -378,8 +378,15 @@ class Postgres(object):
     def encrypt_values(self, encryption_key, table_name, column_name):
         """Encrypt the values of one column in a table
         """
-        values = self.run_query("SELECT id, {0} FROM {1}".format(column_name,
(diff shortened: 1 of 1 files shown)


workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -378,8 +378,15 @@ def get_deployment_creator_ids_and_tokens(self):
     def encrypt_values(self, encryption_key, table_name, column_name):
         """Encrypt the values of one column in a table
         """
-        values = self.run_query("SELECT id, {0} FROM {1}".format(column_name,
-                                                                 table_name))
+        # try to work with _storage_id, if does not exists, try with id
+        key = '_storage_id'
+        values = self.run_query("SELECT {0}, {1} FROM {2}".format(
+            key, column_name, table_name))
+        if values['all'] is None:
```
tehasdf: this is actually what is returned? i thought querying for a non-existent column would be an IntegrityError...?

PR #1458, opened 2018-09-16
Fixed 'cfy status' message with cluster

barucoh:


rest-service/manager_rest/rest/resources_v1/status.py
```diff
@@ -69,7 +69,8 @@ def get(self, **kwargs):
                 job for job in jobs
                 if self._should_be_in_services_output(job)
             ]
-            if config.instance.postgresql_host != 'localhost':
+            if config.instance.postgresql_host != 'localhost' \
+                    and config.instance.postgresql_host != '127.0.0.1:15432':
```
tyacbovi: You can lose the port and check substring, because the user can change the port

barucoh: Actually he can't, because it's configured by our cluster mechanism.
But I will checkout the substring, it makes more sense.

PR #1459, opened 2018-09-16
Reversed logical expression to properly fix 'cfy status'

barucoh:


rest-service/manager_rest/rest/resources_v1/status.py
```diff
@@ -70,8 +70,8 @@ def get(self, **kwargs):
                 if self._should_be_in_services_output(job)
             ]
             # If PostgreSQL is not local, print it as 'remote'
-            if config.instance.postgresql_host.startswith('localhost') \
-                    or config.instance.postgresql_host.startswith('127.0.0.1'):
+            if not (config.instance.postgresql_host.startswith('localhost') or
```
tehasdf: `.startswith` is funny in that it can also take a tuple: 
```python
if not config.instance.postgresql_host.startswith(('localhost', '127.0.0.1'))
```

PR #1468, opened 2018-10-02
CYBL-595: consider quotes optional, as this file may be overwritten b…

isaac-s:
…y 'cfy ldap set' and lose its quotes


geokala:

PR #1475, opened 2018-10-04
CY-666 fix config reset with write - write into relevant files

yarivgiga:


rest-service/manager_rest/config.py
```diff
@@ -94,9 +98,20 @@ def load_from_file(self, filename, namespace=''):
...
     global instance
     instance = configuration
-    if write:
-        with open(os.environ['MANAGER_REST_CONFIG_PATH'], 'w') as f:
-            if instance:
+    if not write:
+        return
+
+    configs = {}
+    for key, value in instance.__dict__:
```
tehasdf: maybe better to use `vars(instance)` instead, that's more public api (also works if the instance has __slots__)

PR #1479, opened 2018-10-07
CY-661 Add dry_run indication to Executions

mcouthon:


resources/rest-service/cloudify/migrations/versions/1fbd6bf39e84_4_5_to_4_5_5.py
```diff
@@ -0,0 +1,28 @@
...
+depends_on = None
+
+
+def upgrade():
+    # server_default accepts string or SQL element only
+    op.add_column('executions', sa.Column('is_dry_run',
+                                          sa.Boolean(),
+                                          nullable=False,
+                                          server_default='f'))
+
```
adigrabow: .

PR #1484, opened 2018-10-08
CYBL-606: add dateext format to all logrotate policies so they can be…

isaac-s:
… rotated more than once a day if needed

### DIFF
--- a/tests/integration_tests/tests/agentless_tests/test_manager_misc.py
+++ b/tests/integration_tests/tests/agentless_tests/test_manager_misc.py
@@ -95,8 +95,9 @@ class MiscManagerTest(AgentlessTestCase):
                               'nginx',
                               'restservice',
                               'stage']:
-            self.logger.info('Cancelling for %s' % logrotate_cfg)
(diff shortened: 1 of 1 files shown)


tests/integration_tests/tests/agentless_tests/test_manager_misc.py
```diff
@@ -87,8 +87,17 @@ def test_logrotation(self):
...
-        self.execute_on_manager(sed_cmd)
+        # We need to do this separately for each logrotate configuration.
+        for logrotate_cfg in ['cloudify-amqp-postgres',
+                              'cloudify-mgmtworker',
+                              'cloudify-rabbitmq',
+                              'composer',
+                              'nginx',
+                              'restservice',
+                              'stage']:
+            self.logger.info('Cancelling for %s' % logrotate_cfg)
```
tehasdf: in general we use `.format` instead of %-formatting
however python's logging can do this: `logger.info('Cancelling for %s', logrotate_cfg)` (so that the potentially-expensive string formatting isn't done unless the log is actually emitted, wrt. logging level etc)
doing %-formatting here is really the worst choice :P

PR #1486, opened 2018-10-08
CY-680: blueprint catalog import

tyacbovi:
Adding resolver to catalog uploaded blueprints.

### DIFF
--- a/rest-service/manager_rest/test/infrastructure/test_parse_with_resolver.py
+++ b/rest-service/manager_rest/test/infrastructure/test_parse_with_resolver.py
@@ -1,0 +1,66 @@
+#########
+# Copyright (c) 2018 Cloudify Platform Ltd. All rights reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+#  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  * See the License for the specific language governing permissions and
+#  * limitations under the License.
+
+import os
+
+from dsl_parser import tasks, constants
+from manager_rest.test.attribute import attr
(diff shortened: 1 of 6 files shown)


rest-service/manager_rest/test/mock_blueprint/blueprint_with_blueprint_import.yaml
```diff
@@ -0,0 +1,9 @@
+tosca_definitions_version: cloudify_dsl_1_3
+
+imports:
+    - blueprint:imported_blueprint
+
+# The blueprint contains two nodes, a vm and a web server hosted in it.
```
tehasdf: move the comment up so it's clear that the comment is about the imported bp

rest-service/manager_rest/test/infrastructure/test_parse_with_resolver.py
```diff
@@ -0,0 +1,70 @@
...
+            self.get_blueprint_path('mock_blueprint'),
+            'blueprint_with_blueprint_import.yaml')
+
+        resolver = ResolverWithCatalogSupport()
+
+        plan = tasks.parse_dsl(dsl_location=dsl_location,
+                               resources_base_path=self.tmpdir,
+                               resolver=resolver)
+
+        nodes = set([node["name"] for node in plan[constants.NODES]])
```
tehasdf: this is 2.7+ only so might as well use set syntax:
```python
nodes = {node["name"] for node in plan[constants.NODES]}
self.assertEqual(nodes, {"test", "vm", "http_web_server"})
```

rest-service/manager_rest/test/infrastructure/test_parse_with_resolver.py
```diff
@@ -0,0 +1,70 @@
...
+            self.get_blueprint_path('mock_blueprint'),
+            'blueprint_with_plugin_import.yaml')
+
+        resolver = ResolverWithCatalogSupport()
+        self.upload_plugin(TEST_PACKAGE_NAME, TEST_PACKAGE_VERSION)
+
+        plan = tasks.parse_dsl(dsl_location=dsl_location,
+                               resources_base_path=self.tmpdir,
+                               resolver=resolver)
+        self.assertIsNone(plan[constants.DEPLOYMENT_PLUGINS_TO_INSTALL])
```
tehasdf: looks like this is actually an empty list, not None

PR #1487, opened 2018-10-08
Fix flaky test (compare lower to lower)

geokala:


tehasdf:
merging as the codecov is not relevant, this doesn't touch amqp-postgres's coverage or code at all

PR #1488, opened 2018-10-08
CY-682 allow setting ldap without supplying admin credentials

yarivgiga:


rest-service/manager_rest/rest/resources_v3/manager.py
```diff
@@ -170,11 +171,26 @@ def _validate_set_ldap_request(self):
...
         })
+        ldap_config['ldap_username'] = ldap_config.get('ldap_username')
+        ldap_config['ldap_password'] = ldap_config.get('ldap_password')
+        ldap_config['ldap_is_active_directory'] = \
+            rest_utils.verify_and_convert_bool(
+                'ldap_is_active_directory',
+                ldap_config.get('skip_uninstall', 'false')
+            )
+
+        if ((ldap_config['ldap_username']
```
geokala: You should probably make these us 'is None', otherwise there will be problems if someone uses something like:
Username = 'anonymous'
Password = ''

(hopefully with a suitably restricted account)

yarivgiga: question is if we are supposed to support empty passwords...

geokala: 1. It's not really about whether we support it- LDAP systems can support it, e.g. for an account that only allows lookup of the entities we need.

2. If we do this then we might cause problems on upgrade for any users who do already have an empty password.

3. If we're not supporting it then we need to return a helpful error message, not just silently fall through to a different behaviour.

geokala: I see now that we won't silently fail through, but if we're going ahead with not supporting blank passwords then I suggest we either put in a separate check for empty-string passwords, or we add to the BadParametersError below a note that blank passwords are not supported.

PR #1490, opened 2018-10-10
CY-644 Capabilities (WIP)

mcouthon:


rest-service/manager_rest/test/endpoints/test_capabilities.py
```diff
@@ -0,0 +1,110 @@
+#########
+# Copyright (c) 2013 GigaSpaces Technologies Ltd. All rights reserved
```
tehasdf: 2018 cloudify platform ltd

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -169,6 +169,7 @@ class Deployment(SQLResourceBase):
     policy_triggers = db.Column(db.PickleType)
     policy_types = db.Column(db.PickleType)
     outputs = db.Column(db.PickleType(comparator=lambda *a: False))
+    capabilities = db.Column(db.PickleType(comparator=lambda *a: False))
```
tehasdf: can this at least be jsonstring

mcouthon: So, originally we didn't use JSONs in postgres because they didn't work in sqlite (unittests), and because we wanted to remain DB neutral (e.g. switching to MySQL or whatever), as this was a Potsgres only feature. If we want to pronounce that we're only using postgres, which is quite reasonable, seeing as how all of our non-python code is hell bent on doing PG specific stuff, then I'm all in favor. But for now, for consistency's sake, I think we should stick with this PickleRick, until we decide to change it everywhere.

rest-service/manager_rest/test/endpoints/test_capabilities.py
```diff
@@ -0,0 +1,110 @@
...
+#  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  * See the License for the specific language governing permissions and
+#  * limitations under the License.
+
+from manager_rest.test import base_test
+from manager_rest.test.attribute import attr
+
+
+@attr(client_min_version=3.1, client_max_version=base_test.LATEST_API_VERSION)
+class CapabilitiesTestCase(base_test.BaseServerTestCase):
```
tyacbovi: Needs test on a non-existing capability

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -85,3 +91,19 @@ def patch(self, deployment_id):
         return get_resource_manager().set_visibility(models.Deployment,
                                                      deployment_id,
                                                      visibility)
+
+
+class DeploymentsIdCapabilities(SecuredResource):
+
+    @swagger.operation(
```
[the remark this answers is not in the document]
mcouthon: Not sure what you mean

rest-service/manager_rest/test/endpoints/test_capabilities.py
mcouthon: Done

PR #1498, opened 2018-10-18
CY-720 Add 'default sort column' to models

tehasdf:
So that queries are ordered by something even if the user didn't
request any sorting (so that pagination makes sense).

### DIFF
--- a/rest-service/manager_rest/storage/models_base.py
+++ b/rest-service/manager_rest/storage/models_base.py
@@ -216,6 +216,15 @@ class SQLModelBase(db.Model):
     def unique_id(cls):
         return 'id'
 
+    @classmethod
+    def default_sort_column(cls):
+        """If no sort is requested, order by the column specified by this.
+
+        This is so that requests with pagination make sense even with no
+        sort requested by the user.
+        """
+        return getattr(cls, cls.unique_id())
+
     def __repr__(self):
         """Return a representation of the class, based on the ordered dict of
         identifiers returned by `_get_identifier_dict`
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/storage/models_base.py
```diff
@@ -216,6 +216,15 @@ def _get_identifier_dict(self):
...
         return 'id'
 
+    @classmethod
+    def default_sort_column(cls):
+        """If no sort is requested, order by the column specified by this.
+
+        This is so that requests with pagination make sense even with no
+        sort requested by the user.
+        """
+        return getattr(cls, cls.unique_id())
```
isaac-s: Does `cls.unique_id()` translate to a DB-proprietary sequence?

tehasdf: it returns whatever the class wants to return, in this case `id` (SQLModelBase) or `_storage_id` (SQLResourceBase) (so for now, yes)

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -83,6 +83,10 @@ def _sort_query(query, sort=None):
                 if order == 'desc':
                     column = column.desc()
                 query = query.order_by(column)
+        else:
+            default_sort = model_class.default_sort_column()
+            if default_sort:
```
mcouthon: How can it be empty?

tehasdf: the implementor of the model class might well do a `return None` if they want no default sort for that model

PR #1499, opened 2018-10-21
CY-693 : adding protection against deleting imported blueprints

tyacbovi:
- The protection will be against deleting blueprints in use in another blueprint (even not in live deployments).
- The force delete option was added in case of need
- Merging two unit tests files that are on the same topic


rest-service/dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==4.5.5.dev1
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-693#egg=cloudify-common==4.5.5.dev1
```
geokala: This shouldn't be in a PR

rest-service/manager_rest/manager_exceptions.py
```diff
@@ -429,6 +429,18 @@ def __init__(self, *args, **kwargs):
         )
 
 
+class BlueprintInUseError(ManagerException):
```
geokala: Don't we already have something like this error when you try to delete a blueprint that an existing deployment uses?

rest-service/manager_rest/rest/resources_v3_1/blueprints.py
```diff
@@ -72,3 +75,23 @@ def put(self, blueprint_id, **kwargs):
         return UploadedBlueprintsManager().\
             receive_uploaded_data(data_id=blueprint_id,
                                   visibility=visibility)
+
+    @swagger.operation(
```
geokala: I didn't realise we were missing this in swagger before. Oh dear. We should probably have some tests checking that all of our API paths are actually in swagger?

geokala: (but not as part of this)

tyacbovi: Agree, also there should be work on updating the swagger docs and fixing our swagger specification generation

rest-service/dev-requirements.txt
tyacbovi: Need here so the unit tests can be run, because this change depends on common change

rest-service/manager_rest/manager_exceptions.py
tyacbovi: Nope, there isn't

rest-service/manager_rest/resource_manager.py
```diff
@@ -379,17 +379,43 @@ def publish_blueprint(self,
...
+                    'blueprint removal.'.format(blueprint_id))
+
         if len(blueprint.deployments) > 0:
             raise manager_exceptions.DependentExistsError(
                 "Can't delete blueprint {0} - There exist "
                 "deployments for this blueprint; Deployments ids: {1}"
                 .format(blueprint_id,
                         ','.join([dep.id for dep
                                   in blueprint.deployments])))
+        # Delete blueprint resources from file server
```
mcouthon: Why did you move it here?

tyacbovi: So I should not need to copy this in the V3.1 and that's the more relevant place (like the plugins)

rest-service/manager_rest/test/endpoints/test_blueprints.py
```diff
@@ -36,18 +37,18 @@ def test_get_empty(self):
     def test_get_nonexistent_blueprint(self):
         try:
             self.client.blueprints.get('15')
-        except CloudifyClientError, e:
+        except exceptions.CloudifyClientError, e:
```
tehasdf: if we're already changing those lines, please also change the `except E, e` to `except E as e`, which because the comma syntax is old and deprecated (and stops working in py3)

rest-service/manager_rest/test/endpoints/test_blueprints.py
```diff
@@ -130,8 +131,9 @@ def test_delete_blueprint(self):
                 post_blueprints_response['id'], 'blueprint.yaml'))
 
         # trying to delete a nonexistent blueprint
-        resp = self.delete('/blueprints/nonexistent-blueprint')
-        self.assertEquals(404, resp.status_code)
+        with self.assertRaises(exceptions.CloudifyClientError) as context:
```
tehasdf: we can change that to a specific exception type (not the catch-all CloudifyClientError), and then we won't need to assert the 404

PR #1501, opened 2018-10-22
Adding new REST request to get Manager's REST config

barucoh:


rest-service/manager_rest/rest/resources_v3_1/manager_config.py
```diff
@@ -49,3 +49,25 @@ def _authorization_config():
...
+        current_app.logger.info('Retrieving Manager REST config. filter: {0}'.
+                                format(request_args))
+        result = dict()
+        filter_by = request_args.get('filter', [])
+        for key in filter_by:
+            result[key] = getattr(config.instance, key)
+        if not filter_by:
+            current_app.logger.debug('Retrieving Manager REST config without a'
+                                     ' filter')
+            result = config.instance.__dict__
```
mcouthon: This is a dangerous practice. Using `__dict__` is usually discouraged, _especially_ if you're returning this object directly to the user.
e.g. https://stackoverflow.com/a/4727598/978089

rest-service/manager_rest/rest/resources_v3_1/manager_config.py
```diff
@@ -49,3 +49,25 @@ def _authorization_config():
...
+        Get the Manager's REST configuration
+        :return: Manager's REST configuration data
+        """
+        request_args = request.args.to_dict(flat=False)
+        current_app.logger.info('Retrieving Manager REST config. filter: {0}'.
+                                format(request_args))
+        result = dict()
+        filter_by = request_args.get('filter', [])
+        for key in filter_by:
+            result[key] = getattr(config.instance, key)
```
mcouthon: Need to have default values and/or handle exceptions when `key` is not an attribute that the config has.

Better yet, instead of making it a general rest call that gets all of the config, maybe we should only return the configurations that we absolutely must.

barucoh: getattr's default value is None.
The reason behind returning everything is for future uses, I didn't want to create this REST endpoint just for this feature to retrieve the database IP, that's why I added the filter.

rest-service/manager_rest/rest/resources_v3_1/manager_config.py
barucoh: I agree, even though only a sys_admin can run this command.
So I think I'll create a subclass of Config in config.py that will override any unsafe information, that way it would seem like the key doesn't exist and we can still use the getattr function.

rest-service/manager_rest/rest/resources_v3_1/manager_config.py
mcouthon: No it's not - you'll get an `AttributeError`.

rest-service/manager_rest/rest/resources_v3_1/manager_config.py
mcouthon: I think it's more sensible to just have a `to_dict` method on the existing `Config` class.

rest-service/manager_rest/rest/resources_v3_1/manager_config.py
barucoh: Right, my bad, I looked in the wrong place.
fixed

rest-service/manager_rest/rest/resources_v3_1/manager_config.py
```diff
@@ -49,3 +49,25 @@ def _authorization_config():
...
+        Get the Manager's REST configuration
+        :return: Manager's REST configuration data
+        """
+        request_args = request.args.to_dict(flat=False)
+        current_app.logger.info('Retrieving Manager REST config. filter: {0}'.
+                                format(request_args))
+        result = dict()
+        filter_by = request_args.get('filter', [])
+        for key in filter_by:
+            result[key] = config.instance.to_dict().get(key, None)
```
tehasdf: None is default in dict.get

tehasdf: actually maybe a bad key requested could also result in a 400

rest-service/manager_rest/rest/resources_v3/cluster.py
```diff
@@ -112,6 +113,7 @@ def put(self, node_id, cluster):
             'host_ip': {'type': unicode},
             'node_name': {'type': unicode},
         })
+        details.update({'database_ip': config.instance.postgresql_host})
```
tehasdf: we have this here, so the additional request in the cloudify-common restclient PR isn't even needed?

rest-service/manager_rest/rest/resources_v3_1/manager_config.py
barucoh: Fixed

rest-service/manager_rest/rest/resources_v3/cluster.py
barucoh: Fixed, moved validations to the joining node's side rather than the Master node's

rest-service/manager_rest/rest/resources_v3_1/manager_config.py
```diff
@@ -49,3 +49,33 @@ def _authorization_config():
...
+        result = dict()
+        filter_by = request_args.get('filter', [])
+        for key in filter_by:
+            try:
+                result[key] = config.instance.to_dict()[key]
+            except KeyError as ke:
+                current_app.logger.error(
+                    'KeyError thrown: {0}'.format(ke.message))
+                raise CommandExecutionException(command='manager.config.get',
+                                                error=ke.message,
```
tehasdf: you'll be yelled at by python, which emits a warning that .message is deprecated
use just str(ke) i guess

barucoh: Fixed

PR #1502, opened 2018-10-22
CY-728 : removing the restriction of required node templates in a blueprint

tyacbovi:
- adding a test for installing an empty valid blueprint
- This was done because now you can import uploaded blueprints, so the full range of blueprints should be allowed


rest-service/dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==4.5.5.dev1
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-728#egg=cloudify-common==4.5.5.dev1
```
tehasdf: remember

PR #1508, opened 2018-10-28
Cy 687 make snapshot restore faster

adigrabow:


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -143,103 +147,115 @@ def restore(self):
...
+                                    deps_with_failed_plugins,
+                                    failed_deployments):
+        """
+        While in the correct workflow context, this method
+        creates a `create_deployment_env` task graph and executes it.
+        Since this method is being executed by threads, it appends errors
+        to thread-safe queues, in order to handle them later outside of
+        this scope.
+        """
+        with current_workflow_ctx.push(workflow_ctx, ctx_params):
```
mcouthon: Explain why this is necessary

PR #1517, opened 2018-11-06
Fix snapshot issues from permissions changes

geokala:


geokala:
amqp-postgres tests are failing because it's using the old key from the current live repo rather than its own branch (see amqp-postgres/tox.ini).

PR #1528, opened 2018-11-12
CY-782 - Stop using `admin_token` file in Cloudify Manager

qooban:
Related to: https://github.com/cloudify-cosmo/cloudify-stage/pull/689


workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -424,7 +419,6 @@ def stage_db_schema_get_current_revision():
     if version['edition'] != 'premium':
         return None
     output = subprocess.check_output([
-        'sudo', '-u', snapshot_constants.STAGE_USER,
```
geokala: I think these sudos are still needed, otherwise you don't haver permissions to access the migration.js, (and everything it accesses) as cfyuser (which the snapshot restore runs as).

workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -445,7 +439,6 @@ def composer_db_schema_get_current_revision():
     if version['edition'] != 'premium':
         return None
     output = subprocess.check_output([
-        'sudo', '-u', snapshot_constants.COMPOSER_USER,
```
geokala: I think these sudos are still needed, otherwise you don't haver permissions to access the migration.js, (and everything it accesses) as cfyuser (which the snapshot restore runs as).

workflows/cloudify_system_workflows/snapshots/utils.py
qooban: Oh, that's very strange, I don't recall removing that line. Reverting.

workflows/cloudify_system_workflows/snapshots/utils.py
qooban: The same here - I don't recall removing those lines. Reverting.

PR #1529, opened 2018-11-13
CY-772 Make filtered events total correct

geokala:


rest-service/manager_rest/rest/resources_v2/events.py
```diff
@@ -112,11 +108,24 @@ def get(self, _include=None, filters=None,
             for event in select_query.params(**params).all()
         ]
 
+        events_total = None
+        logs_total = None
+        for r in results:
```
mcouthon: I'm not sure I understand what's going on here - why are we iterating over all of those results? Is the `full_count` label added to _each_ result? If so, that's kind of weird, and also might make us iterate over all of this loop for nothing (for example, if all we have are 10000 events, with no logs)

rest-service/manager_rest/rest/resources_v1/events.py
```diff
@@ -337,6 +337,7 @@ def select_column(column_name):
                 select_column('level'),
                 literal_column("'cloudify_{}'".format(model.__name__.lower()))
                 .label('type'),
+                func.count('*').over().label('full_count'),
```
mcouthon: Why is adding this to each result better than running a single query?

geokala: Because we don't run a single query. What we currently do is:
1. Run a counting query (which differs enough from the main query to get different total results when some filters are used), using its result as the total.
2. Run the main query, which then obtains the amount of results specified from the defined offset.

Unfortunately, as I say, they give different results even without the limit and offset, because they are massively different queries.
Changing to do it with one query guarantees that we return specifically the total possible results at the time of this query.

In terms of efficiency of data transfer I don't think it'll have that great an effect- it's one extra integer (or possibly long), so at worst that should be 4 or 8 bytes. Every record returned already has at least seven bytes consumed by the 'cloudify_log' (or more for the event) type string, plus the message, the timestamps, the workflow ID, the blueprint ID, the operation, etc, etc, most of which are strings which means that we should be looking at a minimal increase in data transferred in exchange for getting consistent results and only making a single query each time (rather than two for every list request against the REST API).

rest-service/manager_rest/rest/resources_v2/events.py
geokala: We're iterating over all of the results because unfortunately we currently obtain the events and the logs in two separate queries.

I can think of a couple of ways to make this more efficient:
1. Probably the fastest and quite possibly the best approach, but also more complex and potentially error prone (though probably lower technical debt longer term): Retrieve both events and logs in a single query instead of two sub-queries, meaning that only the first result's full_count field needs to be retrieved.

2. Pick up from the parameters whether we need both logs_total and events_total, and set either of them to 0 if they're not needed before starting iterating over the results. This would mean that if you only asked for events then we'd only get the total for events, so we should stop after checking one record. Otherwise, we'd have to scan all of them anyway.

mcouthon: The second optimization is fine enough, but of course, the worst case scenario is left unaddressed. 
A general question - is there no way for us to use our default means of getting this data (i.e. the storage manager)? Why are we doing this whole crazy thing? Performance?

rest-service/manager_rest/rest/resources_v1/events.py
mcouthon: If we have the actual query that's getting the results, can't we do `query.count()` instead of `query.all()` to get the total (maybe excluding offset/limit params)?

rest-service/manager_rest/rest/resources_v2/events.py
geokala: The worst case scenario is /marginally/ improved. In the (unlikely) case that we have 10,000 events and 1 log (or 10,000 logs and 1 event), we'll indeed have to iterate over all of them to get the details.

This is for correctness more than performance- actually getting the total with the specified filters rather than using two different queries performed at slightly different times to get an incorrect total count.

rest-service/manager_rest/rest/resources_v1/events.py
geokala: The problem is that we need to use the offset/limit if we want to actually efficiently get a restricted set of results, otherwise we're going to load every single result into memory in the rest service every time, then slice them (taking a bit more memory) then convert them to json (taking even more memory), which potentially gets us back to the exact same problem we had before.

geokala: (the same problem that required us to introduce pagination, that is)

geokala:
This approach is currently untenable with the tests we are using, and the alternative approach of using the same query structure for both count and select is running into issues with sqlalchemy's query generation so I'm going to kill this and come back to it later.

PR #1535, opened 2018-11-15
Create REST token with the new security config file and restart REST

adigrabow:


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -148,6 +150,58 @@ def restore(self):
...
+            except Exception:
+                pass
+
+    def _get_new_token(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        script_path = os.path.join(dir_path, 'get_new_rest_token.py')
+        command = [MANAGER_PYTHON, script_path, self._tempdir]
+        utils.run(command)
+
+    def _get_token_from_file(self):
```
mcouthon: Maybe a quick explanation of why we expect the file to be there.

PR #1536, opened 2018-11-16
CY-773 Allow 'like' for event operation filter

geokala:


rest-service/manager_rest/rest/resources_v1/events.py
```diff
@@ -58,10 +58,13 @@ class Events(SecuredResource):
 
     DEFAULT_SEARCH_SIZE = 10000
 
+    # Allowed filters structure (added long after this dict was added so...
+    # hopefully it's right!)
```
tehasdf: @geokala please don't sow doubt in the mind of the reader :P it is right

geokala: But I'm trying to grow things!
(comment is updated already, was just reading through some more of the code)

PR #1546, opened 2018-11-26
CY-764 Agents create dosen't update when the agent exists

inbala:


rest-service/manager_rest/rest/resources_v3_1/agents.py
```diff
@@ -80,7 +81,11 @@ def put(self, name):
         try:
             return self._create_agent(name, state, request_dict)
         except manager_exceptions.ConflictError:
-            return self._update_agent(name, state)
+            # Assuming the agent was already created in cases of reinstalling
+            # or healing
+            current_app.logger.info("Can't create agent {0} because it "
```
mcouthon: Let's not sat "Can't", let's say "Not creating agent ..."

PR #1548, opened 2018-11-27
Use PostgreSQL in unittests

tehasdf:
Instead of sqlite, which doesn't support all the things postgres does
(eg. ARRAY type), and well, let's test with what we deploy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,13 +14,15 @@ defaults:
   - &tox_defaults
     docker:
       - image: circleci/python:2.7
-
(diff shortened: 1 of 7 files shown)


geokala:
Does the ordering fix also need applying to the rest service code or does it not share that bug?

tehasdf:
The ordering fix is just clientside here, because we pass `_sort=['id', 'blueprint_id']` instead of just `['id']` to the rest-client calls. So there's nothing to be done inside of restservice.

PR #1551, opened 2018-11-29
CY-818 Add support for Operation and TasksGraph models

tehasdf:


rest-service/manager_rest/resource_manager.py
```diff
@@ -483,6 +483,37 @@ def delete_deployment(self,
         else:
             return self.sm.delete(deployment)
 
+    def resume_execution(self, execution_id):
+        execution = self.sm.list(
+            models.Execution, filters={'id': execution_id},
+            get_all_results=True)[0]
+        if execution.status in ExecutionState.END_STATES or \
```
geokala: Is there a reason this wouldn't be better checking for only allowed states and then raising an error stating what states one is allowed to resume an execution in?

rest-service/manager_rest/resource_manager.py
```diff
@@ -1339,6 +1370,54 @@ def add_node_instance_from_dict(self, instance_dict):
         instance_dict['resource_availability'] = resource_availability
         instance_dict['private_resource'] = private_resource
 
+    def create_operation(self, operation_id, name, dependencies,
+                         parameters, type, graph_id=None,
+                         graph_storage_id=None):
+        # allow passing graph_storage_id directly as an optimization, so that
+        # we don't need to query for the graph based on display id
+        if (graph_id and graph_storage_id) or \
+                (not graph_id and not graph_storage_id):
```
geokala: Can't we use xor (^?) here?

geokala: (well- not (graph_id ^ graph_storage_id))

rest-service/manager_rest/rest/endpoint_mapper.py
```diff
@@ -99,7 +99,11 @@ def setup_resources(api):
         'ManagerConfig': 'config',
         'ManagerRestConfig': 'rest/config',
         'Agents': 'agents',
-        'AgentsName': 'agents/<string:name>'
+        'AgentsName': 'agents/<string:name>',
```
geokala: You probably need to rebase this after my earlier commit- and allow it through the nginx config.

rest-service/manager_rest/rest/endpoint_mapper.py
```diff
@@ -99,7 +99,11 @@ def setup_resources(api):
...
+        'Operations': 'operations',
+        'OperationsId': 'operations/<string:operation_id>',
+        'TasksGraphs': 'tasks_graphs',
+        'TasksGraphsId': 'tasks_graphs/<string:tasks_graph_id>'
```
geokala: Trailing comma would be nice so that the next person to update this doesn't have it listed as them having changed one of your lines.

rest-service/manager_rest/storage/models_base.py
```diff
@@ -130,7 +130,9 @@ class SQLModelBase(db.Model):
         'PickleType': flask_fields.Raw,
         'UTCDateTime': flask_fields.String,
         'Enum': flask_fields.String,
-        'Boolean': flask_fields.Boolean
+        'Boolean': flask_fields.Boolean,
+        'ARRAY': flask_fields.Raw,
+        'JSONString': flask_fields.Raw
```
geokala: Trailing comma here would also be nice (same reasons as above)

rest-service/manager_rest/resource_manager.py
tehasdf: with this PR, let's not, but stick with the "if it works, it works". With another PR, we'll be adding resume just for cancelled and failed

rest-service/manager_rest/resource_manager.py
tehasdf: i think this is more readable, because the xor would require bool() conversions, because if the ids are ints, then xor might well return zero even if both are passed (if they're equal)

PR #1555, opened 2018-12-02
CY-642: import blueprint improvements 

tyacbovi:


rest-service/manager_rest/resource_manager.py
```diff
@@ -398,13 +398,12 @@ def delete_blueprint(self, blueprint_id, force):
...
         if not force:
-            imported_list = [b.plan[constants.IMPORTED]
-                             for b in self.sm.list(
-                    models.Blueprint,
-                    include=['id', 'plan'])]
-            import_blueprint_format = "blueprint:{}".format(blueprint_id)
-            for imported in imported_list:
-                if import_blueprint_format in imported:
+            imported_blueprints_list = [b.plan[constants.IMPORTED_BLUEPRINTS]
+                                        for b in self.sm.list(
```
mcouthon: Why list and not get? Don't we have the unique ID here?

tyacbovi: Correct, after the last improvement this will work

PR #1562, opened 2018-12-04
CY-860: disregard the default executor when installing a wagon; if th…

isaac-s:
…e platform/distro matches the one of the manager, then install it on the manager


tehasdf:
@isaac-s I've rebased this, and re-added the yaml file validation (not the executor check)

PR #1576, opened 2018-12-11
CY-642: adding a mark for cloudify types

tyacbovi:


resources/rest-service/cloudify/types/types.yaml
```diff
@@ -4,6 +4,12 @@
 ##################################################################################
 tosca_definitions_version: cloudify_dsl_1_3
 
+##################################################################################
+# Cloudify basic types metadata
+##################################################################################
+metadata:
+  cloudify_types: {}
```
tehasdf: this needs to be a dict? maybe just `true`?

PR #1583, opened 2018-12-19
Snapshot restore: recreate mgmtworker admin token

tehasdf:


tehasdf:
this effectively reverts #1528 (with a different purpose token though)

PR #1586, opened 2018-12-19
CY-760 Include workflow details with _include

geokala:


rest-service/manager_rest/rest/resources_v2/deployments.py
```diff
@@ -70,6 +70,19 @@ def get(self, _include=None, filters=None, pagination=None, sort=None,
             get_all_results=get_all_results
         )
 
+        if _include and 'workflows' in _include:
+            # Because we coerce this into a list in the model, but our ORM
+            # won't return a model instance when filtering results, we have
+            # to coerce this here as well. This is unpleasant.
+            for item_idx in xrange(len(result.items)):
```
tehasdf: let's not use xrange or range (wrt. python 2+3 compat, but also it's bad style anyway), but let's do `for index, item in enumerate(result.items)` instead. Possibly create a new list as well

PR #1602, opened 2019-01-01
Moved parse logic to CLI

adigrabow:


rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -252,14 +251,9 @@ def get_visibility_parameter(optional=False,
 
 def parse_datetime(datetime_str):
```
inbala: After this change, maybe the function is redundant?

PR #1606, opened 2019-01-03
CY-934 Make import plugin version field to comply with PEP440

davidginzbourg:


rest-service/manager_rest/resolver_with_catalog_support.py
```diff
@@ -100,16 +101,57 @@ def _resolve_plugin_yaml_url(self, import_url):
...
+            for spec in _versions:
+                if not spec:
+                    raise InvalidSpecifier()
+                try:
+                    specs &= SpecifierSet(spec)
+                except InvalidSpecifier:
+                    # If the code below doesn't raise any exception then it's
+                    # the case where a version has been provided with no
+                    # operator to prefix it.
+                    specs &= SpecifierSet('==={}'.format(spec))
```
tehasdf: should this be `===` (three equals) or `==` (two equals)?

davidginzbourg: In the old version it would pick the plugin with exactly what's inside `spec`. Therefore `===`.

PR #1619, opened 2019-01-14
Automated workflow resume: basic integration tests

tehasdf:


tests/integration_tests/tests/agentless_tests/test_resume.py
```diff
@@ -0,0 +1,116 @@
...
+    def _wait_for_log(self, execution):
+        self.logger.info('Waiting for operation to start')
+        while True:
+            logs = self.client.events.list(
+                execution_id=execution.id, include_logs=True)
+            if any(self.wait_message == log['message'] for log in logs):
+                break
+            time.sleep(1)
+
+    def _restart_mgmtworker(self):
```
geokala: This should probably be renamed to indicate why it's placing the target_file; I can understand why having read the rest of this code in context, but it'll likely be less clear in future.

PR #1628, opened 2019-01-25
CY-977: adding test for local namespaced blueprint imports

tyacbovi:


tests/integration_tests/tests/agent_tests/test_namespaced_scripts.py
```diff
@@ -40,3 +40,22 @@ def test_success_deploy_namespaced_blueprint_with_scripts(self):
...
+
+    def test_success_deploy_namespaced_blueprint_with_local_scripts(self):
+        deployment_id = 'dep'
+        dsl_path = resource(
+            'dsl/agent_tests/'
+            'blueprint_with_namespaced_local_blueprint_import.yaml')
+        _, execution_id = self.deploy_application(dsl_path,
+                                                  deployment_id=deployment_id)
+
+        events = self.client.events.list(execution_id=execution_id,
```
adigrabow: Why use events list instead of waiting for the 'install' workflow to terminate?
You can use `wait_for_execution_to_end`.

PR #1640, opened 2019-02-12
CY-1048 amqp_postgres: don't break on logs containing NULs

tehasdf:


amqp-postgres/amqp_postgres/postgres_publisher.py
```diff
@@ -120,6 +120,14 @@
 """
 
 
+def _strip_nul(text):
+    """Remove NUL values from the text, so that it can be treated as text
+
+    There should likely be no NUL values in human-readable logs anyway.
+    """
+    return text.replace('\x00', '?')
```
geokala: Maybe replace it with \\NUL or something similar?

PR #1645, opened 2019-02-18
New DB model and migration file

adigrabow:


rest-service/manager_rest/storage/management_models.py
```diff
@@ -475,4 +475,15 @@ def _get_identifier_dict(self):
...
 
 
+class License(SQLModelBase):
+    __tablename__ = 'licenses'
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    customer_id = db.Column(db.Text, unique=True)
+    license_type = db.Column(db.String(255))
+    expiration_date = db.Column(UTCDateTime)
+    cloudify_version = db.Column(db.Text)
+    capabilities = db.Column(db.PickleType)
```
tehasdf: if possible let's not add more pickles to the db, can we use the JSON type? we have a column type using it defined in `manager_rest.storage.models_base.JSONString`

PR #1651, opened 2019-02-25
CY-1081 amqp-postgres: don't break on ProgrammingError

tehasdf:


amqp-postgres/amqp_postgres/postgres_publisher.py
```diff
@@ -197,9 +197,9 @@ def _message_publisher(self):
                     self._store(conn, items)
                 except psycopg2.OperationalError as e:
                     self.on_db_connection_error(e)
-                except (psycopg2.IntegrityError, ValueError):
-                    logger.exception('Error storing %d logs+events',
-                                     len(items))
+                except Exception:
+                    logger.info('Error storing %d logs+events in batch',
```
tehasdf: any exception will make this go into the no-batch mode; the errors are then analyzed there, if they happen again, no need to print the whole traceback here

amqp-postgres/amqp_postgres/postgres_publisher.py
```diff
@@ -282,6 +282,18 @@ def _store_nobatch(self, conn, items):
             except (psycopg2.IntegrityError, ValueError):
                 logger.debug('Error storing %s: %s', exchange, item)
                 conn.rollback()
+            except psycopg2.ProgrammingError as e:
+                if e.pgcode == psycopg2.errorcodes.UNDEFINED_COLUMN:
```
tehasdf: undefined column happens often on snap-restore, it's not really a surprising condition

amqp-postgres/amqp_postgres/postgres_publisher.py
```diff
@@ -282,6 +282,18 @@ def _store_nobatch(self, conn, items):
...
+                    logger.debug('Error storing %s: %s (undefined column)',
+                                 exchange, item)
+                else:
+                    logger.exception('Error storing %s: %s (ProgrammingError)',
```
tehasdf: any other ProgrammingError is unexpected, log it loudly...

PR #1658, opened 2019-02-27
CY-1084 Generate and authenticate the Execution Token

inbala:
* Create a random string as the token when an execution starts

* Store the token hashed in the DB

* Authenticate The Execution Token : make sure that the token passed exists and valid.
  Valid token is connected to an active execution

### DIFF
--- a/rest-service/manager_rest/security/user_handler.py
+++ b/rest-service/manager_rest/security/user_handler.py
@@ -24,6 +24,9 @@ from cloudify.constants import CLOUDIFY_API_AUTH_TOKEN_HEADER
 from manager_rest.storage.models import User
 from manager_rest.manager_exceptions import NotFoundError
 from manager_rest.storage import user_datastore, get_storage_manager
+from manager_rest.execution_token import (set_current_execution,
+                                          get_current_execution_by_token,
+                                          get_execution_token_from_request)
 
 
 ENCODED_ID_LENGTH = 5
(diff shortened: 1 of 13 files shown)


rest-service/manager_rest/security/user_handler.py
```diff
@@ -43,6 +44,10 @@ def user_loader(request):
     """
     if request.authorization:
         return get_user_from_auth(request.authorization)
+    execution_token = get_execution_token_from_request(request)
+    if execution_token:
+        execution = get_current_execution_by_token(execution_token)
```
tehasdf: this has to be here, can't be inside of the authentication.py instead? currently the auth with execution token flow is split between those two places, harder to follow

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -564,6 +564,12 @@ def count(self, model_class, filters=None):
         count = query.order_by(None).count()  # Fastest way to count
         return count
 
+    def admin_list(self, model_class, filters=None):
```
tehasdf: this does very little compared to the regular `list` method, but i suppose we'll add features (eg. joins, sort) as they're needed?

rest-service/manager_rest/workflow_executor.py
```diff
@@ -113,6 +118,17 @@ def execute_system_workflow(wf_id,
                          context=context, execution_creator=execution_creator)
 
 
+def generate_execution_token(execution_id):
+    sm = get_storage_manager()
+    execution = sm.get(models.Execution, execution_id)
+    execution_token = uuid.uuid4().hex
+
+    # Store the token hashed in the DB
+    execution.token = hash_data(execution_token)
```
tehasdf: tbh there's not much point hashing a uuid4 i don't think, so if we wanted to, we could skip that...

rest-service/manager_rest/security/user_handler.py
inbala: Yes I agree, but the current_user assignment is here

rest-service/manager_rest/storage/storage_manager.py
inbala: Maybe it is better to add a parameter like "skip_permissions" to the regular `list` that says don't consider tenant and user?

PR #1666, opened 2019-03-03
Load config on first use

tehasdf:


resources/rest-service/cloudify/migrations/versions/423a1643f365_4_6_to_5_0.py
```diff
@@ -22,28 +26,181 @@
 depends_on = None
 
 
+Base = declarative_base()
+
+
+class Config(Base):
```
tehasdf: the model is also written in the migration scrip, so that we can use the ORM to insert values in the migration

rest-service/manager_rest/test/base_test.py
```diff
@@ -59,6 +60,11 @@
 )
 
 
+MIGRATION_DIR = os.path.normpath(os.path.join(
```
tehasdf: this is unfortunate, but we must use migrations to set up the db in unittests, not create_all

rest-service/manager_rest/config.py
```diff
@@ -25,67 +25,110 @@
     ('MANAGER_REST_AUTHORIZATION_CONFIG_PATH', 'authorization')
 ]
 SKIP_RESET_WRITE = ['authorization']
+NOT_SET = object()
 
 
-class Config(object):
+class Setting(object):
```
tehasdf: a descriptor - like a `@property`, but will load from db on first access (unless set first)

rest-service/manager_rest/resource_manager.py
```diff
@@ -145,8 +145,6 @@ def _get_conf_for_snapshots_wf():
         config_instance = config.instance
         return {
             'file_server_root': config_instance.file_server_root,
-            'db_address': config_instance.db_address,
```
geokala: Why do we not need to return these here any more?

tehasdf: because they're not used

geokala: Fair enough, approved then. <smiles>

PR #1669, opened 2019-03-04
Config endpoint: add a way to get authorization

tehasdf:


rest-service/manager_rest/rest/resources_v3_1/manager_config.py
```diff
@@ -19,39 +19,59 @@
 from flask_security import current_user
 from flask_restful.reqparse import Argument
 
+
```
geokala: Have blank line, will travel.

PR #1696, opened 2019-03-19
Add license integration tests

adigrabow:


tests/integration_tests/framework/constants.py
```diff
@@ -38,3 +38,5 @@
                             " cloudify_version, capabilities, signature)" \
                             " VALUES('mock_customer', '2050-01-01', 'Spire'," \
                             " false, '4.6', '{mock}', 'mock_signature');"
+
+DELETE_CLOUDIFY_LICENSE_QUERY = "DELETE FROM licenses"
```
inbala: If you use it only once I don't think it should be a constant

tests/integration_tests/framework/docl.py
```diff
@@ -146,6 +149,11 @@ def upload_mock_license():
              '-c "{0}"'.format(INSERT_MOCK_LICENSE_QUERY)))
 
 
+def remove_license():
+    execute(('sudo -u postgres psql cloudify_db '
+             '-c "{0}"'.format(DELETE_CLOUDIFY_LICENSE_QUERY)))
```
inbala: I think it is better to use 
`from integration_tests.framework import postgresql`
`postgresql.run_query("DELETE FROM licenses;")`

tests/integration_tests/tests/agentless_tests/test_license.py
```diff
@@ -0,0 +1,106 @@
...
+        maintenance.
+        """
+        self.client.tokens.get()
+        self.client.tenants.list()
+        self.client.license.list()
+        self.client.snapshots.list()
+        self.client.manager.get_status()
+        self.client.maintenance_mode.status()
+
+    def _upload_license(self, license):
```
inbala: I usually put the private methods at the end of the file

PR #1702, opened 2019-03-24
Component improvements

tyacbovi:
- Refactor unit tests framework and tests
- Refactor methods for retrieving blueprint and executions
- Improve log message wording and levels
- Remove the remainder of unsupported features (reexecute and deployment outputs)
- Removing not relevant integration tests (and flakey)


cloudify_types/cloudify_types/component/polling.py
```diff
@@ -38,7 +38,7 @@ def blueprint_id_exists(client, blueprint_id):
     """
     blueprint_id_filter = {'id': blueprint_id}
     blueprint = client.blueprints.list(_include=['id'],
-                                       filters=blueprint_id_filter)
+                                       **blueprint_id_filter)
```
inbala: Why don't you use the filters?

tyacbovi: I use them, but via the http client you pass them differently them in the resource manager (pass them like I did )

PR #1717, opened 2019-04-01
CY-1176 Add test_different_plugin_versions

glukhman:
added test:
**the plugin is installed with more than one version**
=> make sure we use the latest version

the test **the plugin is not installed** already exists

### DIFF
--- a/tests/integration_tests/resources/plugins/target-aware-mock-v1_33/plugin.yaml
+++ b/tests/integration_tests/resources/plugins/target-aware-mock-v1_33/plugin.yaml
@@ -1,0 +1,4 @@
+plugins:
+   cool:
+     executor: central_deployment_agent
+     package_name: target-aware-mock
--- a/tests/integration_tests/tests/agentless_tests/test_hooks.py
+++ b/tests/integration_tests/tests/agentless_tests/test_hooks.py
@@ -180,8 +180,7 @@ hooks:
     description: test hook
 """
         upload_mock_plugin('target-aware-mock',
-                           '1.0',
-                           'plugins/target-aware-mock')
+                           '1.0',)
 
         self._update_hooks_config(new_config)
         self._start_a_workflow()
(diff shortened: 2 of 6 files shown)


tests/integration_tests/tests/agentless_tests/test_hooks.py
```diff
@@ -191,6 +190,36 @@ def test_implementation_plugin(self):
...
+          input2: input2_test
+        description: test hook
+    """
+        old_version = '1.0'
+        new_version = '1.33'
+
+        # Upload the new one before the old, on purpose
+        upload_mock_plugin('target-aware-mock', new_version,)
+        upload_mock_plugin('target-aware-mock', old_version,)
+
```
inbala: It would be nice to add a plugins list query with the package name `target-aware-mock` and see we get both of the versions.

tests/integration_tests/resources/plugins/target-aware-mock-v1_33/plugin.yaml
```diff
@@ -0,0 +1,4 @@
+plugins:
+   cool:
+     executor: central_deployment_agent
+     package_name: target-aware-mock
```
inbala: Add newline at the end of the file

tests/integration_tests/tests/agentless_tests/test_hooks.py
```diff
@@ -191,6 +190,42 @@ def test_implementation_plugin(self):
...
+        # Upload the new one before the old, on purpose
+        upload_mock_plugin('target-aware-mock-v1_33', new_version)
+        upload_mock_plugin('target-aware-mock', old_version)
+
+        self._update_hooks_config(new_config)
+        self._start_a_workflow()
+        event_type_msg = "workflow_started"
+        workflow_id_msg = "create_deployment_environment"
+        input1_msg = "input1_test"
+        newest_version_msg = "version 1.33"
```
inbala: Maybe add a message explaining we verify that the new version of the plugin was called

PR #1732, opened 2019-04-14
CY-1182 cancel kill integration test

YoniItzhak:


tests/integration_tests/tests/agentless_tests/test_executions.py
```diff
@@ -603,6 +604,31 @@ def test_cancel_execution_and_then_force_cancel(self):
...
+        The function executes a new execution, writes the process to a local
+        file, kills the execution and asserts the process was also killed.
+        """
+        execution, node_inst_id, deployment_id = \
+            self._execute_from_resource(workflow_id='read_write',
+                                        workflow_params=None,
+                                        resource_file='dsl'
+                                                      '/read_write_workflows'
+                                                      '.yaml')
+        self._check_state('active_and_wrote', "Execution was expected to go "
```
adigrabow: What if by the time we arrive to this line the status changes to `asleep`? 
Not sure if this is needed here..

tests/integration_tests/tests/agentless_tests/test_executions.py
```diff
@@ -603,6 +604,31 @@ def test_cancel_execution_and_then_force_cancel(self):
...
+                                              "into 'active_and_wrote' "
+                                              "status", node_inst_id)
+        self._check_state('asleep', "Execution was expected to go"
```
adigrabow: Modify _check_state to use `assertEquals` instead of sending the error message

tests/integration_tests/tests/agentless_tests/test_executions.py
```diff
@@ -603,6 +604,31 @@ def test_cancel_execution_and_then_force_cancel(self):
...
+                                    " into 'sleeping' status", node_inst_id)
+        content = self.client.node_instances.get(node_inst_id). \
+            runtime_properties['file_content']
+        execution = self.client.executions.cancel(execution.id, force=True,
+                                                  kill=True)
+        expected_status = Execution.KILL_CANCELLING
+        self.assertEquals(expected_status, execution.status)
```
adigrabow: How can you be sure the execution is already killed? Maybe add a retry or sleep

tests/integration_tests/tests/agentless_tests/test_executions.py
```diff
@@ -737,11 +763,19 @@ def test_update_execution_status(self):
         self.assertEquals(Execution.TERMINATED, execution.status)
         self.assertEquals('', execution.error)
 
-    def _execute_and_cancel_execution(self, workflow_id, force=False,
-                                      wait_for_termination=True,
-                                      is_wait_for_asleep_node=True,
-                                      workflow_params=None):
-        dsl_path = resource('dsl/sleep_workflows.yaml')
+    def _check_state(self, expected_instance_state, error_msg, node_inst_id,
+                     retry_range=30, sleep_time=1):
```
adigrabow: Can remove retry_range, error_msg and sleep_time

tests/integration_tests/resources/dsl/read_write_workflows.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - cloudify/types/types.yaml
+    - plugins/testmockoperations.yaml
+    - plugins/mock_workflows.yaml
+
+node_templates:
+    test_node:
+        type: cloudify.nodes.Root
+        interfaces:
+            test_interface:
+                write_operation: testmockoperations.testmockoperations.tasks.write_process_id_to_kill_process
```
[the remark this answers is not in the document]
YoniItzhak: I changed it to write_pid and it writes the current pid to a runtime property. Moreover, I don't use read_pid and changed sleep_operation --> sleep.

tests/integration_tests/tests/agentless_tests/test_executions.py
YoniItzhak: I tried to do it, but I couldn't find a way to retry the assertion process. So I thought of a work-around that still sends a Runtime error but in a basic format (kind of what David did). Tell me what you think. 

    def _check_state(self, expected_instance_state, node_inst_id):
        instance = self.client.node_instances.get(node_inst_id)
        for retry in range(30):
            if self.assertEquals(instance.state, expected_instance_state):
                break
            time.sleep(1)
        else:
            raise RuntimeError('Expected instance state is: {},'
                               'but the actual state is: {}'.format(
                                expected_instance_state, instance.state))

tests/integration_tests/tests/agentless_tests/test_executions.py
YoniItzhak: Added.
        for retry in range(10):
            if self.assertEquals(expected_status, execution.status):
                break
        else:
            raise AssertionError("Process is not killed")

tests/integration_tests/tests/agentless_tests/test_executions.py
YoniItzhak: I think it is needed to make sure the pid was written as a runtime property, but I'll be glad to talk about it and see if there's maybe something I don't understand

tests/integration_tests/tests/agentless_tests/test_executions.py
```diff
@@ -603,6 +605,28 @@ def test_cancel_execution_and_then_force_cancel(self):
...
+        dep = self.deploy(dsl_path, wait=False, client=self.client)
+        do_retries(verify_deployment_env_created, 30, deployment_id=dep.id)
+        execution = self.client.executions.start(deployment_id=dep.id,
+                                                 workflow_id='install')
+        pid = do_retries(docl.read_file, file_path='/tmp/pid.txt')
+        path = '/proc/{}/status'.format(pid)
+        execution = self.client.executions.cancel(execution.id,
+                                                  force=True, kill=True)
+        self.assertEquals(Execution.KILL_CANCELLING, execution.status)
+        # if the process is still running, docl.read_file will raise an error
```
adigrabow: if the process is still running docl.read_file will raise an error. We use do_retries to give the `kill cancel` operation time to kill the process.

PR #1733, opened 2019-04-14
Cy 1183 fix ldap set configuration

glukhman:


rest-service/manager_rest/config.py
```diff
@@ -182,6 +183,37 @@ def load_from_db(self):
...
+            If a schema is specified for the given setting, the new value
+            must validate.
+            """
+        from datetime import datetime
+        from flask_security import current_user
+        from manager_rest.storage import get_storage_manager, models
+        from manager_rest.manager_exceptions import ConflictError
+
+        sm = get_storage_manager()
+        names = [key for key in config_dict if not config_dict[key] is None]
```
inbala: Can be `config_dict.keys()`, why would the value be `None`?
Maybe we want to update the value to `None`

rest-service/manager_rest/config.py
```diff
@@ -182,6 +183,37 @@ def load_from_db(self):
...
+                raise ConflictError('{0} is not editable'.format(item.name))
+            if item.schema:
+                try:
+                    jsonschema.validate(config_dict[item.name], item.schema)
+                except jsonschema.ValidationError as e:
+                    raise ConflictError(e.args[0])
+            item.value = config_dict[item.name]
+            item.updated_at = datetime.now()
+            item.updated_by = current_user
+            sm.update(item)
```
inbala: It runs update query for each one, I think it is better to use `orm.Session` like in the function `load_from_db` so we will have one query to get all the values and second query to update all of them.

tests/integration_tests/resources/scripts/set_ldap.py
```diff
@@ -0,0 +1,53 @@
...
+
+from manager_rest import config
+from manager_rest.flask_utils import setup_flask_app, set_admin_current_user
+
+
+def set_ldap(config_dict):
+    app = setup_flask_app()
+    # mock current user, and reload rest configuration
+    set_admin_current_user(app)
+    config.instance.load_from_file("/opt/manager/cloudify-rest.conf")
```
inbala: Instead of the string `/opt/manager/cloudify-rest.conf`, maybe use the constant `CONFIG_FILE_LOCATION` that is in `from integration_tests.framework import constants`

tests/integration_tests/resources/scripts/set_ldap.py
```diff
@@ -0,0 +1,53 @@
...
+import socket
+import argparse
+
+from manager_rest import config
+from manager_rest.flask_utils import setup_flask_app, set_admin_current_user
+
+
+def set_ldap(config_dict):
+    app = setup_flask_app()
+    # mock current user, and reload rest configuration
```
inbala: Capital letter in the beginning of comments, also in line 39

tests/integration_tests/resources/scripts/set_ldap.py
```diff
@@ -0,0 +1,53 @@
...
+    set_admin_current_user(app)
+    config.instance.load_from_file("/opt/manager/cloudify-rest.conf")
+
+    # Update the config table on manager DB to include LDAP configurations
+    config.instance.update_db(config_dict)
+
+    # Restart the rest service to load the new LDAP configuration
+    os.system('systemctl stop cloudify-restservice')
+    os.system('systemctl start cloudify-restservice')
```
inbala: I think one command of `restart` is fine
`os.system('systemctl restart cloudify-restservice')`

PR #1737, opened 2019-04-15
Snapshots: update the networks check to use the db

tehasdf:


workflows/cloudify_system_workflows/snapshots/networks.py
```diff
@@ -19,75 +19,62 @@
...
-    def get_networks_from_snapshot(tempdir):
-        networks_dump_path = join(tempdir, Networks._NETWORKS)
-        with open(networks_dump_path, 'r') as f:
-            return json.load(f)
+def get_current_networks(client):
+    broker_networks = set()
+    manager_networks = set()
+    for manager in client.manager.get_managers():
+        manager_networks.update(manager.networks or {})
+    for manager in client.manager.get_brokers():
```
barucoh: Would make more sense to put 'broker' when using 'get_brokers'

PR #1757, opened 2019-04-24
CY-1184 Add nested levels to migration file

glukhman:


resources/rest-service/cloudify/migrations/versions/423a1643f365_4_6_to_5_0.py
```diff
@@ -124,6 +124,13 @@ def upgrade():
             schema={'type': 'number'},
             is_editable=True
         ),
+        Config(
+            name='ldap_nested_levels',
+            value=1,
```
inbala: Are we sure we want the default to be 1?

rest-service/manager_rest/rest/resources_v3/manager.py
```diff
@@ -176,8 +176,8 @@ def _validate_set_ldap_request(self):
             'ldap_dn_extra': {}
         })
         # Not allowing empty username or password
-        ldap_config['ldap_username'] = ldap_config.get('ldap_username') or None
-        ldap_config['ldap_password'] = ldap_config.get('ldap_password') or None
+        ldap_config['ldap_username'] = ldap_config.get('ldap_username') or ''
+        ldap_config['ldap_password'] = ldap_config.get('ldap_password') or ''
```
inbala: Can be changed to `ldap_config['ldap_password'] = ldap_config.get('ldap_password', '')`

PR #1786, opened 2019-05-09
Get details on snapshot DB restore failure

geokala:


tehasdf:
patch as cloudify-cosmo/patchify#86

PR #1793, opened 2019-05-14
Cloudify types update

tyacbovi:
- Updating component type definition and depends on lifecycle op

### DIFF
--- a/resources/rest-service/cloudify/types/types.yaml
+++ b/resources/rest-service/cloudify/types/types.yaml
@@ -470,7 +470,6 @@ relationships:
       operation:
         description: The target's lifecycle operation name.
         type: string
-        default: precreate
 
   cloudify.relationships.file_system_depends_on_volume:
     derived_from: cloudify.relationships.depends_on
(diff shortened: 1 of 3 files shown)


resources/rest-service/cloudify/types/types.yaml
```diff
@@ -370,31 +370,20 @@ node_types:
...
-            headers: Headers to be added to request.
-            query_params: Query parameters to be added to the request.
-            cert: Path to a copy of the server's self-signed certificate.
-            trust_all: if `False`, the server's certificate
-                       (self-signed or not) will be verified.
-            username: Cloudify User username.
-            password: Cloudify User password.
-            token: Cloudify User token.
-            tenant: Cloudify Tenant name.
+          Client configuration, if empty will be reused manager client.
```
inbala: Maybe should be `Client configuration, if empty manager client will be reused.`

resources/rest-service/cloudify/types/types.yaml
```diff
@@ -370,31 +370,20 @@ node_types:
...
+        type: cloudify.datatypes.RemoteCloudifyManagerClient
         default: {}
       plugins:
         description: >
           Optional, list of plugins for upload.
```
inbala: Maybe update the description to describe the dict format

PR #1802, opened 2019-05-22
CY-1053 Add methods to broker management endpoint

geokala:


rest-service/manager_rest/rest/resources_v3_1/manager.py
```diff
@@ -229,6 +229,60 @@ class RabbitMQBrokers(SecuredResource):
...
         )
+
+    @exceptions_handled
+    @authorize('broker_add')
+    @marshal_with(models.RabbitMQBroker)
+    def post(self):
+        """Add a broker to the database."""
+        _broker = rest_utils.get_json_and_verify_params({
+            'name': {'type': unicode},
+            'address': {'type': unicode},
```
barucoh: Add validation to ensure `address` is in `networks`?

rest-service/manager_rest/rest/resources_v3_1/manager.py
```diff
@@ -201,6 +209,7 @@ def delete(self):
             current_app.logger.info('Manager deleted successfully')
             managers_list = get_storage_manager().list(models.Manager)
             remove_manager(managers_list)  # Removing manager from cluster
+        update_agents(sm)
```
tehasdf: it's unfortunate to have this after the db-commit, because this does a LOT of things (amqp connection etc) and so can throw a lot of exceptions.
Maybe we can do something like remove the manager and re-throw if it couldn't be sent. But also what about partially-sent updates...
I dont think we can do anything really sensible here but give it a think

rest-service/manager_rest/rest/resources_v3_1/manager.py
```diff
@@ -223,12 +232,68 @@ def _get_ca_cert_id(self, sm, _manager):
...
+        _broker = rest_utils.get_json_and_verify_params({
+            'name': {'type': unicode},
+            'address': {'type': unicode},
+            'port': {'type': int, 'optional': True},
+            'networks': {'type': dict, 'optional': True},
+        })
+        sm = get_storage_manager()
+
+        # Get the first broker in the list to get the ca_cert and credentials
+        first_broker = sm.list(models.RabbitMQBroker, None, None).items[0]
```
tehasdf: no need for the additional `None`, that's just asking for trouble

rest-service/manager_rest/rest/resources_v3_1/manager.py
```diff
@@ -223,12 +232,68 @@ def _get_ca_cert_id(self, sm, _manager):
...
+
+        new_broker = models.RabbitMQBroker(
+            name=_broker['name'],
+            host=_broker['address'],
+            management_host=_broker['address'],
+            port=_broker.get('port'),
+            networks=_broker['networks'],
+            username=first_broker.username,
+            password=first_broker.password,
+            ca_cert=first_broker.ca_cert,
```
tehasdf: we might as well add the same ca-cert handling (with `get_ca_cert_id` etc) as we have in the managers?
What is the part that breaks if rabbitmqs have different cacerts? Agents will survive it.
Creation of the rabbitmq cluster itself breaks?

rest-service/manager_rest/rest/resources_v3_1/manager.py
tehasdf: in the future i think we should get rid of `address` entirely because (even currently) it does nothing, doesn't it...

rest-service/manager_rest/rest/resources_v3_1/manager.py
```diff
@@ -223,12 +232,68 @@ def _get_ca_cert_id(self, sm, _manager):
...
+        )
+        result = sm.put(new_broker)
+        current_app.logger.info('Broker added successfully')
+        update_agents(sm)
+        return result
+
+    @exceptions_handled
+    @authorize('broker_delete')
+    @marshal_with(models.RabbitMQBroker)
+    def delete(self):
```
tehasdf: unfortunate that this is using the same weird handling as `Managers`, ie. request body contains the id to be deleted, as opposed to `DELETE /brokers/broker-123`. You think perhaps we can make this that, or do you want consistency with managers (but not with anything else in the system)?
Either way, we should also change managers to be the other way

rest-service/manager_rest/rest/resources_v3_1/manager.py
```diff
@@ -223,12 +232,68 @@ def _get_ca_cert_id(self, sm, _manager):
...
+        })
+        sm = get_storage_manager()
+        broker_to_delete = sm.get(
+            models.RabbitMQBroker,
+            None,
+            filters={'name': _broker['name']}
+        )
+
+        result = sm.delete(broker_to_delete)
+        if _broker['name'] == result.name:
```
tehasdf: why would that ever be false?!?!

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: We could notify and provide instructions on which endpoint to call to fix it (do we have an endpoint that triggers update-agents?)

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: Ah- filters defaults to None?

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: I think the rabbitmq clustering will be broken, yes. But agreed, I'll switch over to the method used there- for some reason the manager I originally wrote this on had slightly outdated code (maybe from yesterday)

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: Agreed, I was thinking we should change it to be properly restful.

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: I don't know- I just copied the approach taken in managers. Was intending to ask about that: How can either of them (broker or manager check) be false, and... what do we do if they are? Presumably that means it deleted something it shouldn't have?

rest-service/manager_rest/rest/resources_v3_1/manager.py
tehasdf: Yea both do

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: Note: we'll be validating this.

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: Update agents to be added to an endpoint+CLI (JIRA needed)

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: So can just do sm.list(models.RabbitMQBroker)?

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: After further discussion we've agreed that since we're not going to really support (or test) this use-case, it won't be added.

rest-service/manager_rest/rest/resources_v3_1/manager.py
geokala: This will go into a separate PR

PR #1811, opened 2019-05-25
CY-1316 Mgmtworker: get brokers from REST on startup

tehasdf:


mgmtworker/mgmtworker/worker.py
```diff
@@ -201,6 +204,30 @@ def make_amqp_worker(args):
     return AMQPConnection(handlers=handlers, connect_timeout=None)
 
 
+def prepare_broker_config():
+    client = get_rest_client(
+        tenant='default_tenant', api_token=get_admin_api_token())
+    brokers = client.manager.get_brokers().items
+    with tempfile.NamedTemporaryFile(
```
geokala: This needs to get deleted at some point though, right? Otherwise we just end up with an ever increasing pile of certs?

PR #1812, opened 2019-05-25
CY-1259: SharedResource operation

tyacbovi:
SharedResource operation for running a remote workflow on it


cloudify_types/cloudify_types/shared_resource/execute_shared_resource_workflow.py
```diff
@@ -0,0 +1,91 @@
...
+                                     redirect_logs):
+    if not _verify_shared_resource_node():
+        raise NonRecoverableError('Tried to execute "{0}" workflow on a non '
+                                  'SharedResource node "{1}"'.format(
+                                    workflow_id, ctx.target.node.id))
+
+    ctx.logger.info("Setting up required input for executing workflow on"
+                    "a SharedResource.")
+    # Cloudify client setup
+    if client_config:
```
EarthmanT: Should we be getting the local client or the target node client (what if they are not the same)?

tyacbovi: it's a good idea to take the client definition from the target node (of course if not define set the local one)

PR #1818, opened 2019-05-29
CY-1259: adding SharedResource needed support

tyacbovi:
Which will be used for marking that deployment is valid


cloudify_types/cloudify_types/shared_resource/shared_resource.py
```diff
@@ -48,6 +48,16 @@ def __init__(self, operation_inputs):
         self.deployment = self.config.get('deployment', '')
         self.deployment_id = self.deployment.get('id', '')
 
+    def _mark_verified_shared_resource_node(self):
+        """
+        Used to mark SharedResource node that is valid after verification
+        that the deployment exists, which will be used be users like
+        the UI.
+        """
+        ctx.instance.runtime_properties['deployment'] = dict()
```
barucoh: What we discussed

PR #1824, opened 2019-06-02
Cy 1244 support upload plugin in cluster

adigrabow:


rest-service/manager_rest/rest/resources_v2/plugins.py
```diff
@@ -174,8 +176,11 @@ def get(self, plugin_id, **kwargs):
         """
         # Verify plugin exists.
         plugin = get_storage_manager().get(models.Plugin, plugin_id)
-
-        archive_name = plugin.archive_name
+        # While installing a plugin we add an `installing` prefix to the
+        # archive_name, since the archive is saved without this prefix we
+        # remove it before searching for the archive in the file system.
+        archive_name = (plugin.archive_name[11:] if INSTALLING_PREFIX
```
barucoh: Several places in the code, should use find and replace instead of [magic_number:] based on the `INSTALLING_PREFIX`

rest-service/manager_rest/rest/resources_v2/plugins.py
```diff
@@ -174,8 +176,11 @@ def get(self, plugin_id, **kwargs):
...
+                        in plugin.archive_name else plugin.archive_name)
```
barucoh: Should use `.starts_with` instead of in

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -63,10 +65,31 @@ def post(self, **kwargs):
...
+    def put(self, plugin_id, **kwargs):
+        """
+        Update the plugin's archive_name (remove the `installing-` prefix). '
+        This method is called from the PluginInstaller after the plugin
+        installation has ended.
+        """
+
+        sm = get_storage_manager()
+        plugin = sm.get(models.Plugin, plugin_id)
+        if INSTALLING_PREFIX in plugin.archive_name:
```
barucoh: Same as previous comment

rest-service/manager_rest/upload_manager.py
```diff
@@ -659,9 +660,26 @@ def _prepare_and_process_doc(self,
...
                                        version=new_plugin.package_version))
+            if self._is_plugin_installing(new_plugin, plugin):
+                raise manager_exceptions.ConflictError(
+                    'a plugin archive by the name of {archive_name} for '
+                    'package with name {package_name} and version {version} '
+                    'is currently being installed'.format(
+                        archive_name=new_plugin.archive_name,
+                        package_name=new_plugin.package_name,
+                        version=new_plugin.package_version))
+        new_plugin.archive_name = INSTALLING_PREFIX + new_plugin.archive_name
```
barucoh: Maybe use '{0}{1}'.format instead of `+` as it is more consistent to the rest of our code?
We don't usually use `+`

rest-service/manager_rest/upload_manager.py
```diff
@@ -659,9 +660,26 @@ def _prepare_and_process_doc(self,
...
         sm.put(new_plugin)
         return new_plugin, new_plugin.archive_name
 
+    @staticmethod
+    def _is_plugin_installing(new_plugin, plugin):
+        """
+        Plugins that are currently being installed have an `installing` prefix.
+        """
+        return (plugin.archive_name == INSTALLING_PREFIX +
```
barucoh: Same as previous comment

rest-service/manager_rest/upload_manager.py
```diff
@@ -659,9 +660,26 @@ def _prepare_and_process_doc(self,
...
+                        package_name=new_plugin.package_name,
+                        version=new_plugin.package_version))
+        new_plugin.archive_name = INSTALLING_PREFIX + new_plugin.archive_name
         sm.put(new_plugin)
         return new_plugin, new_plugin.archive_name
 
+    @staticmethod
+    def _is_plugin_installing(new_plugin, plugin):
+        """
+        Plugins that are currently being installed have an `installing` prefix.
```
barucoh: `installing-`

rest-service/manager_rest/constants.py
```diff
@@ -61,3 +61,4 @@
 MODELS_TO_PERMISSIONS = {'NodeInstance': 'node_instance'}
 FORBIDDEN_METHODS = ['POST', 'PATCH', 'PUT']
 SANITY_MODE_FILE_PATH = '/opt/manager/sanity_mode'
+INSTALLING_PREFIX = 'installing-'
```
tyacbovi: Maybe it's better to change it to INSTALLING_STATUS_PREFIX

rest-service/manager_rest/rest/resources_v2/plugins.py
tyacbovi: I think it's better to use here the partition function of string for getting the archive name

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -63,10 +65,31 @@ def post(self, **kwargs):
...
         Upload a plugin
         """
+
         visibility = rest_utils.get_visibility_parameter(
             optional=True,
             is_argument=True,
             valid_values=VisibilityState.STATES
         )
         with rest_utils.skip_nested_marshalling():
             return super(Plugins, self).post(visibility=visibility)
```
barucoh: Before this line there should be a `SQL insert` of the plugin with the new PREFIX to verify that the plugin is not being installed.
Verifying that there is not `installing-<plugin_name> is not sufficient since 2 machines can check that and get an OK for it, and immediately after that try to insert the `installing-<plugin_name>`, only one will succeed causing the other to fail which will fail the request.

rest-service/manager_rest/rest/resources_v2/plugins.py
tyacbovi: Maybe also move this to common, it can be used also for the removing of the prefix

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -63,10 +65,31 @@ def post(self, **kwargs):
...
+
+
+class PluginsId(resources_v2_1.PluginsId):
+
+    @rest_decorators.exceptions_handled
+    @authorize('provider_context_get')
+    @rest_decorators.marshal_with(models.Plugin)
+    def put(self, plugin_id, **kwargs):
+        """
+        Update the plugin's archive_name (remove the `installing-` prefix). '
```
tyacbovi: I will not mention the impl details, because the impl could be changed in the future and this doc could not be updated

rest-service/manager_rest/rest/resources_v3_1/plugins.py
tyacbovi: I would log if the 'else' because this should not happen and this is a source of a bug

rest-service/manager_rest/upload_manager.py
tyacbovi: using the .format approach is better memory wise

rest-service/manager_rest/upload_manager.py
tyacbovi: Also I will not mention the impl detail

rest-service/manager_rest/upload_manager.py
```diff
@@ -659,9 +660,26 @@ def _prepare_and_process_doc(self,
...
+                    'is currently being installed'.format(
+                        archive_name=new_plugin.archive_name,
+                        package_name=new_plugin.package_name,
+                        version=new_plugin.package_version))
+        new_plugin.archive_name = INSTALLING_PREFIX + new_plugin.archive_name
         sm.put(new_plugin)
         return new_plugin, new_plugin.archive_name
 
+    @staticmethod
+    def _is_plugin_installing(new_plugin, plugin):
```
tyacbovi: I will also put this in common, so all logic will be in the same place

rest-service/manager_rest/upload_manager.py
```diff
@@ -659,9 +660,26 @@ def _prepare_and_process_doc(self,
                     '{version}'.format(archive_name=new_plugin.archive_name,
                                        package_name=new_plugin.package_name,
                                        version=new_plugin.package_version))
+            if self._is_plugin_installing(new_plugin, plugin):
+                raise manager_exceptions.ConflictError(
+                    'a plugin archive by the name of {archive_name} for '
```
tyacbovi: I will improve the existing wording by
                   '"{archive_name}" plugin archive for'
                    ' "{package_name}"package and  {version} version '
                    'is already being installed'

rest-service/manager_rest/upload_manager.py
tehasdf: not always, in this case `+` uses less memory than `format` because `+` involves creating 3 strings: the 2 arguments, and the resulting output
and `format` involves creating 4 strings: the 2 arguments, the resulting output, and the `'{0}{1}'` format string :P

that doesn't really matter though of course, and str.format is great and fits here

amqp-postgres/dev-requirements.txt
```diff
@@ -1,2 +1,2 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==master
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-1244-support-upload-plugin-in-cluster#egg=cloudify-common==master
```
tehasdf: this file doesn't really matter and won't even really use that, however what does matter is mgmtworker/requirements.txt - if you want CI to build the mgmtworker rpm using that common branch, change it there

tyacbovi: You mean cloudify-manager/packaging/mgmtworker/requirements.txt, no?

tehasdf: yes ofc

PR #1839, opened 2019-06-13
Fix test_managers unittests

tehasdf:


rest-service/manager_rest/rest/resources_v3_1/manager.py
```diff
@@ -172,10 +172,11 @@ def post(self):
         )
         result = sm.put(new_manager)
         current_app.logger.info('Manager added successfully')
-        if manager.get('fs_sync_node_id'):
+        if add_manager and manager.get('fs_sync_node_id'):
             managers_list = get_storage_manager().list(models.Manager)
             add_manager(managers_list)
-        update_agents(sm)
+        if update_agents:
```
barucoh: Why is this separate from the previous condition?
Like in the remove one...

PR #1842, opened 2019-06-13
CY-1060 Add a plugins update API

davidginzbourg:
- Fixes the case where deployment update doesn't update  sourced
centrally deployed plugins

- Adds an `update_plugins` flag to pass to the update workflow. This flag
controls the plugin installation, if set to False - it won't do any
plugin un/installations.


resources/rest-service/cloudify/types/types.yaml
```diff
@@ -635,6 +635,13 @@ workflows:
         type: boolean
       node_instances_to_reinstall:
         default: []
+      central_plugins_to_install:
```
tyacbovi: Please set types of list

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +608,60 @@ def _execute_workflow(self,
         )
         return new_execution
 
+    def _extract_plugins_changes(self, dep_update):
+        """Extracts plugins that need to be installed and uninstalled.
```
tyacbovi: or uninstalled

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +608,60 @@ def _execute_workflow(self,
...
+        :param dep_update: a DeploymentUpdate object.
+        :return: plugins that need installation and uninstallation (a tuple).
+        """
+
+        def is_centrally_deployed(plugin):
+            return plugin[constants.PLUGIN_EXECUTOR_KEY] \
+                   == constants.CENTRAL_DEPLOYMENT_AGENT
+
+        def append_plugins_to_list(source_dict, key_list, target_list):
+            for k in key_list:
```
tyacbovi: use .extend to join lists

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +608,60 @@ def _execute_workflow(self,
...
+            new_plan[constants.DEPLOYMENT_PLUGINS_TO_INSTALL],
+            new_plan[constants.WORKFLOW_PLUGINS_TO_INSTALL],
+            filter_func=is_centrally_deployed)
+        # Convert to plugin_name->plugin dict
+        new_plugins = {p[constants.PLUGIN_NAME_KEY]: p
+                       for p in plugins_to_install_new}
+        old_plugins = {p[constants.PLUGIN_NAME_KEY]: p
+                       for p in plugins_to_install_old}
+        new_p_names = set(new_plugins)
+        old_p_names = set(old_plugins)
```
tyacbovi: may not merge this handling from 640-645 to the relevant extract_and_merge_plugins_to_uninstall

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +608,60 @@ def _execute_workflow(self,
...
+
+        central_plugins_to_install, central_plugins_to_uninstall = [], []
+        append_plugins_to_list(new_plugins,
+                               new_p_names.difference(old_p_names),
```
tyacbovi: .difference can be inside this func with better name

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +608,60 @@ def _execute_workflow(self,
...
+        plugins_to_install_new = extract_and_merge_plugins_to_uninstall(
+            new_plan[constants.DEPLOYMENT_PLUGINS_TO_INSTALL],
+            new_plan[constants.WORKFLOW_PLUGINS_TO_INSTALL],
+            filter_func=is_centrally_deployed)
+        # Convert to plugin_name->plugin dict
+        new_plugins = {p[constants.PLUGIN_NAME_KEY]: p
+                       for p in plugins_to_install_new}
+        old_plugins = {p[constants.PLUGIN_NAME_KEY]: p
+                       for p in plugins_to_install_old}
+        new_p_names = set(new_plugins)
```
tyacbovi: why is this needed? same data is saved twice

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +608,60 @@ def _execute_workflow(self,
...
+        deployment = self.sm.get(models.Deployment, dep_update.deployment_id)
+        old_plan = deployment.blueprint.plan
+        new_plan = dep_update.deployment_plan
+        # Plugins to install in the old plan
+        plugins_to_install_old = extract_and_merge_plugins_to_install(
+            old_plan[constants.DEPLOYMENT_PLUGINS_TO_INSTALL],
+            old_plan[constants.WORKFLOW_PLUGINS_TO_INSTALL],
+            filter_func=is_centrally_deployed)
+        # Plugins to install in the new plan
+        plugins_to_install_new = extract_and_merge_plugins_to_uninstall(
```
tyacbovi: same code for both cases, so why not add this to 'extract_and_merge_plugins_to_uninstall'

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+from manager_rest.resource_manager import get_resource_manager
+from manager_rest.plugin_update.constants import (STATES,
+                                                  ACTIVE_STATES)
+from manager_rest.manager_exceptions import (ConflictError,
+                                             PluginsUpdateError,
+                                             IllegalActionError)
+from manager_rest.constants import (FILE_SERVER_BLUEPRINTS_FOLDER,
+                                    HIDDEN_BLUEPRINT_SUFFIX)
+
+from dsl_parser.constants import (PLUGIN_NAME_KEY,
```
tyacbovi: should be before the manager_rest

rest-service/manager_rest/plugin_update/constants.py
```diff
@@ -0,0 +1,35 @@
...
+#  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  * See the License for the specific language governing permissions and
+#  * limitations under the License.
+
+from collections import namedtuple
+
+PluginsUpdatePhases = namedtuple('PluginUpdatePhases', ['INITIAL', 'FINAL'])
+PHASES = PluginsUpdatePhases(INITIAL='initiate', FINAL='finalize')
+
+PluginsUpdateStates = namedtuple('PluginsUpdateStates',
```
tyacbovi: What about preview?

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+            id=update_id,
+            created_at=utils.get_formatted_timestamp())
+        plugins_update.set_blueprint(blueprint)
+        return self.sm.put(plugins_update)
+
+    def initiate_plugins_update(self, blueprint_id, force=False):
+        """Creates a temporary blueprint and executes the plugins update
+        workflow.
+        """
+
```
tyacbovi: not needed space

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+    def validate_no_active_updates_per_blueprint(self,
+                                                 blueprint_id,
+                                                 force=False):
+        active_updates = self.list_plugins_updates(
+            filters={'blueprint_id': blueprint_id,
+                     'state': ACTIVE_STATES}).items
+        if not active_updates:
+            return
+
+        if not force:
```
tyacbovi: So force is the only way to update?

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+            raise ConflictError(
+                'There are plugins updates still active, update IDs: '
+                '{0}'.format(', '.join([u.id for u in active_updates])))
+
+        execs_to_updates = {u.execution_id: u for u in active_updates
+                            if u.execution_id}
```
tyacbovi: Can u.execution_id be none?

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+        # real active updates are those with an execution in a running status
+        real_active_updates = [
```
tyacbovi: Better to use running_updates then to add the comment

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+            execs_to_updates[ex.id]
+            for ex in self.sm.list(models.Execution,
```
tyacbovi: ex => execution

rest-service/manager_rest/constants.py
```diff
@@ -64,3 +64,4 @@
 }
 FORBIDDEN_METHODS = ['POST', 'PATCH', 'PUT']
 SANITY_MODE_FILE_PATH = '/opt/manager/sanity_mode'
+HIDDEN_BLUEPRINT_SUFFIX = '_temp'
```
tyacbovi: maybe chose something like _bluprint

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+            # the active updates aren't really active - either their
+            # executions were failed/cancelled, or the update failed at
+            # the finalizing stage.
+            # updating their states to failed and continuing.
+            for plugins_update in active_updates:
+                plugins_update.state = STATES.FAILED
+                self.sm.update(plugins_update)
+
+    @staticmethod
+    def _get_available_random_resource_id(prefix):
```
tyacbovi: Should be more simple, like prefix+uuid4 (cut this if needed by 256-len pre -len suf)+suffix

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+    def initiate_plugins_update(self, blueprint_id, force=False):
+        """Creates a temporary blueprint and executes the plugins update
+        workflow.
+        """
+
+        self.validate_no_active_updates_per_blueprint(blueprint_id, force)
+        blueprint = self.sm.get(models.Blueprint, blueprint_id)
+        temp_plan = self.get_reevaluated_plan(blueprint)
+
+        deployments_to_update = \
```
tyacbovi: use ( ) to avoid \ , don't do this always

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+    def _validate_plugins_update_state(plugin_update):
+        if plugin_update.state != STATES.EXECUTING_WORKFLOW:
+            raise IllegalActionError(
+                "Cannot finalize plugins update {0}, "
+                "it's not in the {1} state.".format(plugin_update.id,
+                                                    STATES.EXECUTING_WORKFLOW))
+
+    def _validate_execution_status(self, plugin_update):
+        execution = self.sm.get(models.Execution, plugin_update.execution_id)
+        if execution.status in ExecutionState.END_STATES \
```
tyacbovi: ( )

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+                and execution.status != ExecutionState.TERMINATED:
+            plugin_update.state = STATES.FAILED
+            self.sm.update(plugin_update)
+            exec_stat_txt = 'failed' if \
```
tyacbovi: use more words in the var name  :)

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+                execution.status == ExecutionState.FAILED else 'cancelled'
+            raise PluginsUpdateError(
+                'The execution of plugins update {0} {1}.'.format(
+                    plugin_update.id,
+                    exec_stat_txt))
+
+    def _get_deployments_to_update(self, blueprint_id):
+        return self.sm.list(models.Deployment,
```
tyacbovi: Not necessary to put in new func

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+                            filters={'blueprint_id': blueprint_id},
+                            sort={'id': 'asc'}).items
+
+    def get_plugins_update(self, plugins_update_id):
```
tyacbovi: Not necessary to put in new func

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+                             substr_filters=None):
+        return self.sm.list(models.PluginsUpdate,
+                            include=include,
+                            filters=filters,
+                            pagination=pagination,
+                            substr_filters=substr_filters,
+                            sort=sort)
+
+    def get_reevaluated_plan(self, blueprint):
+
```
tyacbovi: not needed space

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+def _did_plugins_to_install_change(temp_plan, plan):
+    temp_plugins = temp_plan[DEPLOYMENT_PLUGINS_TO_INSTALL] + \
+                   temp_plan[WORKFLOW_PLUGINS_TO_INSTALL]
+    plugins = plan[DEPLOYMENT_PLUGINS_TO_INSTALL] + plan[
+        WORKFLOW_PLUGINS_TO_INSTALL]
+    temp_plugins_dict = {p[PLUGIN_NAME_KEY]: p for p in temp_plugins}
+    plugins_dict = {p[PLUGIN_NAME_KEY]: p for p in plugins}
+    for p_name in temp_plugins_dict:
+        if temp_plugins_dict[p_name] != plugins_dict.get(p_name, None):
```
tyacbovi: do this in var loop assignment

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def _did_plugins_to_install_change(temp_plan, plan):
+    temp_plugins = temp_plan[DEPLOYMENT_PLUGINS_TO_INSTALL] + \
+                   temp_plan[WORKFLOW_PLUGINS_TO_INSTALL]
+    plugins = plan[DEPLOYMENT_PLUGINS_TO_INSTALL] + plan[
+        WORKFLOW_PLUGINS_TO_INSTALL]
+    temp_plugins_dict = {p[PLUGIN_NAME_KEY]: p for p in temp_plugins}
+    plugins_dict = {p[PLUGIN_NAME_KEY]: p for p in plugins}
+    for p_name in temp_plugins_dict:
```
tyacbovi: can be one liner

rest-service/manager_rest/rest/resources_v2/blueprints.py
```diff
@@ -54,6 +55,7 @@ def get(self, _include=None, filters=None, pagination=None, sort=None,
             '_get_all_results',
             request.args.get('_get_all_results', False)
         )
+        not_like_filters = {'id': '%{}'.format(HIDDEN_BLUEPRINT_SUFFIX)}
```
tyacbovi: needs a better name and comment

rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
```diff
@@ -160,26 +157,24 @@ def _get_and_validate_blueprint_and_inputs(deployment_id, request_json):
         return blueprint, inputs, reinstall_list
 
     @staticmethod
-    def _get_params_and_validate(deployment_id,
-                                 request_json,
-                                 preserve_old_behavior=False):
+    def _parse_args(deployment_id, request_json, from_post=False):
```
tyacbovi: 'from_post' not clear..

rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
```diff
@@ -244,3 +240,16 @@ def get(self,
                 substr_filters=search
             )
         return deployment_updates
+
+
+def _get_blueprint_abs_path(server_root, blueprint):
```
tyacbovi: very tied to plugins update, should be a more relevant name

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -70,3 +75,89 @@ def post(self, **kwargs):
...
+        Phases:
+            1. Creates a temporary blueprint and executes a deployment update
+            (will update only the plugins) for all the deployments of the
+            given blueprint.
+            2. Updates the original blueprint plan and deletes the temp one.
+
+        :param id: the blueprint ID to update it's deployments' plugins if
+        phase == 'initiate', otherwise (phase == 'finalize') the plugin update
+        ID.
+        :param phase: either 'initiate' or 'finalize'.
```
tyacbovi: It is only internal, right? so maybe mark it like that

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -70,3 +75,89 @@ def post(self, **kwargs):
...
+        """
+
+        def get_args():
```
tyacbovi: new func not needed

rest-service/manager_rest/test/endpoints/test_blueprints.py
```diff
@@ -449,3 +450,19 @@ def test_force_delete_used_blueprints_via_import(self):
                           first_blueprint_id)
 
         self.client.blueprints.delete(first_blueprint_id, True)
+
+    @attr(client_min_version=base_test.LATEST_API_VERSION,
```
tyacbovi: client_min_version -> 3.1

rest-service/manager_rest/test/endpoints/test_blueprints.py
```diff
@@ -449,3 +450,19 @@ def test_force_delete_used_blueprints_via_import(self):
...
+          client_max_version=base_test.LATEST_API_VERSION)
+    def test_not_listing_hidden_blueprints(self):
+        blueprint_file = 'blueprint_with_inputs.yaml'
+        blueprint_path = os.path.join(
+            self.get_blueprint_path('mock_blueprint'),
+            blueprint_file)
+        b0_id = 'b0' + HIDDEN_BLUEPRINT_SUFFIX + 'asdf'
+        b1_id = 'b1' + HIDDEN_BLUEPRINT_SUFFIX
```
tyacbovi: The user should not be able to upload a hidden blueprint, because it can not be managed. Maybe change it to a flag prop

rest-service/manager_rest/test/endpoints/test_deployment_updates.py
```diff
@@ -123,6 +124,89 @@ def test_remove_node_and_relationship(self):
...
+
+        self._deploy_base(deployment_id, 'setup_with_plugin_1.yaml',
+                          skip_plugins_validation=True)
+
+        self._update(deployment_id, 'setup_with_plugin_2.yaml')
+        dep_update = \
+            self.client.deployment_updates.list(deployment_id=deployment_id)[0]
+
+        execution = self.client.executions.get(dep_update.execution_id)
+        exec_params = execution.parameters
```
tyacbovi: Use the original var, the name is more readable

rest-service/manager_rest/test/endpoints/test_deployment_updates.py
```diff
@@ -123,6 +124,89 @@ def test_remove_node_and_relationship(self):
...
+        for changed_param in changed_params:
```
tyacbovi: There is an assert that asserts lists without order

rest-service/manager_rest/test/endpoints/test_deployment_updates.py
```diff
@@ -123,6 +124,89 @@ def test_remove_node_and_relationship(self):
             self.assertEquals(1 if param in changed_params else 0,
                               len(execution.parameters[param]))
 
+    def test_plugin_installation_updates_plugins_by_default(self):
+        def assert_all_not_in(plugin_names, plugins_set):
+            for plugin_name in plugin_names:
+                self.assertNotIn(plugin_name, plugins_set)
+
+        def assert_all_in(plugin_names, plugins_set):
```
tyacbovi: There is an assert that asserts lists without order

rest-service/manager_rest/test/resources/deployment_update/depup_step/setup_with_plugin_1.yaml
```diff
@@ -0,0 +1,33 @@
...
+  should_uninstall:
+    executor: central_deployment_agent
+    source: .
+
+  should_skip_install:
+    executor: central_deployment_agent
+    source: .
+    package_version: "1"
+
+  should_reinstall:
```
tyacbovi: What will happen if a tried to put here a plugin with version 2 and update with version 1? this needs also a test

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
+#########
+# Copyright (c) 2015 GigaSpaces Technologies Ltd. All rights reserved
```
tyacbovi: missed this :)

rest-service/manager_rest/test/endpoints/test_deployment_updates.py
```diff
@@ -123,6 +124,89 @@ def test_remove_node_and_relationship(self):
...
+        self.assertEqual(
+            "1",
+            central_plugins_to_uninstall['should_reinstall'][
+                constants.PLUGIN_PACKAGE_VERSION])
+        self.assertEqual(
+            "2",
+            central_plugins_to_install['should_reinstall'][
+                constants.PLUGIN_PACKAGE_VERSION])
+
+    def test_plugin_installation_update_plugins_is_false_in_execution(self):
```
tyacbovi: maybe test_no_plugin_update_in_execution

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+
+import unittest
+from mock import patch
+from functools import wraps
+
+from manager_rest.test import base_test
+from manager_rest.test.attribute import attr
+from manager_rest.test.base_test import BaseServerTestCase
+
+from manager_rest.plugin_update.constants import STATES
```
tyacbovi: small style comment, better to move manager rest to be the last

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+    def wrapper(*args, **kwargs):
+        with patch('manager_rest.plugin_update.manager'
+                   '._did_plugins_to_install_change') as _f:
+            _f.return_value = True
+            return f(*args, **kwargs)
+
+    return wrapper
+
+
+@attr(client_min_version=base_test.LATEST_API_VERSION,
```
tyacbovi: version->3.1

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+        self.wait_for_deployment_creation(self.client, 'd123')
+        plugins_update1 = self.client.plugins_update.update_plugins(
+            'hello_world')
+        plugins_update2 = self.client.plugins_update.update_plugins(
+            'hello_world', force=True)
+        plugins_update_set = set(p.id for p in self.client.plugins_update
+                                 .list().items)
+        self.assertEqual(2, len(plugins_update_set))
+        self.assertIn(plugins_update1.id, plugins_update_set)
+        self.assertIn(plugins_update2.id, plugins_update_set)
```
tyacbovi: maybe verify the force also

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+    def test_raises_with_nonexisting_blueprint(self):
+        with self.assertRaises(CloudifyClientError) as e:
+            self.client.plugins_update.update_plugins("non_existing")
+            self.assertEqual(404, e.exception.status_code)
+
+    @suppress_plugins_change_detection
+    def test_raises_blueprint_has_no_deployments(self):
+        self.put_file(*self.put_blueprint_args(blueprint_id='hello_world'))
+        with self.assertRaisesRegexp(CloudifyClientError,
+                                     "The blueprint hello_world has no "
```
tyacbovi: I think it will be more readable to put "" around the blu name

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+                             plugins_update.deployments_to_update)
+        execution = self.client.executions.get(plugins_update.execution_id)
+        self.assertDictEqual(
+            execution.parameters,
+            {'update_id': plugins_update.id,
+             'deployments_to_update': ['d1', 'd2'],
+             'temp_blueprint_id': plugins_update.temp_blueprint_id})
+
+    @suppress_plugins_change_detection
+    def test_plugins_manager_is_a_singleton(self):
```
tyacbovi: Not needed

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+        self.assertTrue(
+            get_plugins_updates_manager() == get_plugins_updates_manager())
+
+    @suppress_plugins_change_detection
```
tyacbovi: Actually, it seems you can put this on the entire class, you can turn the patch on and off in the class setup teardown

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+    def test_finalize_updates_original_blueprint_plan(self):
+        self.put_file(*self.put_blueprint_args(blueprint_id='hello_world'))
+        self.client.deployments.create('hello_world', 'd123')
+        self.wait_for_deployment_creation(self.client, 'd123')
+        plugins_update_id = self.client.plugins_update.update_plugins(
+            'hello_world').id
+        blueprint = self._sm.get(models.Blueprint, 'hello_world')
+        plan = deepcopy(blueprint.plan)
+        plan.inputs['dummy'] = 'hi'
```
tyacbovi: why not check that the plugins are different? if the user did a plugin update only that part will change not the inputs

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+        self.wait_for_deployment_creation(self.client, 'd123')
+        plugins_update = self.client.plugins_update.update_plugins(
+            'hello_world')
+        self.assertNotEqual(plugins_update.state, STATES.SUCCESSFUL)
+        plugins_update = self.client.plugins_update.finalize_plugins_update(
+            plugins_update.id)
+        self.assertEqual(plugins_update.state, STATES.SUCCESSFUL)
+
+    @suppress_plugins_change_detection
+    def test_finalize_deletes_temp_blueprint(self):
```
tyacbovi: what about failure in installing the plugins?

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,365 @@
...
+        plugins_update = self.client.plugins_update.update_plugins(
+            'hello_world')
+        self._sm.get(models.Blueprint, plugins_update.temp_blueprint_id)
+        plugins_update = self.client.plugins_update.finalize_plugins_update(
+            plugins_update.id)
+        with self.assertRaises(NotFoundError):
+            self._sm.get(models.Blueprint, plugins_update.temp_blueprint_id)
+
+
+random_resource_id = PluginsUpdateManager._get_available_random_resource_id
```
tyacbovi: why not import them and use "as" in the import ?

tests/integration_tests/tests/agent_tests/test_plugin_update.py
```diff
@@ -17,63 +17,258 @@
...
         self._execute_workflows()
-        self._assert_on_values('2.0')
+        self._assert_on_values(self.versions[1])
 
-    def _perform_update(self):
-        self.client.deployment_updates.update_with_existing_blueprint(
-            deployment_id=self.setup_deployment_id,
-            blueprint_id=self.mod_name
+    @setup_for_sourced_plugins
+    def test_sourced_plugin_doesnt_update(self):
```
tyacbovi: test_sourced_plugin_doesnt_update -> test_sourced_plugin_doesnt_update_with_same_version

tests/integration_tests/tests/test_cases.py
```diff
@@ -615,11 +620,17 @@ def upload_mock_plugin(self, plugin_name, plugin_path=None):
...
 
+        # execution = self.client.executions.get(execution_id)
+        # self.wait_for_execution_to_end(execution)
         self._wait_for_execution_by_wf_name('install_plugin')
 
     def _wait_for_execution_by_wf_name(self, wf_name):
-        install_plugin_ex = [ex for ex in self.client.executions.list(
-            include_system_workflows=True) if ex.workflow_id == wf_name][0]
+        install_plugin_ex = [
+            ex for ex in self.client.executions.list(
```
tyacbovi: execution

tests/integration_tests_plugins/dockercompute/operations.py
```diff
@@ -127,6 +127,7 @@ def _init_script_agent_setup(container_id, install_agent_script):
 
 def _remote_agent_setup(container_id):
     _wait_for_ssh_setup(container_id)
+    _docker_exec(container_id, 'mkdir -p /root/.ssh/')
```
tyacbovi: why is this needed?

tests/integration_tests_plugins/testmockoperations/tasks.py
```diff
@@ -20,11 +20,10 @@
 import shutil
 
 import cloudify.utils
-from cloudify.manager import get_rest_client
-from cloudify.manager import get_node_instance_ip
 from cloudify.decorators import operation
-from cloudify.exceptions import NonRecoverableError
 from cloudify.exceptions import RecoverableError
+from cloudify.exceptions import NonRecoverableError
```
tyacbovi: you can combine imports here

tests/integration_tests/tests/agent_tests/test_plugin_update.py
```diff
@@ -17,63 +17,258 @@
...
                 with open(fpath) as f:
                     s = f.read()
-                s = s.replace('1.0', '2.0')
+                s = s.replace(v1, v2)
                 with open(fpath, 'w') as f:
                     f.write(s)
 
-    def _assert_on_values(self, version):
+    def _assert_cda_values(self, version):
         # Calling like this because "cda" op/wf would be written on the manager
```
tyacbovi: maybe add here the cda is in the comment

tests/integration_tests/tests/agent_tests/test_plugin_update.py
```diff
@@ -17,63 +17,258 @@
...
+        self.wait_for_execution_to_end(execution)
+
+    def _get_work_dir_for_plugin(self):
+        return os.path.join(self.workdir, self.plugin_name)
+
+    def _get_source_dir_for_plugin(self):
+        plugin_dir = '{0}'.format(self.plugin_name)
+        return resource(os.path.join('plugins', plugin_dir))
+
+    def _copy_plugin_files_to_work_dir(self):
```
tyacbovi: is there a test that the temp blueprint is really deleted?

workflows/cloudify_system_workflows/plugins.py
```diff
@@ -32,3 +35,43 @@ def _operate_on_plugin(ctx, plugin, action):
...
+def update(ctx, update_id, temp_blueprint_id, deployments_to_update, **_):
+    """Execute deployment update for all the given deployments_to_update.
+
+    :param update_id: plugins update ID.
+    :param temp_blueprint_id: temporary blueprint ID that should be used for
+    this workflow only.
+    :param deployments_to_update: deployments to perform the update on, using
+    the temp blueprint ID provided.
+    """
+    err_msg_format = 'Deployment update has failed with execution ID: {}.'
```
tyacbovi: I prepare it as a function, because you need this var and format it everywhere

workflows/cloudify_system_workflows/plugins.py
```diff
@@ -32,3 +35,43 @@ def _operate_on_plugin(ctx, plugin, action):
...
+                       '{}...'.format(dep))
+        execution_id = client.deployment_updates \
+            .update_with_existing_blueprint(deployment_id=dep,
+                                            blueprint_id=temp_blueprint_id,
+                                            skip_install=True,
+                                            skip_uninstall=True,
+                                            skip_reinstall=True) \
+            .execution_id
+
+        wait_for(client.executions.get,
```
tyacbovi: point to think, if we want to give the user the ability to specify the timeout to wait

workflows/cloudify_system_workflows/plugins.py
```diff
@@ -32,3 +35,43 @@ def _operate_on_plugin(ctx, plugin, action):
...
+                 'status',
+                 lambda x: x in ExecutionState.END_STATES,
+                 RuntimeError,
+                 err_msg_format.format(execution_id))
+        execution_status = client.executions.get(execution_id).status
+        if execution_status in (ExecutionState.FAILED,
+                                ExecutionState.CANCELLED):
+            raise RuntimeError("Deployment update of deployment {0} with "
+                               "execution ID {1} failed, stopped this "
+                               "plugins update.".format(dep, execution_id))
```
tyacbovi: maybe add the plugin update id here, so the user could search for it

workflows/cloudify_system_workflows/tests/snapshots/tests_utils.py
```diff
@@ -1,79 +0,0 @@
...
-import subprocess
-import unittest
-import tempfile
-
-from pytest import mark
-
-from cloudify_system_workflows.snapshots.utils import make_zip64_archive
-
-
-class MakeZip64Test(unittest.TestCase):
```
tyacbovi: Why deleted this?

tests/integration_tests/tests/test_cases.py
```diff
@@ -615,11 +620,17 @@ def upload_mock_plugin(self, plugin_name, plugin_path=None):
         with utils.zip_files([wagon_path, yaml_path]) as zip_path:
             self.client.plugins.upload(zip_path)
 
+        # execution = self.client.executions.get(execution_id)
```
tyacbovi: Can be removed

tests/integration_tests/tests/agent_tests/test_plugin_update.py
```diff
@@ -17,63 +17,258 @@
...
+        self._execute_workflows()
+        self._assert_on_values(self.versions[0])
+
+        self._perform_update()
+
+        # Execute mod (V 2.0) workflows
+        self._execute_workflows()
+        self._assert_on_values(self.versions[1])
 
+    @setup_for_sourced_plugins
```
tyacbovi: This patch is in use in all tests or not affecting them, no?

workflows/tox.ini
```diff
@@ -6,6 +6,7 @@ envlist=py27
 install_command = pip install -U {opts} {packages}
 deps =
     -rdev-requirements.txt
+    -rtest-requirements.txt
     pytest
```
tyacbovi: please move pytest and pytest-cov to the test-req

workflows/cloudify_system_workflows/tests/test_plugins.py
```diff
@@ -0,0 +1,123 @@
...
+import unittest
+from functools import wraps
+
+from mock import patch, MagicMock, PropertyMock, call
+
+from cloudify import decorators
+from cloudify.models_states import ExecutionState
+
+
+def mock_workflow_decorator(f=None, *_, **__):
```
tyacbovi: style issue, move this func and the decorators.workflow after the imports

workflows/cloudify_system_workflows/tests/test_plugins.py
```diff
@@ -0,0 +1,123 @@
...
+
+    return wrapper
+
+
+decorators.workflow = mock_workflow_decorator
+
+from cloudify_system_workflows.plugins import update as update_func  # noqa
+
+
+class TestPluginsUpdate(unittest.TestCase):
```
tyacbovi: maybe also test that you can not run this if another execution is already running in the system (not have to be also plugins update)

rest-service/manager_rest/deployment_update/manager.py
davidginzbourg: I don't think this belongs to the `extract_and_merge` function. This is a comparison logic between two lists of plugins, and `extract_and_merge` only deals with merging workflow + deployment plugins. Your thoughts?

rest-service/manager_rest/deployment_update/manager.py
davidginzbourg: Moved `.difference` inside and renamed function to `extend_list_from_dict(source_dict,                                source_keys, keys_to_filter_out, target_list)`

rest-service/manager_rest/deployment_update/manager.py
davidginzbourg: Your'e right. Thought it was saving some coding, but it only saved one line. Removed.

rest-service/manager_rest/deployment_update/manager.py
davidginzbourg: I think you didn't consider that the `extract_and_merge` functions are using [here](https://github.com/cloudify-cosmo/cloudify-manager/blob/CY-1060-plugin-update/workflows/cloudify_system_workflows/deployment_environment.py#L40) and [here](https://github.com/cloudify-cosmo/cloudify-manager/blob/CY-1060-plugin-update/workflows/cloudify_system_workflows/deployment_environment.py#L81)

rest-service/manager_rest/plugin_update/constants.py
davidginzbourg: Asked Ido, waiting for his answer.

rest-service/manager_rest/plugin_update/manager.py
davidginzbourg: Yeah, the execution ID is added only at the [last stage](https://github.com/cloudify-cosmo/cloudify-manager/blob/CY-1060-plugin-update/rest-service/manager_rest/plugin_update/manager.py#L151)

rest-service/manager_rest/plugin_update/manager.py
davidginzbourg: Thanks, that's really a better name.

rest-service/manager_rest/deployment_update/manager.py
tyacbovi: You don't have to put that there, I meant that there is a repetition that u can encapsulate

rest-service/manager_rest/plugin_update/manager.py
davidginzbourg: Not sure if what I did now is better. I can't use a generator, I need to save the list.

rest-service/manager_rest/plugin_update/manager.py
davidginzbourg: Resolving after the discussion we had

rest-service/manager_rest/plugin_update/manager.py
davidginzbourg: It has special filter and sort keys, this might introduce errors otherwise. That's why I think it's best to keep it in a func.

rest-service/manager_rest/plugin_update/manager.py
davidginzbourg: Don't know why I constructed the temp dict. I removed the redundant dict. I don't think it can be shortened further.

rest-service/manager_rest/rest/resources_v2/blueprints.py
davidginzbourg: Irrelevant now. Arg removed.

rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
davidginzbourg: Changed to `using_post_request`

rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
davidginzbourg: Changed to `_check_if_temp_of_plugins_update_and_get_abs_path`, is it better?

rest-service/manager_rest/test/endpoints/test_deployment_updates.py
davidginzbourg: There's `assertItemsEqual` but that checks for equality and I need to assert it's a sublist of the other.

rest-service/manager_rest/test/endpoints/test_blueprints.py
davidginzbourg: Changed to v2 for now. Waiting for Ido to answer.

rest-service/manager_rest/test/endpoints/test_deployment_updates.py
davidginzbourg: Same as #discussion_r293823307

rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
tyacbovi: _get_plugin_update_blueprint_abs_path

rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
tyacbovi: Why not "is_plugin_update" although I don't like this specification

davidginzbourg: This isn't related to plugins update, just whether the deployment has been performed via a PUT request or a POST

rest-service/manager_rest/test/resources/deployment_update/depup_step/setup_with_plugin_1.yaml
davidginzbourg: Added, FYI the changes are detected by comparing the two dicts.

rest-service/manager_rest/plugin_update/constants.py
davidginzbourg: Not in the scope for 5.0.

rest-service/manager_rest/test/endpoints/test_blueprints.py
davidginzbourg: No need for a new class, using `client_min_version=2`

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: It doesn't save the `force` flag and the mechanism itself is tested in other tests.

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: I surrounded it with ' '

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: This just checks that the plan has been updated, it doesn't really matter. For the sake of future implementations I made a change in the plugins key.

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: The first one is a class method so it's not possible, regarding the second - I don't know what I was thinking. Good catch

tests/integration_tests/tests/agent_tests/test_plugin_update.py
davidginzbourg: Then it might sound as if the blueprint had the same version or something, but here it explicitely didn't upate the plugins although the node data shows it has plugin v2.

tests/integration_tests_plugins/dockercompute/operations.py
davidginzbourg: I suspect that in the new centos image, there's no `.ssh` dir. Over at Jenkins the image has it. Maybe over at Jenkins we have a custom centos image but the one you download from the internet doesn't have the dir.

tests/integration_tests/tests/agent_tests/test_plugin_update.py
davidginzbourg: [Yep](https://github.com/cloudify-cosmo/cloudify-manager/blob/77dfdd9179c6be859320fe461aa6874757c3b94e/rest-service/manager_rest/test/endpoints/test_plugins_updates.py#L242)

workflows/cloudify_system_workflows/plugins.py
davidginzbourg: Made it an inner function, removed the var.

workflows/cloudify_system_workflows/plugins.py
davidginzbourg: Wrote it down in the JIRA

workflows/cloudify_system_workflows/plugins.py
davidginzbourg: You're absolutely right. Added.

workflows/cloudify_system_workflows/tests/snapshots/tests_utils.py
davidginzbourg: We've discussed this at the office. It was marked as skip by you a year ago and never ran since.

workflows/cloudify_system_workflows/tests/test_plugins.py
davidginzbourg: Wouldn't that be the responsibility of the scheduler?

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -70,3 +75,85 @@ def post(self, **kwargs):
...
+            (will update only the plugins) for all the deployments of the
+            given blueprint.
+            2. Updates the original blueprint plan and deletes the temp one.
+
+        :param id: the blueprint ID to update it's deployments' plugins if
+        phase == 'initiate', otherwise (phase == 'finalize') the plugin update
+        ID.
+        :param phase: either 'initiate' or 'finalize' (internal).
+        """
+        if phase == PHASES.INITIAL:
```
tyacbovi: if "phase" is binary, why not use a bool flag?

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -70,3 +75,85 @@ def post(self, **kwargs):
...
+class PluginsUpdate(SecuredResource):
+
+    @rest_decorators.exceptions_handled
+    @authorize('plugins_update_create')
+    @rest_decorators.marshal_with(models.PluginsUpdate)
+    def post(self, id, phase):
+        """
+        Supports two stages of a plugin update.
+        Phases:
+            1. Creates a temporary blueprint and executes a deployment update
```
tyacbovi: add here "1.  INITIAL - " and also for the second

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -70,3 +75,85 @@ def post(self, **kwargs):
...
+            (will update only the plugins) for all the deployments of the
+            given blueprint.
+            2. Updates the original blueprint plan and deletes the temp one.
+
+        :param id: the blueprint ID to update it's deployments' plugins if
+        phase == 'initiate', otherwise (phase == 'finalize') the plugin update
```
tyacbovi: initiate is different from INITIAL, is it a mistake? I think it's better to not use the const values because they can change and the docstring could be not changed with

rest-service/manager_rest/test/endpoints/test_blueprints.py
```diff
@@ -449,3 +451,23 @@ def test_force_delete_used_blueprints_via_import(self):
...
+    @attr(client_min_version=2,
+          client_max_version=base_test.LATEST_API_VERSION)
+    def test_not_listing_hidden_blueprints(self):
+        blueprint_file = 'blueprint_with_inputs.yaml'
+        blueprint_path = os.path.join(
+            self.get_blueprint_path('mock_blueprint'),
+            blueprint_file)
+        b0_id = 'b0'
+        b1_id = 'b1'
+        self.client.blueprints.upload(blueprint_path, b0_id)
```
tyacbovi: there is a way to send the HTTP req without the client, so I think it's a better way to go because it will test the is_hidden assignment

rest-service/manager_rest/test/endpoints/test_deployment_updates.py
```diff
@@ -123,6 +124,99 @@ def test_remove_node_and_relationship(self):
...
+                constants.PLUGIN_PACKAGE_VERSION])
+        self.assertEqual(
+            "1",
+            central_plugins_to_install['should_reinstall_v2'][
+                constants.PLUGIN_PACKAGE_VERSION])
+
+    def test_plugin_updates_in_execution_is_disabled(self):
+        deployment_id = 'dep'
+        changed_params = ['update_plugins',
+                          # Plugins to install and uninstall are still
```
tyacbovi: Can this be avoided? not to this check if not required

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,352 @@
...
+
+from manager_rest.plugin_update.constants import STATES
+from manager_rest.manager_exceptions import NotFoundError
+from manager_rest.storage import get_storage_manager, models
+from manager_rest.plugin_update.manager import \
+    (PluginsUpdateManager,
+     _did_plugins_to_install_change as plugins_change_detector)
+
+
+def _plugins_update_test_setup(self):
```
tyacbovi: Create a base test class and inherit from that, by this u will save a couple of lines and the code will look better

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,352 @@
...
+
+    self.plugin_change_patcher = patch(
+        'manager_rest.plugin_update.manager'
+        '._did_plugins_to_install_change')
+    self.plugin_change_patched = True
+    self.addCleanup(plugin_change_cleanup)
+    self.plugin_change_patcher.start().return_value = True
+
+
+@attr(client_min_version=3.1,
```
tyacbovi: Not client_min_version=2?

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,352 @@
...
+                                 .list().items)
+        self.assertEqual(2, len(plugins_update_set))
+        self.assertIn(plugins_update1.id, plugins_update_set)
+        self.assertIn(plugins_update2.id, plugins_update_set)
+
+    def test_returns_empty_list_with_no_plugins_updates(self):
+        self.assertEqual(0, len(self.client.plugins_update.list().items))
+
+
+@attr(client_min_version=base_test.LATEST_API_VERSION,
```
tyacbovi: Also here

rest-service/manager_rest/test/endpoints/test_deployment_updates.py
tyacbovi: Are you sure about the less restrict 'in'? can it be more strict like startswith?

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -0,0 +1,352 @@
...
+        self.client.deployments.create('hello_world', 'd1')
+        self.wait_for_deployment_creation(self.client, 'd1')
+        plugins_update = self.client.plugins_update.update_plugins(
+            'hello_world')
+        response_plugins_update = self.client.plugins_update.get(
+            plugins_update.id)
+        self.assertEqual(plugins_update.id, response_plugins_update.id)
+
+
+@attr(client_min_version=base_test.LATEST_API_VERSION,
```
tyacbovi: Also here

tests/integration_tests/tests/agent_tests/test_plugin_update.py
```diff
@@ -17,65 +17,260 @@
...
 
 
 class TestPluginUpdate(AgentTestWithPlugins):
+    versions = ['1.0', '2.0']
+    dsl_resources_path = resource(os.path.join('dsl', 'agent_tests'))
+    blueprint_name_prefix = 'plugin_update_'
+    setup_deployment_ids = None
+
+    def tearDown(self):
+        deps_to_remove = []
```
tyacbovi: Why is this needed? the dep id are uuid

tests/integration_tests/tests/agent_tests/test_plugin_update.py
tyacbovi: but there is more then one reason for this to happen

tests/integration_tests_plugins/dockercompute/operations.py
tyacbovi: There was no update to the image in the recent time

workflows/cloudify_system_workflows/tests/test_plugins.py
tyacbovi: no

workflows/cloudify_system_workflows/tests/test_plugins.py
```diff
@@ -0,0 +1,124 @@
...
+
+        def returns_failed_execution_only_if_is_stop_at(execution_id):
+            if execution_id == stop_at:
+                return PropertyMock(status=test_params['curr_exec_status'])
+            return PropertyMock(status=ExecutionState.TERMINATED)
+
+        def _assert_update_func_raises():
+            with self.assertRaisesRegexp(
+                    RuntimeError,
+                    "Deployment update of deployment {0} with execution ID {0}"
```
tyacbovi: there is two {0}

workflows/cloudify_system_workflows/tests/test_plugins.py
```diff
@@ -0,0 +1,124 @@
...
+    return wrapper
+
+
+decorators.workflow = mock_workflow_decorator
+
+
+class TestPluginsUpdate(unittest.TestCase):
+    @patch('cloudify_system_workflows.plugins.get_rest_client')
+    @patch('cloudify_system_workflows.plugins.wait_for')
+    def test_stops_updating_when_one_update_fails(self,
```
tyacbovi: The test is very hard to read and understand what it does

workflows/cloudify_system_workflows/tests/test_plugins.py
```diff
@@ -0,0 +1,124 @@
...
+        test_params['curr_exec_status'] = ExecutionState.CANCELLED
+        _assert_update_func_raises()
+
+    @patch('cloudify_system_workflows.plugins.get_rest_client')
+    @patch('cloudify_system_workflows.plugins.wait_for')
+    def test_doesnt_stop_updating(self, _, mock_get_rest_client):
+        def update_with_existing_blueprint_mock(deployment_id, *_, **__):
+            return PropertyMock(execution_id=deployment_id)
+
+        mock_rest_client = MagicMock()
```
tyacbovi: Maybe some of the setup could be shared across the tests

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -492,11 +521,16 @@ def _execute_update_workflow(self,
...
             # Whether or not execute install/uninstall/reinstall,
-            # order of execution, and behavior in failure while uninstalling
+            # order of execution, behavior in failure while uninstalling, and
+            # whether or not to update the plugins.
             'skip_install': skip_install,
             'skip_uninstall': skip_uninstall,
             'ignore_failure': ignore_failure,
             'install_first': install_first,
+            'update_plugins': update_plugins,
+
```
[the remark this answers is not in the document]
davidginzbourg: Added a comment above the section below it. I think it's easier to read this this way.

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +607,59 @@ def _execute_workflow(self,
...
+                with_repetition=is_old_plan)
+
+        def is_centrally_deployed(plugin):
+            return plugin[constants.PLUGIN_EXECUTOR_KEY] \
+                   == constants.CENTRAL_DEPLOYMENT_AGENT
+
+        def extend_list_from_dict(source_dict, filter_out_dict, target_list):
+            target_list.extend(
+                source_dict[k]
+                for k in source_dict if k not in filter_out_dict)
```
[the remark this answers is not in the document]
davidginzbourg: This is the strictest, it compares the `src key` and tries to find an exact match in the `filter out dict`

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +607,59 @@ def _execute_workflow(self,
...
+
+        central_plugins_to_install, central_plugins_to_uninstall = [], []
+        extend_list_from_dict(source_dict=new_plugins,
+                              filter_out_dict=old_plugins,
+                              target_list=central_plugins_to_install)
+        extend_list_from_dict(source_dict=old_plugins,
+                              filter_out_dict=new_plugins,
+                              target_list=central_plugins_to_uninstall)
+        # Deal with the intersection between the old and new plugins
+        intersection = (k for k in new_plugins if k in old_plugins)
```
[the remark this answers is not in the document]
davidginzbourg: This is the same as it used to be with the set (time complexity wise), but without wasting space.

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -573,6 +607,59 @@ def _execute_workflow(self,
...
+        for plugin_name in intersection:
+            old_plugin = old_plugins[plugin_name]
+            new_plugin = new_plugins[plugin_name]
+            if new_plugin == old_plugin:
```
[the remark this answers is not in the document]
davidginzbourg: Yes

rest-service/manager_rest/plugin_update/manager.py
```diff
@@ -0,0 +1,260 @@
...
+        return plan
+
+
+def _did_plugins_to_install_change(temp_plan, plan):
+    temp_plugins = temp_plan[DEPLOYMENT_PLUGINS_TO_INSTALL] + \
+                   temp_plan[WORKFLOW_PLUGINS_TO_INSTALL]
+    plugins = plan[DEPLOYMENT_PLUGINS_TO_INSTALL] + plan[
+        WORKFLOW_PLUGINS_TO_INSTALL]
+    plugins_dict = {p[PLUGIN_NAME_KEY]: p for p in plugins}
+    return any(plugin for plugin in temp_plugins
```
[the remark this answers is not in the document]
davidginzbourg: This creates a list and I just need to tell whether that condition is true for at least one of the items. I think `any(..)` is what I need

rest-service/manager_rest/rest/resources_v3_1/plugins.py
davidginzbourg: More human friendly IMO, `False/True` won't have any true meaning and I think it'll be confusing especially in case of tracking rest calls

rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
tyacbovi: A little bit strange

rest-service/manager_rest/rest/resources_v3_1/plugins.py
davidginzbourg: The phase is the initial phase but you ask the server to initiate

davidginzbourg: I changed the docstring to contain the constant name

rest-service/manager_rest/deployment_update/manager.py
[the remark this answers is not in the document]
tyacbovi: Remember to take a note of that in the docs

rest-service/manager_rest/plugin_update/manager.py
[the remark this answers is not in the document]
tyacbovi: yes, but u create a couple of data structures that could be not created, not critical.
Also plugins_dict is not a clear name :)

rest-service/manager_rest/rest/resources_v3_1/plugins.py
tyacbovi: It's kind of error-prone (string comp)

workflows/cloudify_system_workflows/tests/test_plugins.py
davidginzbourg: Yeah, the deployment ID and execution ID are the same

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: Now it does, saved and tested

tests/integration_tests/tests/agent_tests/test_plugin_update.py
davidginzbourg: I think this is a mistake as well. Will verify in a bit

tests/integration_tests/tests/agent_tests/test_plugin_update.py
davidginzbourg: I don't quite understand, in here I pass `update_plugins=False` and expect the plugins don't update. Why not use that as a name?

rest-service/manager_rest/deployment_update/manager.py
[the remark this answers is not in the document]
tyacbovi: Sounds good

tests/integration_tests/tests/agent_tests/test_plugin_update.py
tyacbovi: because that connection it's clear, try to view it if a person that wants to understand what the feature does without reading the code or a doc

rest-service/manager_rest/plugin_update/manager.py
tyacbovi: what is the connection to '( )'?

tests/integration_tests/tests/agent_tests/test_plugin_update.py
davidginzbourg: `test_sourced_plugin_remains_with_same_version`?

rest-service/manager_rest/plugin_update/manager.py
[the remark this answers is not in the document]
tyacbovi: current_plugins?

tests/integration_tests/tests/agent_tests/test_plugin_update.py
davidginzbourg: It's a mistake, removed.

rest-service/manager_rest/plugin_update/manager.py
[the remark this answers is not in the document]
davidginzbourg: I still think name_to_plugin_dict explains it better

tests/integration_tests_plugins/dockercompute/operations.py
davidginzbourg: Then maybe some other program changed its behavior. The dir doesn't exist at that point, it fails without this command. Also this looks like a common command, saw another procedure over at `docl`. Your thoughts?

rest-service/manager_rest/plugin_update/manager.py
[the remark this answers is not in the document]
tyacbovi: name_to_plugin, deal? :)

tests/integration_tests_plugins/dockercompute/operations.py
tyacbovi: it's not critical, but strange

rest-service/manager_rest/plugins_update/manager.py
```diff
@@ -0,0 +1,245 @@
...
+        active_updates = self.list_plugins_updates(
+            filters={'blueprint_id': blueprint_id,
+                     'state': ACTIVE_STATES}).items
+        if not active_updates:
+            return
+
+        if not force:
+            raise ConflictError(
+                'There are plugins updates still active, update IDs: '
+                '{0}'.format(', '.join([u.id for u in active_updates])))
```
tehasdf: remember that `str.join` takes iterables, and a generator expression is fine, so this can just be 
```python
', '.join(u.id for u in active_updates)
```
(without the `[ ]`)

davidginzbourg: Woops

workflows/cloudify_system_workflows/tests/test_plugins.py
davidginzbourg: Waiting for Ido so see if I need to add that wait for running execs to finish.

workflows/cloudify_system_workflows/tests/test_plugins.py
```diff
@@ -0,0 +1,121 @@
...
+        return lambda _f, *_, **__: mock_workflow_decorator(_f, *_, **__)
+
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        return f(*args, **kwargs)
+
+    return wrapper
+
+
+decorators.workflow = mock_workflow_decorator
```
tyacbovi: Needs a comment here

davidginzbourg: It's redundant, removed. Thanks.

PR #1843, opened 2019-06-14
Move managers and brokers endpoints out to premium

tehasdf:
The parts of managers and brokers that are premium-only should
be implemented in cloudify-premium. There's no reason they would
be implemented here.

Additionally, this means that tests that are defined in
cloudify-premium, check the code that is defined there as well.

There should be an accompanying cloudify-premium PR to this.


tehasdf:
also see cloudify-cosmo/cloudify-premium#554

PR #1867, opened 2019-07-07
Cy 1401 improve snapshot dump agents

glukhman:


workflows/cloudify_system_workflows/snapshots/agents.py
```diff
@@ -51,16 +51,19 @@ def restore(self, tempdir, version):
...
-                _include=['id'],
-                _get_all_results=True
-            )
-            for deployment in tenant_deployments:
-                result[tenant_name][deployment.id] = \
-                    self._get_deployment_result(tenant_client, deployment.id)
+        client = get_rest_client()
+        agents = client.agents.list(all_tenants=True)
+        node_instances = list(client.node_instances.list(
+            id=[item['id'] for item in agents.items],
```
inbala: I think it will be nicer to get the ids first :
```
agent_ids = [agent.id for agent in agents.items]
```
And then use it in the list

workflows/cloudify_system_workflows/snapshots/agents.py
```diff
@@ -51,16 +51,19 @@ def restore(self, tempdir, version):
...
+            _all_tenants=True))
+        for inst in node_instances:
```
inbala: I prefer `instance` over `inst`

workflows/cloudify_system_workflows/snapshots/agents.py
```diff
@@ -51,16 +51,19 @@ def restore(self, tempdir, version):
...
+            result.setdefault(inst['tenant_name'], {})
```
inbala: Maybe all the content of the `for` should be in another function `add_instance_to_result`

PR #1868, opened 2019-07-08
CY-1441: Allowing to pass integer secrets in Component 

tyacbovi:


tests/integration_tests/tests/agentless_tests/component/test_component_basics.py
```diff
@@ -76,6 +76,150 @@ def test_component_creation_with_secrets_and_plugins(self):
...
+        blueprint_path = self.make_yaml_file(main_blueprint)
+        self.deploy_application(blueprint_path, deployment_id=deployment_id)
+        deployment = self.client.deployments.get('component')
+        self.assertEquals(deployment.inputs['integer'], 1)
+        self.assertEquals(deployment.inputs['list'], [1, 2])
+        self.assertEquals(deployment.inputs['dict'], {'a': 1})
+        self.assertEquals(deployment.inputs['string'], 'a')
+        self.assertEquals(deployment.inputs['float'], 1.5)
+        self.assertEquals(deployment.inputs['boolean'], False)
+        self.assertEquals(deployment.inputs['regex'], '^.$')
```
adigrabow: The assertions are the same, maybe use a method to do them?

tehasdf: `assertEquals` is deprecated since 2.6, use `assertEqual`

cloudify_types/cloudify_types/component/component.py
```diff
@@ -232,7 +232,7 @@ def _set_secrets(self):
         for secret_name in self.secrets:
             self._http_client_wrapper('secrets', 'create', {
                 'key': secret_name,
-                'value': self.secrets[secret_name],
+                'value': str(self.secrets[secret_name]),
```
tehasdf: this is most likely gonna break non-ascii secrets in py2
for a 2+3-compat version you can use:
```python
u'{0}'.format(self.secrets[secret_name])
```

tests/integration_tests/tests/agentless_tests/component/test_component_basics.py
tyacbovi: because it's a one liner it will not really change something

PR #1871, opened 2019-07-08
CY-1444: handling creation of existing secrets

tyacbovi:
When trying to upload already existing secrets with the same value, we should pass else fail

### DIFF
--- a/cloudify_types/cloudify_types/component/component.py
+++ b/cloudify_types/cloudify_types/component/component.py
@@ -21,7 +21,6 @@ from cloudify.exceptions import NonRecoverableError
 from cloudify_rest_client.client import CloudifyClient
 from cloudify_rest_client.exceptions import CloudifyClientError
 
-
 from .constants import (
     EXECUTIONS_TIMEOUT,
     POLLING_INTERVAL,
--- a/cloudify_types/cloudify_types/component/component.py
+++ b/cloudify_types/cloudify_types/component/component.py
@@ -36,8 +36,7 @@ from .utils import (
     deployment_id_exists,
     update_runtime_properties,
     get_local_path,
-    zip_files,
-    get_secret_by_name
+    zip_files
 )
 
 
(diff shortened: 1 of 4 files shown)


cloudify_types/cloudify_types/component/component.py
```diff
@@ -230,6 +231,20 @@ def _set_secrets(self):
...
+            existing_secret_value = get_secret_by_name(self.client,
+                                                       secret_name)
+            if existing_secret_value:
+                if self.secrets[secret_name] == existing_secret_value:
+                    ctx.logger.info('Not creating secret "{0}" because it is '
+                                    'already exists with same value...'.format(
+                                     secret_name))
+                    continue
+                else:
+                    raise NonRecoverableError('Tried to upload a secret "{0}"'
```
geokala: Wouldn't it be better to not check the value at all, but instead just check to see if a secret by the same name exists and say "Not modifying pre-existing secret <name>..."?

geokala: Or even to have a flag that could be provided (e.g. --force to override them or another flag to fail if any already exist)

cloudify_types/cloudify_types/component/component.py
```diff
@@ -193,6 +194,8 @@ def _upload_plugins(self):
             ctx.instance.runtime_properties['plugins'] = []
 
         for plugin in self.plugins.values():
+            ctx.logger.info('Verifying that plugin is not already uploaded..')
```
geokala: Is it verifying that?

cloudify_types/cloudify_types/component/component.py
```diff
@@ -230,6 +231,16 @@ def _set_secrets(self):
             return
 
         for secret_name in self.secrets:
+            existing_secret_value = get_secret_by_name(self.client,
```
geokala: Can this perhaps be made a little more efficient by retrieving a list of all secret names before entering this loop and then just checking if secret_name in all_secrets?

PR #1881, opened 2019-07-12
CY-1442: Component - not uploading an existing plugin

tyacbovi:


cloudify_types/cloudify_types/component/tests/test_deployment.py
```diff
@@ -191,28 +191,69 @@ def test_upload_plugins(self):
...
             ):
-                zip_files = mock.Mock(return_value="_zip")
                 with mock.patch(
-                    'cloudify_types.component.component.zip_files',
-                    zip_files
-                ):
-                    # dist of plugins
+                    'cloudify_types.component.component.should_upload_plugin'
+                ) as should_upload_plugin:
+                    should_upload_plugin.return_value = True
```
tehasdf: btw you can write:
```python
with mock.patch('cloudify_types.component.component.should_upload_plugin', return_value=True) as should_upload_plugin:
```

cloudify_types/cloudify_types/component/tests/test_deployment.py
```diff
@@ -191,28 +191,69 @@ def test_upload_plugins(self):
                 'cloudify_types.component.component.get_local_path',
                 get_local_path
             ):
-                zip_files = mock.Mock(return_value="_zip")
                 with mock.patch(
-                    'cloudify_types.component.component.zip_files',
-                    zip_files
-                ):
-                    # dist of plugins
+                    'cloudify_types.component.component.should_upload_plugin'
```
tehasdf: odd that the new tests mock out the new function, no? :P

cloudify_types/cloudify_types/component/tests/test_deployment.py
```diff
@@ -163,7 +162,7 @@ def test_create_deployment_failed(self):
                     timeout=MOCK_TIMEOUT)
 
                 self.assertIn('No execution Found for component'
-                              ' \"test\" deployment', error.message)
+                              ' "test" deployment', error.message)
```
tehasdf: if already changing this line, might as well make it `str(error)` as well instead of `error.message` (.message is deprecated since 2.6 and removed in 3)

cloudify_types/cloudify_types/component/component.py
```diff
@@ -185,14 +186,16 @@ def upload_blueprint(self):
...
 
-        if 'plugins' not in ctx.instance.runtime_properties:
-            ctx.instance.runtime_properties['plugins'] = []
+        ctx.instance.runtime_properties['plugins'] = []
+        existing_plugins = self._http_client_wrapper(
+                    'plugins', 'list')
 
-        for plugin in self.plugins.values():
-            ctx.logger.info('Creating plugin zip archive..')
+        for plugin_name, plugin in self.plugins.iteritems():
```
tehasdf: might as well use `.items()` and not `.iteritems()` for future py3-compat

cloudify_types/cloudify_types/component/utils.py
```diff
@@ -133,3 +134,21 @@ def blueprint_id_exists(client, blueprint_id):
 def deployment_id_exists(client, deployment_id):
     deployment = get_deployment_by_id(client, deployment_id)
     return True if deployment else False
+
+
+def should_upload_plugin(plugin_yaml_path, existing_plugins):
+    with open(plugin_yaml_path, 'r') as plugin_yaml_file:
+        plugin_yaml = yaml.load(plugin_yaml_file)
```
tehasdf: unfortunate that we `yaml.load` here, do we have a util to use instead of yaml directly?

maybe use safe_load? remember that `yaml.load` can run (python) code

cloudify_types/cloudify_types/component/utils.py
```diff
@@ -133,3 +134,21 @@ def blueprint_id_exists(client, blueprint_id):
...
+        plugins = plugin_yaml.get('plugins')
+        for _, plugin_info in plugins.iteritems():
```
tehasdf: ditto iteritems

cloudify_types/cloudify_types/component/utils.py
```diff
@@ -133,3 +134,21 @@ def blueprint_id_exists(client, blueprint_id):
...
+            package_name = plugin_info.get('package_name')
+            package_version = str(plugin_info.get('package_version'))
```
tehasdf: why str()? Combined with get, this might end up being the string "None", is that what we want?

cloudify_types/cloudify_types/component/utils.py
```diff
@@ -133,3 +134,21 @@ def blueprint_id_exists(client, blueprint_id):
 def deployment_id_exists(client, deployment_id):
     deployment = get_deployment_by_id(client, deployment_id)
     return True if deployment else False
+
+
+def should_upload_plugin(plugin_yaml_path, existing_plugins):
+    with open(plugin_yaml_path, 'r') as plugin_yaml_file:
+        plugin_yaml = yaml.load(plugin_yaml_file)
+        plugins = plugin_yaml.get('plugins')
```
tehasdf: can reduce indentation (exit `with` block) here

cloudify_types/cloudify_types/component/utils.py
```diff
@@ -133,3 +134,21 @@ def blueprint_id_exists(client, blueprint_id):
...
+        for _, plugin_info in plugins.iteritems():
+            package_name = plugin_info.get('package_name')
+            package_version = str(plugin_info.get('package_version'))
+            distribution = plugin_info.get('distribution')
+            identical_plugins = [
```
tehasdf: we don't use `identical_plugins`, so there's no reason to have a list containing `plugin.id`
I think I'd write this like:
```python
for plugin in existing_plugins:
   if (plugin.package_name == package_name and 
       plugin.package_version == package_version and
       plugin.distribution == distribution):
        return False
return True
```
or using `any()`

tests/integration_tests/tests/agentless_tests/component/test_component_basics.py
```diff
@@ -76,6 +77,85 @@ def test_component_creation_with_secrets_and_plugins(self):
...
+            'cloudify_script_plugin-1.2-py27-none-any.wgn',
+            'cloudify_script_plugin/1_2/plugin.yaml'
+        )
+
+        blueprint_path = self.make_yaml_file(main_blueprint)
+        self.deploy_application(blueprint_path, deployment_id=deployment_id)
+        plugins_list = self.client.plugins.list()
+        self.assertTrue(plugins_list[0]['package_name'],
+                        self.TEST_PACKAGE_NAME)
+        self.assertEqual(len(self.client.plugins.list()), 1)
```
tehasdf: we can just assert on `plugins_list`, but I think you did it like this for symmetry w/ the post-undeploy check?

tehasdf: when asserting len AND asserting something on `plugins_list[0]`, please change the ordering to assert on the length FIRST, so that if the list is empty, we get a test failure, not an IndexError in the `[0]`

cloudify_types/cloudify_types/component/utils.py
tyacbovi: so maybe we need to switch to safe load in all other places?

cloudify_types/cloudify_types/component/utils.py
tyacbovi: Because of that, the parser can convert the version into a float by accident

PR #1882, opened 2019-07-12
amqp-postgres: insert timestamp at utc

tehasdf:
This reverts #1504

We only use the TIMESTAMP WITHOUT TIME ZONE type, and those are not
affected by the timezone setting. However, the now() function is.

This means that when we used now() without the `AT TIME ZONE`
specifier, we inserted timestamps at local timezone in amqp-postgres,
however we did insert timestamps at utc in the restservice side, due
to #1493

The comment was wrong when it said that the "DB is changing": the
DB timezone setting doesn't affect neither inserts or selects from
TIMESTAMP WITHOUT TIME ZONE columns, it only affects the output of
the now() function (and it also affects TIMESTAMP WITH TIME ZONE
columns, but we don't use those).

### DIFF
--- a/tests/integration_tests/tests/agentless_tests/test_events.py
+++ b/tests/integration_tests/tests/agentless_tests/test_events.py
@@ -21,7 +21,7 @@ from requests.exceptions import ConnectionError
 from integration_tests import AgentlessTestCase
 from integration_tests.framework.postgresql import run_query
 from integration_tests.tests.utils import get_resource as resource
-from integration_tests.framework import utils
+from integration_tests.framework import docl, utils
 from cloudify_rest_client.exceptions import CloudifyClientError
 
 from manager_rest.flask_utils import get_postgres_conf
@@ -231,6 +231,10 @@ class EventsAlternativeTimezoneTest(EventsTest):
             "ALTER USER {} SET TIME ZONE '{}'"
             .format(postgres_conf.username, cls.TIMEZONE)
         )
+        # restart all users of the db so that they get a new session which
+        # uses the just-set timezone
+        docl.execute(
+            "systemctl restart cloudify-amqp-postgres cloudify-restservice")
 
     def setUp(self):
         """Update postgres timezone and create a deployment."""
--- a/amqp-postgres/amqp_postgres/postgres_publisher.py
+++ b/amqp-postgres/amqp_postgres/postgres_publisher.py
@@ -50,12 +50,8 @@ EVENT_INSERT_QUERY = """
     VALUES %s
 """
 
-# we use now() and not 'AT UTC'
-# to align with insertions from storage manager
-# that uses UTC in the insert, but DB is changing
-# according to timezone in postgres.conf
 EVENT_VALUES_TEMPLATE = """(
-        now(),
+        now() at time zone 'utc',
         CAST (%(timestamp)s AS TIMESTAMP),
         %(execution_id)s,
         %(tenant_id)s,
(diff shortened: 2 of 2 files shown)


tehasdf:
This is tested by the `EventsAlternativeTimezoneTest` integration test.
Specifically, it was easy to trigger a failure by running:
```
pytest integration_tests/tests/agentless_tests/test_events.py::EventsAlternativeTimezoneTest::test_snapshots_events integration_tests/tests/agentless_tests/test_events.py::EventsAlternativeTimezoneTest::test_timestamp_in_utc```
Because only the snapshot test restarted amqp-postgres (and rest), so only after that test, the `SET TIME ZONE` in `setUpClass` took effect at all.

**Comment by tehasdf** [comment]
Here's a detailed explanation of how timezones work with the TIMESTAMP WITHOUT TIME ZONE column, and the timezone setting:

Prepare the test: with the timezone set to UTC, insert some timestamps
```
cloudify_db=# create table t(name text, x timestamp without time zone);
CREATE TABLE
cloudify_db=# show time zone;
 TimeZone 
----------
 UTC
(1 row)

cloudify_db=# 
cloudify_db=# insert into t(name, x) values('utc: now()', now());
INSERT 0 1
cloudify_db=# insert into t(name, x) values('utc: now() at tz utc', now() at time zone 'utc');
INSERT 0 1
cloudify_db=# insert into t(name, x) values('utc: literal', '2019-07-12T10:59:53.189Z');
INSERT 0 1
```

Then we see that in all cases we SELECT the same values that we inserted, and
they are at UTC (ie. the time they report is what time in UTC it was when the
commands were run)
```
cloudify_db=# select * from t;
         name         |             x              
----------------------+----------------------------
 utc: now()           | 2019-07-12 11:06:13.668688
 utc: now() at tz utc | 2019-07-12 11:06:13.671271
 utc: literal         | 2019-07-12 10:59:53.189
(3 rows)
```


Now, change the timezone setting, and insert some timestamps again:
```
cloudify_db=# set time zone 'Asia/Jerusalem';
SET
cloudify_db=# show time zone;
    TimeZone    
----------------
 Asia/Jerusalem
(1 row)

cloudify_db=# 
cloudify_db=# insert into t(name, x) values('jerusalem: now()', now());
INSERT 0 1
cloudify_db=# insert into t(name, x) values('jerusalem: now() at tz utc', now() at time zone 'utc');
INSERT 0 1
cloudify_db=# insert into t(name, x) values('jerusalem: literal', '2019-07-12T10:59:53.189Z');
INSERT 0 1
```

Show all the inserted timestamps again. We see that the timestamps inserted earlier
didn't change, so SELECTs from this column don't depend on the timezone setting.
We also see that literal inserts and inserts of `now() at time zone 'utc'` are
the same as before, so INSERTs into this column don't depend on the timezone
setting either.
Only thing that does depend on the timezone setting is a bare `now()`
```
cloudify_db=# select * from t;
            name            |             x              
----------------------------+----------------------------
 utc: now()                 | 2019-07-12 11:06:13.668688
 utc: now() at tz utc       | 2019-07-12 11:06:13.671271
 utc: literal               | 2019-07-12 10:59:53.189
 jerusalem: now()           | 2019-07-12 14:06:13.67592
 jerusalem: now() at tz utc | 2019-07-12 11:06:13.67776
 jerusalem: literal         | 2019-07-12 10:59:53.189
(6 rows)
```

File: tests/integration_tests/tests/agentless_tests/test_events.py [hunk 1]
```diff
@@ -231,6 +231,10 @@ def setUpClass(cls):
             "ALTER USER {} SET TIME ZONE '{}'"
             .format(postgres_conf.username, cls.TIMEZONE)
         )
+        # restart all users of the db so that they get a new session which
+        # uses the just-set timezone
+        docl.execute(
```
tehasdf: without this restart, the above `ALTER USER` did absolutely nothing!!! because all users of the db just stayed on their previous sessions

PR #1887, opened 2019-07-22
Add restart workflow integration tests

tehasdf:
Tests that check that restart workflow runs both stop and start,
and also that it's resumable in both of these phases (the workflow
itself - the operations might well not be, but in the test, they are)

### DIFF
--- a/tests/integration_tests/tests/agentless_tests/test_events.py
+++ b/tests/integration_tests/tests/agentless_tests/test_events.py
@@ -16,10 +16,10 @@
 import uuid
 import time
 from datetime import datetime, timedelta
+from integration_tests.framework import docl
 from integration_tests import AgentlessTestCase
 from integration_tests.framework.postgresql import run_query
 from integration_tests.tests.utils import get_resource as resource
-from integration_tests.framework import docl
 
 from manager_rest.flask_utils import get_postgres_conf
 
--- a/tests/integration_tests/tests/agentless_tests/test_events.py
+++ b/tests/integration_tests/tests/agentless_tests/test_events.py
@@ -15,14 +15,11 @@
 
 import uuid
 import time
-import json
 from datetime import datetime, timedelta
-from requests.exceptions import ConnectionError
 from integration_tests import AgentlessTestCase
 from integration_tests.framework.postgresql import run_query
 from integration_tests.tests.utils import get_resource as resource
-from integration_tests.framework import docl, utils
-from cloudify_rest_client.exceptions import CloudifyClientError
+from integration_tests.framework import docl
 
 from manager_rest.flask_utils import get_postgres_conf
 
@@ -151,8 +148,7 @@ class EventsTest(AgentlessTestCase):
         # Make sure 'snapshots create' events appear
         snapshot_id = 's{0}'.format(uuid.uuid4())
         execution = self.client.snapshots.create(snapshot_id, False, False)
-        self._wait_for_events_to_update_in_DB(
-            execution, CREATE_SNAPSHOT_SUCCESS_MSG)
+        self.wait_for_event(execution, CREATE_SNAPSHOT_SUCCESS_MSG)
 
         # Make sure 'snapshots restore' events appear
         self.undeploy_application(
@@ -163,8 +159,7 @@ class EventsTest(AgentlessTestCase):
         # requests to avoid the deadlock described in CY-1455
         time.sleep(10)
 
-        self._wait_for_events_to_update_in_DB(
-            execution, RESTORE_SNAPSHOT_SUCCESS_MSG)
+        self.wait_for_event(execution, RESTORE_SNAPSHOT_SUCCESS_MSG)
 
     def _events_list(self, **kwargs):
         if 'deployment_id' not in kwargs:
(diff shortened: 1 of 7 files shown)


tests/integration_tests/tests/agentless_tests/test_workflow.py
```diff
@@ -57,6 +57,27 @@ def test_execute_operation(self):
...
+        blueprint_id = self.id()
+        deployment, _ = self.deploy_application(
+            dsl_path,
+            blueprint_id=blueprint_id,
+            timeout_seconds=15
+        )
+        execution = self.execute_workflow('restart', deployment.id)
+        events = self.client.events.list(
+            execution_id=execution.id, include_logs=True)
+        self.assertTrue(
```
tyacbovi: maybe you should add ordering assert, that the stopping happened before the starting

tests/integration_tests/tests/agentless_tests/test_workflow.py
```diff
@@ -57,6 +57,27 @@ def test_execute_operation(self):
...
+            any('stopping machine' in event['message'] for event in events))
+        self.assertTrue(
+            any('starting machine' in event['message'] for event in events))
+        data = self.get_plugin_data(
+            plugin_name='cloudmock',
+            deployment_id=deployment.id
+        )
+        self.assertEqual(list(data['machines'].values())[0], 'running')
```
tyacbovi: What does this test?

tests/integration_tests/tests/agentless_tests/test_workflow.py
```diff
@@ -57,6 +57,27 @@ def test_execute_operation(self):
         outputs = self.client.deployments.outputs.get(deployment.id).outputs
         self.assertEquals(outputs['ip_address'], '')
 
+    def test_restart(self):
```
tyacbovi: maybe test_restart => test_restart_executions_flow or something that will describe more what ur testing for

tests/integration_tests/tests/agentless_tests/test_resume.py
```diff
@@ -329,3 +335,44 @@ def test_force_resume_cancelled_nonresumable(self):
                         .runtime_properties['resumed'])
         self.assertTrue(self.client.node_instances.get(instance2.id)
                         .runtime_properties['marked'])
+
+    def test_restart(self):
```
tyacbovi: test_resume_failed_restart

tests/integration_tests/tests/agentless_tests/test_resume.py
```diff
@@ -329,3 +335,44 @@ def test_force_resume_cancelled_nonresumable(self):
...
+        """Test that the restart builtin workflow can be resumed.
+
+        We'll test resuming at both the "stop" stage and the "start" stage.
```
tyacbovi: I prefer for tests debugging to separate this test to two

tests/integration_tests/tests/agentless_tests/test_resume.py
```diff
@@ -329,3 +335,44 @@ def test_force_resume_cancelled_nonresumable(self):
...
+            deployment_id=dep.id)
+
+        # wait for the 'stop' operation to begin, and cancel
+        self._wait_for_log(execution, message=self.wait_message + ' - STOP')
+        self.client.executions.cancel(execution.id)
+        self.wait_for_execution_to_end(execution)
+
+        # now resume...
+        self.client.executions.resume(execution.id)
+        self._create_target_file('/tmp/wait_stop')
```
tyacbovi: maybe change the name to create_execution_lock or something that will describe the point of this line and related line in the rest of the test.

tests/integration_tests/tests/agentless_tests/test_workflow.py
tehasdf: yeah, nothing which isn't already tested by checking the logs

tests/integration_tests/tests/agentless_tests/test_resume.py
tehasdf: nothing is failed here... and "resume" is unneeded because it's already in the class name

tehasdf: oh i guess `test_restart_resume` is fine, makes for easier `pytest -k`

tests/integration_tests/tests/agentless_tests/test_resume.py
tehasdf: i dont want to do that because that makes the test much weaker. Currently it's the only test that resumes 2 separate graphs in a single execution; if it was separate then it'd be 2 tests that resume just one graph, and we already have tests that do that

tests/integration_tests/tests/agentless_tests/test_workflow.py
tehasdf: "test_restart_workflow" sounds good (instead of "test_restart"), that way it's easy to understand that it's about the workflow, and not about restarting anything (eg services, agents)

"flow" I think adds more questions than it answers

also added a docstring

tests/integration_tests/tests/agentless_tests/test_resume.py
tehasdf: but it's not a lock, and I don't want to add additional naming/abstraction here, I think working on this level is good, makes it easy to understand how does it work.

Added a docstring to the method describing why is it needed

tyacbovi: I don't like the added handling of '/tmp/wait_stop' file, it requires the reader to get into the test framework to really understand why this line is connected

tehasdf: changed it to be runtime_properties instead of files, and a list of "unlocked operations" in runtime props

tests/integration_tests/tests/agentless_tests/test_workflow.py
```diff
@@ -57,6 +59,29 @@ def test_execute_operation(self):
...
+        dsl_path = get_resource('dsl/basic.yaml')
+        blueprint_id = self.id()
+        deployment, _ = self.deploy_application(
+            dsl_path,
+            blueprint_id=blueprint_id,
+            timeout_seconds=15
+        )
+        execution = self.execute_workflow('restart', deployment.id)
+        # event storing is async, allow some time for them to be stored
+        time.sleep(2)
```
tyacbovi: this can be replaced by _wait_for_events_to_update_in_DB

tests/integration_tests/tests/agentless_tests/test_resume.py
```diff
@@ -91,10 +93,38 @@ def _stop_mgmtworker(self):
         self.logger.info('Stopping mgmtworker')
         self.execute_on_manager('systemctl stop cloudify-mgmtworker')
 
-    def _create_target_file(self):
-        self.execute_on_manager('touch {0}'.format(self.target_file))
-        self.addCleanup(self.execute_on_manager,
-                        'rm -rf {0}'.format(self.target_file))
+    def _unlock_operation(self, operation_name, node_ids=None, client=None):
```
tyacbovi: Cool :)

tests/integration_tests/tests/agentless_tests/test_events.py
```diff
@@ -15,14 +15,11 @@
...
 import time
-import json
 from datetime import datetime, timedelta
-from requests.exceptions import ConnectionError
 from integration_tests import AgentlessTestCase
 from integration_tests.framework.postgresql import run_query
 from integration_tests.tests.utils import get_resource as resource
-from integration_tests.framework import docl, utils
-from cloudify_rest_client.exceptions import CloudifyClientError
+from integration_tests.framework import docl
```
tyacbovi: very small style - push this line up for length order

PR #1899, opened 2019-08-05
CY-1559 hide credentials in logs

YoniItzhak:


workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -336,8 +337,13 @@ def _restore_dump(self, dump_file, db_name, table=None):
         run_shell(command)
 
     @staticmethod
-    def _append_dump(dump_file, query):
-        ctx.logger.debug('Adding to end of dump: {0}'.format(query))
+    def _append_dump(dump_file, query, without_password=True):
+        if without_password:
+            ctx.logger.debug('Adding to end of dump: {0}'.format(query))
+        else:
+            ctx.logger.debug("Adding to end of dump: UPDATE users SET "
```
barucoh: You have duplicate code here, and [here](https://github.com/cloudify-cosmo/cloudify-manager/blob/master/workflows/cloudify_system_workflows/snapshots/postgres.py#L180)
I would update the `_get_admin_user_update_query` to receive `username` and `password` as `None` and in case not None put the strings received instead of getting the credentials from `_get_admin_credentials`.
That way also the string will stay consistent in case it changes in the future.

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -512,3 +518,13 @@ def dump_license_to_file(self, tmp_dir):
     def restore_license_from_dump(self, tmp_dir):
         dump_file = os.path.join(tmp_dir, LICENSE_DUMP_FILE)
         self._restore_dump(dump_file, self._db_name, table='licenses')
+
+    @staticmethod
+    def _print_postgres_config(config):
+        config_password = config.postgresql_password
+        config_username = config.postgresql_username
```
barucoh: can be changed to:
`postgres_password, postgres_username = config.postgresql_password, config.postgresql_username`

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -512,3 +518,13 @@ def dump_license_to_file(self, tmp_dir):
...
+        config.postgresql_password = '********'
+        config.postgresql_username = '********'
```
barucoh: Can be changed to:
`config.postgresql_password = config.postgresql_username = '********'`

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -512,3 +518,13 @@ def dump_license_to_file(self, tmp_dir):
...
+        ctx.logger.debug('Init Postgres config: {0}'.format(config))
+        config.postgresql_password = config_password
+        config.postgresql_username = config_username
```
barucoh: Same as above

workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -34,20 +34,25 @@
 SCHEMA_SCRIPT = '/opt/manager/resources/cloudify/migrations/schema.py'
 
 
-class DictToAttributes(object):
-    def __init__(self, dic):
-        self._dict = dic
-
-    def __getattr__(self, name):
-        return self._dict[name]
+class DictToAttributes(dict):
```
barucoh: Since you are not doing a deep copy anymore, I think this is not needed?

workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -34,20 +34,25 @@
...
+            raise TypeError('expected dict')
+
+    def __getattr__(self, key):
+        return super(DictToAttributes, self).__getitem__(key)
+
+    def __setattr__(self, key, value):
+        super(DictToAttributes, self).__setitem__(key, value)
 
     def __str__(self):
-        try:
```
barucoh: Same

workflows/cloudify_system_workflows/snapshots/utils.py
YoniItzhak: The class DictToAttributes, as I understand it, is created in order to allow the action `config.<key>` instead of `config[key]`. 
What do you mean by not needed?

workflows/cloudify_system_workflows/snapshots/utils.py
YoniItzhak: __str__ is used to print the config dictionary in the function '_print_postgres_config(config)' under postgres.py.
Otherwise it would print the dictionary keys and values as unicode (i.e. {u'key': u'value'}).
Should we drop it??

workflows/cloudify_system_workflows/snapshots/postgres.py
YoniItzhak: I'm not quite sure I understood what you meant. When `append_dump` receives a query that contains credentials it should print the query with the credentials hidden by asterisk but append the real query to the dump file, how this change helps us with it? 
In any case, your point is right, we should make the function more generic. I'll work on it.

PR #1910, opened 2019-08-18
CYBL-888: set debug message's level to debug, and actually print the execution ID

isaac-s:


cloudify_types/cloudify_types/component/polling.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
```
tyacbovi: We actually order the import by length

cloudify_types/cloudify_types/component/polling.py
```diff
@@ -104,21 +97,27 @@ def redirect_logs(client, execution_id, instance_ctx=None):
                 event.get('message', "")
             )
             message = message.encode('utf-8')
+            level = event.get('level')
```
tyacbovi: maybe export this handling to some util file

cloudify_types/cloudify_types/component/polling.py
```diff
@@ -175,7 +174,7 @@ def is_deployment_execution_at_state(client,
 
     if not execution_id:
         raise NonRecoverableError(
-            'Execution id was not found for "{0}" deployment.'.format(
```
tyacbovi: I thought this form is more correct, like "I am going to empire state building"

cloudify_types/cloudify_types/component/polling.py
```diff
@@ -184,20 +183,19 @@ def is_deployment_execution_at_state(client,
...
-        'Execution "{0}" of component "{1}" state is {2}'.format(execution_id,
-                                                                 dep_id,
-                                                                 execution))
+        'Execution "%s" of component "%s" state is %s',
+        execution_id, dep_id, execution)
 
     if log_redirect:
         ctx.logger.debug(
-            'Execution info with log_redirect is {0}'.format(execution))
+            'Execution info with log_redirect is %s', execution)
```
tyacbovi: Why format will not work here?

tehasdf: format does work but in general this form is more correct and I wish we were using it everywhere... I'm not sure whether it's better for new changes to keep consistent with existing code, or use the better form.

The difference is that, note that this isn't actually doing the formatting right here (the `execution` is a separate argument to `ctx.logger.debug`). This way, the logging mechanism only formats the value into the string if the log will actually be emitted. If the logging level is set so that debug messages aren't shown, string formatting (which is potentially an expensive operation) won't be done this way

cloudify_types/cloudify_types/component/polling.py
```diff
@@ -104,21 +97,27 @@ def redirect_logs(client, execution_id, instance_ctx=None):
...
             message = message.encode('utf-8')
+            level = event.get('level')
 
-            ctx.logger.info(
-                'Message {0} for Event {1} for execution_id {1}'.format(
-                    message, event))
+            # If the event dict had a 'level' key, then the value is
+            # a string. In that case, convert it to uppercase and get
+            # the matching Python logging constant.
+            if isinstance(level, basestring):
```
tehasdf: plz avoid basestring checks, and isinstance checks in general. It seems that in this case just a `if level` would work just as well, or perhaps `if level and not isinstance(level, int)` (if it can be int, but the comment says that it'll be a string)

isaac-s: Done.

PR #1911, opened 2019-08-18
CY-1601: deleting not existing Component deployment

tyacbovi:
- In the install process, the install of Component deployment could fail and not be created. So when a user will try to clean this
up by uninstalling the entire deployment with that Component that should not fail and just pass (with warning msg).

### DIFF
--- a/cloudify_types/cloudify_types/component/component.py
+++ b/cloudify_types/cloudify_types/component/component.py
@@ -376,20 +376,28 @@ class Component(object):
 
         ctx.logger.info('Delete component\'s "{0}" deployment'
                         .format(self.deployment_id))
-        self._http_client_wrapper('deployments',
-                                  'delete',
-                                  dict(deployment_id=self.deployment_id))
 
-        ctx.logger.info("Waiting for component's deployment delete.")
-        poll_result = poll_with_timeout(
-            lambda: deployment_id_exists(self.client, self.deployment_id),
-            timeout=self.timeout,
-            expected_result=False)
+        poll_result = True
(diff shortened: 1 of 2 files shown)


cloudify_types/cloudify_types/component/component.py
```diff
@@ -376,20 +376,28 @@ def delete_deployment(self):
...
+                                      dict(deployment_id=self.deployment_id))
 
-        ctx.logger.info("Little wait internal cleanup services.")
+            ctx.logger.info("Waiting for component's deployment delete.")
+            poll_result = poll_with_timeout(
+                lambda: deployment_id_exists(self.client, self.deployment_id),
+                timeout=self.timeout,
+                expected_result=False)
+
+        ctx.logger.debug("Little wait internal cleanup services.")
```
barucoh: I would say:
"Internal services cleanup, please wait..."

cloudify_types/cloudify_types/component/component.py
```diff
@@ -376,20 +376,28 @@ def delete_deployment(self):
...
         time.sleep(POLLING_INTERVAL)
-        ctx.logger.info("Wait for stop all system workflows.")
 
+        ctx.logger.info("Wait for stop all system workflows.")
```
barucoh: "Waiting for all system workflows to stop/finish..."
In any case, I would change this, and the former log to be printed with every poll if it's possible with the retry number?
Just better UX instead of just keeping the user waiting with this message without seeing any actual progress...

PR #1913, opened 2019-08-23
CY-1539: keeping blueprint_id in executions list

tyacbovi:
- Adding blueprint_id to Execution table

- Creating the migration script and adding the field
- Adding blueprint_id to relevant execution creation

### DIFF
--- a/rest-service/manager_rest/deployment_update/manager.py
+++ b/rest-service/manager_rest/deployment_update/manager.py
@@ -590,10 +590,9 @@ class DeploymentUpdateManager(object):
                 execution_parameters),
             is_system_workflow=False
         )
-        if deployment:
-            new_execution.set_deployment(deployment,
-                                         deployment_update.new_blueprint_id)
-            deployment_update.execution = new_execution
+        new_execution.set_deployment(deployment,
+                                     deployment_update.new_blueprint_id)
+        deployment_update.execution = new_execution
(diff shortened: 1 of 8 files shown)


rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -591,7 +591,8 @@ def _execute_workflow(self,
             is_system_workflow=False
         )
         if deployment:
```
davidginzbourg: Pretty sure this if-clause is not necessary, `deployment` is being used as if it isn't `None` in previous lines.

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -279,7 +279,7 @@ def deployment(cls):
         return one_to_many_relationship(cls, Deployment, cls._deployment_fk)
 
     deployment_id = association_proxy('deployment', 'id')
-    blueprint_id = association_proxy('deployment', 'blueprint_id')
+    blueprint_id = db.Column(db.Text, nullable=True)
```
tehasdf: i understand that this isn't a fk because we do want to allow deleting the old blueprints?

tyacbovi: Indeed, like in the case of updating a dep to new blueprint and deleting the old one because we don't need it, but want the log

PR #1918, opened 2019-09-02
CY-1687 Use flask's error handling

tehasdf:
Instead of the mix of flask's, flask-restful's (and a function to
choose between the two), and a custom decorator, only use flask's
error handling.


tehasdf:
This touches many files because it removes all usages of `exceptions_handled`, so make sure to read each commit separately to keep sanity

PR #1919, opened 2019-09-04
CY-1710 Reevaluate operation inputs on deployment update

tehasdf:
See each commit message for details

### DIFF
--- a/rest-service/manager_rest/deployment_update/handlers.py
+++ b/rest-service/manager_rest/deployment_update/handlers.py
@@ -1,9 +1,13 @@
 
 from copy import deepcopy
+from sqlalchemy import and_, cast
+from sqlalchemy.dialects.postgresql import JSON
 
 import utils as deployment_update_utils
 
 from cloudify.constants import COMPUTE_NODE_TYPE
+from cloudify.workflows import tasks as cloudify_tasks
+from cloudify.models_states import ExecutionState
 
 from dsl_parser.constants import (HOST_AGENT,
                                   PLUGIN_INSTALL_KEY,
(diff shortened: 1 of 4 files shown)


rest-service/manager_rest/deployment_update/handlers.py
```diff
@@ -184,6 +187,51 @@ def modify(self, ctx, current_entities):
...
+                      cloudify_tasks.TASK_PENDING]
+        })
+        new_op_inputs = new_operation['inputs']
+        for op in ops:
+            try:
+                op.parameters['task_kwargs']['kwargs'].update(new_op_inputs)
+                op.parameters['task_kwargs']['kwargs']['__cloudify_context'][
+                    'has_intrinsic_functions'] = True
+                op.parameters['task_kwargs']['cloudify_context'][
+                    'has_intrinsic_functions'] = True
```
tehasdf: it's currently in these two places, CY-1711 is about removing one of those places

rest-service/manager_rest/deployment_update/handlers.py
```diff
@@ -184,6 +187,51 @@ def modify(self, ctx, current_entities):
...
+            'status': [
+                ExecutionState.PENDING,
+                ExecutionState.STARTED,
+                ExecutionState.CANCELLED,
+                ExecutionState.FAILED
+            ]
+        })
+        graphs = self.sm.list(models.TasksGraph, filters={
+            'execution_id': [e.id for e in executions]
+        })
```
tehasdf: unfortunate that we have to do essentially joins on the python side, but that's what the storage-manager makes us do...

rest-service/manager_rest/deployment_update/handlers.py
```diff
@@ -184,6 +187,51 @@ def modify(self, ctx, current_entities):
...
+            # path in the parameters dict that stores the node name
+            node_name_path = ('task_kwargs', 'cloudify_context', 'node_name')
+            # this will use postgres' json operators
+            return cast(column, JSON)[node_name_path].astext == node.id
+
+        if not ctx.deployment_update.update_executions:
+            return
+        executions = self.sm.list(models.Execution, filters={
+            'deployment_id': ctx.deployment_update.deployment_id,
+            'status': [
```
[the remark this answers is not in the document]
tehasdf: queued and scheduled don't have their operations in the db yet, that only happens when the execution starts

rest-service/manager_rest/deployment_update/handlers.py
tehasdf: this can be removed after cloudify-cosmo/cloudify-common#361

PR #1923, opened 2019-09-08
CY-1725 : Allowing http endpoint auth for readonly access to DB

tyacbovi:


rest-service/manager_rest/security/authentication.py
```diff
@@ -77,7 +77,8 @@ def authenticate(self, request):
             # (User + Password), otherwise the counter will be reset on
             # every UI refresh (every 4 sec) and accounts won't be locked.
             user.failed_logins_counter = 0
-        user.last_login_at = datetime.now()
+        if not readonly:
+            user.last_login_at = datetime.now()
         user_datastore.commit()
```
tehasdf: so this should not commit at all if readonly, because it'll just break if anything is changed on the user

however now we note that the other thing that can be changed here is the failed logins counter, and so we realize that this has security implications because a readonly login allows you to circumvent the failed login block...

rest-service/manager_rest/test/security/test_authentication.py
```diff
@@ -32,8 +33,18 @@
 class AuthenticationTests(SecurityTestBase):
 
     def test_secured_client(self):
+        alice = self.sm.list(management_models.User,
+                             filters={'username': 'alice'})
+        self.assertEqual(len(alice), 1)
+        login_time = alice[0].last_login_at
+        self.assertEqual(login_time, None)
```
tehasdf: `assertIsNone`, `assertIsNotNone`

rest-service/manager_rest/test/security/test_authentication.py
```diff
@@ -32,8 +33,18 @@
...
         self._assert_user_authorized(username='alice',
                                      password='alice_password')
+        alice = self.sm.list(management_models.User,
+                             filters={'username': 'alice'})
+        self.assertEqual(len(alice), 1)
+        login_time = alice[0].last_login_at
+        self.assertNotEqual(login_time, None)
```
tehasdf: also add a unittest that does the same, but with the status endpoint, and checks that the last login didnt change?

and in the test also try passing a wrong password, which currently would 500 if the db was indeed readonly (which it won't be in the unittest)

PR #1929, opened 2019-09-16
CY-1711 Accessing operations: use the right cloudify-context

tehasdf:
The context used to be duplicated in task_kwargs.cloudify_context and
in task_kwargs.kwargs.__cloudify_context.

Now it's only the latter, so stop using the first one, which doesn't
exist anymore.

### DIFF
--- a/tests/integration_tests/tests/agentless_tests/test_depends_on_lifecycle_operation.py
+++ b/tests/integration_tests/tests/agentless_tests/test_depends_on_lifecycle_operation.py
@@ -80,14 +80,14 @@ node_templates:
                 operations_id[op.id]['dependencies'] = op.dependencies
                 operations_id[op.id]['info'] = op.info
 
-                task_kwargs = op.parameters.get('task_kwargs', {})
-                if task_kwargs.get('cloudify_context', None):
-                    node_task = op.parameters['task_kwargs']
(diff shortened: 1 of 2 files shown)


tests/integration_tests/tests/agentless_tests/test_depends_on_lifecycle_operation.py
```diff
@@ -80,14 +80,14 @@ def _test_full_flow(self, expected_info, tested_operation):
...
-                task_kwargs = op.parameters.get('task_kwargs', {})
-                if task_kwargs.get('cloudify_context', None):
-                    node_task = op.parameters['task_kwargs']
-                    cloudify_context = node_task['cloudify_context']
-                    op_name = cloudify_context['operation']['name']
-                    node_name = cloudify_context['node_name']
-                    operations_info[(op_name,
-                                     node_name)] = op.containing_subgraph
+                try:
+                    cloudify_context = op.parameters['task_kwargs'][
```
tyacbovi: What do u say about doing ".get"?

tehasdf: it would have to be a chain of 3 get calls, so this is better

PR #1933, opened 2019-09-18
CY-1632 Fix wagon files not being deleted after plugin installation fails

davidginzbourg:


tests/integration_tests/tests/agentless_tests/test_plugins.py
```diff
@@ -83,28 +93,86 @@ def test_delete_plugin_not_found(self):
...
+        self._uninstall_plugin_and_assert_values(plugin, 1)
+        self._uninstall_plugin_and_assert_values(plugin2, 0)
+
+    def _upload_plugin_and_assert_values(self,
+                                         package_name,
+                                         package_version,
+                                         plugins_count,
+                                         wagon_files_count,
+                                         corrupt_plugin=False):
+        _plugin = test_utils.upload_mock_plugin(
```
tyacbovi: _plugin => plugin it's only in that scope

tests/integration_tests/tests/agentless_tests/test_plugins.py
```diff
@@ -83,28 +93,86 @@ def test_delete_plugin_not_found(self):
...
+            package_name,
+            package_version,
+            corrupt_plugin=corrupt_plugin)
+        wagon_file = _plugin.archive_name[len(INSTALLING_PREFIX):]
+
+        ex = self._get_latest_execution('install_plugin')
+        if corrupt_plugin:
+            with self.assertRaisesRegexp(
+                    RuntimeError, 'Workflow execution failed: .*'):
+                self.wait_for_execution_to_end(ex)
```
tyacbovi: won't u need always to wait for the execution to finish?

tests/integration_tests/tests/agentless_tests/test_plugins.py
```diff
@@ -83,28 +93,86 @@ def test_delete_plugin_not_found(self):
...
+            ex = self._get_latest_execution('uninstall_plugin')
+        self.wait_for_execution_to_end(ex)
+
+        find_output = self.execute_on_manager(
```
tyacbovi: function out the search for garbage plugin files

tests/integration_tests/tests/agentless_tests/test_plugins.py
```diff
@@ -83,28 +93,86 @@ def test_delete_plugin_not_found(self):
...
         plugins = self.client.plugins.list()
         self.assertEqual(len(plugins), 0)
+
+
+class TestPluginsSystemState(AgentlessTestCase):
+    def test_installing_corrupted_plugin_doesnt_affect_system_integrity(self):
+        self._upload_plugin_and_assert_values(TEST_PACKAGE_NAME,
+                                              TEST_PACKAGE_VERSION,
+                                              plugins_count=0,
+                                              # Wagon has a bug that leaves 1
```
tyacbovi: I will remove the comments and put them in the commit msg, because it will be more clear there for future fixes

tests/integration_tests/tests/agentless_tests/test_plugins.py
```diff
@@ -30,6 +34,12 @@ def test_get_plugin_by_id(self):
             put_plugin_response = test_utils.upload_mock_plugin(
                     TEST_PACKAGE_NAME,
                     TEST_PACKAGE_VERSION)
+            ex = self._get_latest_execution('install_plugin')
```
tyacbovi: I guess this section is not relevant anymore

workflows/cloudify_system_workflows/plugins.py
```diff
@@ -21,7 +21,15 @@
 
 @workflow(system_wide=True)
 def install(ctx, plugin, **_):
-    return _operate_on_plugin(ctx, plugin, 'install')
+    try:
+        execution_result = _operate_on_plugin(ctx, plugin, 'install')
+    except Exception as e:
+        client = get_rest_client()
+        plugins = client.plugins.list(id=plugin['id'])
```
tyacbovi: why not plugin.get?

workflows/cloudify_system_workflows/plugins.py
```diff
@@ -21,7 +21,15 @@
...
+        if plugins:
+            client.plugins.delete(plugin_id=plugin['id'], force=True)
```
tyacbovi: Add some logging msg

rest-service/manager_rest/upload_manager.py
```diff
@@ -831,7 +844,7 @@ def _prepare_and_process_doc(self,
                                  archive_target_path,
                                  **kwargs):
         plugins = []
-        caravan_ = self.Caravan(archive_target_path)
+        caravan_ = kwargs['caravan_']
```
tyacbovi: this is a bit unclear, what will be returned here?

tests/integration_tests/tests/test_cases.py
```diff
@@ -541,6 +541,22 @@ def _reset_file_system(self):
         self.clear_directory('/opt/manager/resources/blueprints')
         self.clear_directory('/opt/manager/resources/uploaded-blueprints')
 
+    def _get_latest_execution(self, workflow_id):
+        ex_list = self.client.executions.list(
```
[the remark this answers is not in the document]
geokala: I'm assuming you mean equivalent rather than aberration?
(for the actual point in question, I'm 50/50- in this context it doesn't feel too unclear to me but I can definitely understand confusion if you're juggling another part of the code at the same time)

tests/integration_tests/tests/agentless_tests/test_plugins.py
geokala: I'd suggest putting them in a jira as well.
Either way, leaving them next to the code that would look wrong without them seems like a better idea to me than putting them in a commit message, meaning that someone looking at the code has to realise that the code is right, or go and do a git blame and hope there wasn't a refactor since a random comment about the code was left in a commit message.

tests/integration_tests/tests/test_cases.py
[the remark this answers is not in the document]
tyacbovi: I meant abbreviation :). I think it's better to always of the clearer side and not conflict with some standards in other parts of the code

tests/integration_tests/tests/agentless_tests/test_plugins.py
tyacbovi: If the code had changed with those lines, so those the commit msg needs to add info about that.
And there is always git history :)

tests/integration_tests/tests/test_cases.py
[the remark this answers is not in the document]
davidginzbourg: I keep forgetting to not use "ex". I hope I'll remember it this time :)

tests/integration_tests/tests/agentless_tests/test_plugins.py
davidginzbourg: This code block always waits for the `install_plugin` execution. If that's what you mean

tests/integration_tests/tests/agentless_tests/test_plugins.py
davidginzbourg: Why not?

workflows/cloudify_system_workflows/plugins.py
davidginzbourg: Just to make sure that this doesn't fail on plugin.get if the plugin entry doesn't exist. This way the correct error will be raised

rest-service/manager_rest/upload_manager.py
davidginzbourg: Better?

tests/integration_tests/tests/agentless_tests/test_plugins.py
tyacbovi: So why do we need to add here the wait for execution?

PR #1937, opened 2019-09-22
execute_wf : fix wrong usage

tyacbovi:
- execute_wf tried to use a plugin, but there is no plugin to use
- execute_wf for cascading wf of Component was given the wrong blueprint_id

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -712,7 +712,7 @@ class ResourceManager(object):
             for component_dep_id in components_dep_ids:
                 self.execute_workflow(component_dep_id,
                                       workflow_id,
-                                      blueprint_id,
+                                      None,
                                       parameters,
                                       allow_custom_parameters,
                                       force,
--- a/rest-service/manager_rest/workflow_executor.py
+++ b/rest-service/manager_rest/workflow_executor.py
@@ -63,7 +63,8 @@ def execute_workflow(name,
         'is_system_workflow': False,
         'wait_after_fail': wait_after_fail,
         'resume': resume,
-        'execution_token': generate_execution_token(execution_id)
+        'execution_token': generate_execution_token(execution_id),
+        'plugin': {}
     }
 
     plugin_name = workflow['plugin']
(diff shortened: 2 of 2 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -712,7 +712,7 @@ def execute_workflow(self,
             for component_dep_id in components_dep_ids:
                 self.execute_workflow(component_dep_id,
                                       workflow_id,
-                                      blueprint_id,
+                                      None,
```
tehasdf: @tyacbovi how about fetch the deployment and pass the right blueprint id, no reason to do a half-solution?

rest-service/manager_rest/workflow_executor.py
```diff
@@ -75,17 +64,35 @@ def execute_workflow(name,
...
+    if plugin is not None and plugin['package_name']:
+        sm = get_storage_manager()
+        filter_plugin = {'package_name': plugin.get('package_name'),
+                         'package_version': plugin.get('package_version')}
+        managed_plugins = sm.list(models.Plugin, filters=filter_plugin).items
+        if managed_plugins:
+            plugin['visibility'] = managed_plugins[0].visibility
+            plugin['tenant_name'] = managed_plugins[0].tenant_name
+
+        context.update({
```
tehasdf: this can be written simpler, just `context['plugin'] = {'name': ...}`

rest-service/manager_rest/resource_manager.py
tyacbovi: Well it does that :)
https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/resource_manager.py#L635

rest-service/manager_rest/workflow_executor.py
tyacbovi: yep, updated the context and forgot about this

rest-service/manager_rest/resource_manager.py
tehasdf: :+1:

PR #1938, opened 2019-09-22
CY-1598 Update the single manager status command

inbala:
* Check the connection of RabbitMQ and not only the
  service

* Don't check the DB connection because passing our
  authentication implies PostgreSQL is healthy

* Check syncthing functionallity - valid when at least
  one device was seen recently

* Change the response format

### DIFF
--- a/rest-service/manager_rest/rest/resources_v1/executions.py
+++ b/rest-service/manager_rest/rest/resources_v1/executions.py
@@ -14,6 +14,8 @@
 #  * limitations under the License.
 #
 
+from datetime import datetime
+
 from flask_restful.reqparse import Argument
 from flask_restful_swagger import swagger
 from flask_restful.inputs import boolean
@@ -30,7 +32,7 @@ from manager_rest.rest.rest_utils import (
     get_args_and_verify_arguments,
     get_json_and_verify_params,
     verify_and_convert_bool,
-    parse_datetime
+    parse_datetime_string
 )
 from manager_rest.security import SecuredResource
 from manager_rest.security.authorization import authorize
@@ -105,7 +107,7 @@ class Executions(SecuredResource):
         scheduled_time = request_dict.get('scheduled_time', None)
 
         if scheduled_time:
-            scheduled_time = parse_datetime(scheduled_time)
+            scheduled_time = self._parse_scheduled_time(scheduled_time)
 
         if parameters is not None and parameters.__class__ is not dict:
             raise manager_exceptions.BadParametersError(
@@ -120,6 +122,16 @@ class Executions(SecuredResource):
             queue=queue, wait_after_fail=wait_after_fail,
             scheduled_time=scheduled_time)
         return execution, 201
+
+    def _parse_scheduled_time(self, scheduled_time):
+        scheduled_utc = parse_datetime_string(scheduled_time)
+        if scheduled_utc <= datetime.utcnow():
+            raise manager_exceptions.BadParametersError(
+                'Date `{0}` has already passed, please provide'
+                ' valid date. \nExpected format: YYYYMMDDHHMM+HHMM or'
+                ' YYYYMMDDHHMM-HHMM i.e: 201801012230-0500'
+                ' (Jan-01-18 10:30pm EST)'.format(scheduled_time))
+        return scheduled_utc
 
 
 class ExecutionsId(SecuredResource):
--- a/rest-service/manager_rest/rest/resources_v1/status.py
+++ b/rest-service/manager_rest/rest/resources_v1/status.py
@@ -22,8 +22,8 @@ from cloudify.constants import BROKER_PORT_SSL
 
 from manager_rest import config
 from manager_rest.rest import responses
+from manager_rest.security import SecuredResource
 from manager_rest.rest.rest_decorators import marshal_with
-from manager_rest.security import SecuredResourceReadonlyMode
 from manager_rest.security.authorization import authorize
 try:
     from manager_rest.systemddbus import get_services
@@ -44,7 +44,7 @@ OPTIONAL_SERVICES = {
 }
 
 
-class Status(SecuredResourceReadonlyMode):
+class Status(SecuredResource):
 
     @swagger.operation(
         responseClass=responses.Status,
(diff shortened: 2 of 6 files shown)


rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -0,0 +1,170 @@
...
+            current_app.logger.error(
+                'Syncthing check failed with {err_type}: {err_msg}'.format(
+                    err_type=type(err),
+                    err_msg=str(err),
+                )
+            )
+            self._add_or_update_service(services,
+                                        'File Sync Service',
+                                        INACTIVE_STATE)
+            return FAIL_STATE
```
tyacbovi: This should be inactive ? No?

rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -0,0 +1,170 @@
...
+
+            # Syncthing is valid when at least one device was seen recently
+            if last_seen > min_last_seen:
+                syncthing_status = ACTIVE_STATE
+                break
+
+        self._add_or_update_service(services,
+                                    'File Sync Service',
+                                    syncthing_status)
+        return self._get_updated_status(manager_status, syncthing_status)
```
tyacbovi: Like I said this should not be inside

rest-service/manager_rest/test/endpoints/test_status.py
```diff
@@ -21,14 +21,16 @@
 from manager_rest.storage import management_models
 
 
-@attr(client_min_version=1, client_max_version=base_test.LATEST_API_VERSION)
+@attr(client_min_version=3.1, client_max_version=base_test.LATEST_API_VERSION)
```
tyacbovi: Should keep old tests

rest-service/manager_rest/test/endpoints/test_status.py
```diff
@@ -13,22 +13,40 @@
...
 
 from manager_rest.test import base_test
 from manager_rest.test.attribute import attr
 from manager_rest.storage import management_models
 
 
-@attr(client_min_version=1, client_max_version=base_test.LATEST_API_VERSION)
-class StatusTestCase(base_test.BaseServerTestCase):
+@attr(client_min_version=1, client_max_version=3)
+class PreviousStatusTestCase(base_test.BaseServerTestCase):
```
tyacbovi: PreviousStatusTestCase => StatusV1TestCase

rest-service/manager_rest/test/endpoints/test_status.py
```diff
@@ -13,22 +13,40 @@
...
 
     def test_get_status(self):
-        self._assert_last_login_time_value(None)
+        before_status = datetime.utcnow()
         result = self.client.manager.get_status()
+        last_login_at = self.sm.get(management_models.User,
+                                    current_user.id).last_login_at
+        after_status = datetime.strptime(last_login_at,
```
tyacbovi: How did it work before?

rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -147,8 +147,8 @@ def _check_syncthing(self, services):
             self._add_or_update_service(services, display_name, status)
             return status
 
-        interval = syncthing_config['options']['reconnectionIntervalS']
-        min_last_seen = datetime.utcnow() - timedelta(seconds=interval + 1)
+        interval = syncthing_config['options']['reconnectionIntervalS'] + 1
```
tyacbovi: Mabe put a comment of the reason here

rest-service/manager_rest/rest/resources_v1/executions.py
```diff
@@ -121,6 +123,16 @@ def post(self, **kwargs):
             scheduled_time=scheduled_time)
         return execution, 201
 
+    def _parse_scheduled_time(self, scheduled_time):
```
geokala: Shouldn't this be in a separate PR?

rest-service/manager_rest/rest/resources_v1/status.py
```diff
@@ -78,32 +78,34 @@ def get(self, **kwargs):
...
+    # If we reach here then the service is optional and not installed
+    return False
 
-    def _get_systemd_manager_services(self):
-        """Services the status of which we keep track of.
 
-        :return: a dict of {service_name: label}
-        """
-        services = {}
+def get_systemd_manager_services(base_services=BASE_SERVICES,
```
geokala: We shouldn't be setting default arg values to mutable values. At the moment, we don't break it, but it's generally bad practise as it can lead to unexpected interesting situations (TM) in future.

rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -0,0 +1,175 @@
...
+
+        self._add_or_update_service(services, display_name, status)
+        return status
+
+    def _get_manager_status(self, systemd_statuses, rabbitmq_status,
+                            syncthing_status):
+        statuses = systemd_statuses.union({rabbitmq_status, syncthing_status})
+
+        # statuses set can maximum have 2 statuses Active and Inactive
+        return HEALTHY_STATE if len(statuses) == 1 and \
```
geokala: Maybe better as: return HEALTHY_STATE if all(status == HEALTHY_STATE for status in statuses) else FAIL_STATE
?

rest-service/manager_rest/rest/resources_v1/executions.py
inbala: I needed it for the syncthing check in GET /status

rest-service/manager_rest/rest/resources_v3_1/status.py
inbala: But then I don't use the advantage of the set

geokala: I think it's more readable with the all(x) approach though? Is there somewhere else that we get significant advantage from making this a set?

rest-service/manager_rest/rest/resources_v1/executions.py
geokala: Fair enough, I'll mark this resolved then.

PR #1954, opened 2019-10-07
IPSetter: only reload nginx if it is up

tehasdf:
No reason to otherwise, and it actually breaks (returns 1) if it's not up


tehasdf:
moving this to cfy_manager

PR #1966, opened 2019-10-22
Cy 1814 plugin bundle cluster failure

glukhman:


rest-service/manager_rest/resource_manager.py
```diff
@@ -279,6 +280,8 @@ def install_plugin(self, plugin):
             self._validate_plugin_yaml(plugin)
 
         if not utils.plugin_installable_on_current_platform(plugin):
+            remove_status_prefix(plugin)
```
barucoh: Remove from this PR and create a different PR for it as discussed.
For the different PR, a plugin should be deleted from the DB if it cannot be installed.

packaging/mgmtworker/requirements.txt
```diff
@@ -1,5 +1,5 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==master
-git+https://github.com/cloudify-cosmo/cloudify-agent@master#egg=cloudify-agent[celery,kerberos]==master
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-1814-Plugin-bundle-cluster-failure#egg=cloudify-common==CY-1814-Plugin-bundle-cluster-failure
```
barucoh: Don't forget to change back to master before you merge

rest-service/manager_rest/resource_manager.py
```diff
@@ -26,7 +26,8 @@
 
 from cloudify.cryptography_utils import encrypt
 from cloudify.workflows import tasks as cloudify_tasks
-from cloudify.plugins.install_utils import INSTALLING_PREFIX
+from cloudify.plugins.install_utils import (INSTALLING_PREFIX,
```
barucoh: flake8 will fail after you remove the code at the bottom

rest-service/manager_rest/resource_manager.py
barucoh: OK After some further discussions, this solution should suffice, no need to delete from the DB.

PR #1969, opened 2019-10-24
CY-1856 Create db_nodes table

inbala:


rest-service/manager_rest/storage/management_models.py
```diff
@@ -616,6 +617,24 @@ class RabbitMQBroker(_WithCACert, SQLModelBase):
     # additional params, as **kwargs to creating the connection
     params = db.Column(JSONString, nullable=True)
     networks = db.Column(JSONString)
+    node_id = db.Column(db.Text, unique=True, nullable=False)
+    is_external = db.Column(db.Boolean, default=False, nullable=False)
+
+    @classmethod
+    def unique_id(cls):
```
tyacbovi: why do we need this func and the next one?

inbala: Because we don't have `id` field and we inherit SQLModelBase.

PR #1983, opened 2019-11-03
CY-1890 Fix allowed specific endpoints check in maintenance mode or with no license

inbala:
* This check used to approve endpoints that included allowed endpoints.
  For example, the 'status' endpoint is allowed, so this check approved
  also the new (not merged yet) 'cluster-status' endpoint.


tyacbovi:
Maybe can we add tests for this?

PR #1990, opened 2019-11-05
update docker image information and instructions

EarthmanT:


tests/integration_tests/resources/dockerfiles/__init__.py
```diff
@@ -13,30 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from io import StringIO
+from tempfile import mkdtemp
```
tyacbovi: small style notice, please separate 3rd party imports and order by len

PR #1995, opened 2019-11-12
CY-1906 Make no-op not an error when running a plugins update

davidginzbourg:


rest-service/manager_rest/plugins_update/manager.py
```diff
@@ -115,38 +115,46 @@ def initiate_plugins_update(self, blueprint_id, force=False):
         self.validate_no_active_updates_per_blueprint(blueprint_id, force)
         blueprint = self.sm.get(models.Blueprint, blueprint_id)
         temp_plan = self.get_reevaluated_plan(blueprint)
+        no_op = not _did_plugins_to_install_change(temp_plan, blueprint.plan)
```
tyacbovi: maybe no_op => no_change or =>nothing_todo

rest-service/manager_rest/plugins_update/manager.py
```diff
@@ -115,38 +115,46 @@ def initiate_plugins_update(self, blueprint_id, force=False):
...
         * Changes all the deployments' blueprint back from the temp blueprint
            to the original one
         * Deletes the temporary blueprint
         * Updates the plugins update state
+
+        If this is a no-op then it just updates the plugins update state to
+        NO_CHANGES_REQUIRED.
         """
-        plugins_update = self.sm.get(models.PluginsUpdate, plugins_update_id)
+        if no_op:
```
tyacbovi: why not function this out and use it instead of line 141? and not change this API

rest-service/manager_rest/resource_manager.py
```diff
@@ -964,6 +971,10 @@ def _execute_system_workflow(self,
             if deployment:
                 execution.set_deployment(deployment)
 
+        if no_op:
```
tyacbovi: Can it be done only in the plugins update logic? because this is a general place and no_op is not available mode for other wf

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -118,12 +118,11 @@ def test_raises_with_nonexisting_blueprint(self):
...
-    def test_raises_blueprint_has_no_deployments(self):
+    def test_no_op_blueprint_has_no_deployments(self):
         self.put_file(*self.put_blueprint_args(blueprint_id='hello_world'))
-        with self.assertRaisesRegexp(CloudifyClientError,
-                                     "The blueprint 'hello_world' has no "
-                                     "deployments to update."):
-            self.client.plugins_update.update_plugins("hello_world")
+        plugins_update = self.client.plugins_update.update_plugins(
+            'hello_world')
+        self.assertEqual(plugins_update.state, STATES.NO_CHANGES_REQUIRED)
```
tyacbovi: Stage shows this state?

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -154,6 +153,30 @@ def test_raises_while_plugins_updates_are_active(self):
...
+        plugins_update.state = STATES.SUCCESSFUL
+        self._sm.update(plugins_update)
+
+        plugins_update_id = self.client.plugins_update.update_plugins(
+            'hello_world').id
+        plugins_update = self._sm.get(models.PluginsUpdate, plugins_update_id)
+        plugins_update.state = STATES.FAILED
+        self._sm.update(plugins_update)
+
+        plugins_update_id = self.client.plugins_update.update_plugins(
```
tyacbovi: may does this make sense? would not it "fix" the update?

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -177,16 +200,14 @@ def test_raises_with_real_active_plugins_updates_and_forced(self):
                 'update IDs: {0}'.format(plugins_update.id)):
             self.client.plugins_update.update_plugins('hello_world', True)
 
-    def test_raises_when_no_plugin_change_detected(self):
+    def test_no_op_when_no_plugin_change_detected(self):
```
tyacbovi: add another test for two no_op requests

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -105,8 +105,10 @@ def post(self, id, phase):
             return get_plugins_updates_manager().initiate_plugins_update(
                 blueprint_id=id, force=args.get('force'))
         elif phase == PHASES.FINAL:
+            sm = get_storage_manager()
```
tyacbovi: what was the issue here

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: I don't think there's a UI for this, at least _I_ haven't created anything for it

rest-service/manager_rest/resource_manager.py
davidginzbourg: Well, I'd have to copy code from this function into the "update_plugin" function, or at least break this function to other functions which is more of a mess IMO

rest-service/manager_rest/rest/resources_v3_1/plugins.py
davidginzbourg: Change of API, I'm reverting this

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: Why it makes sense* ?

This just checks that an update can be run after a "non-active" plugin update has been made, which is similar to what you thought _should_ happen.

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: Though this is necessary IMO, one test for "no deployments to update" and one for "no plugins to update" - two different scenarios.

davidginzbourg: `test_no_op_blueprint_has_no_deployments` is the "no deployments to update" scenario.

rest-service/manager_rest/plugins_update/manager.py
davidginzbourg: Reverting

rest-service/manager_rest/plugins_update/manager.py
davidginzbourg: Changed

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
tyacbovi: Oh, so you need to split this test to the different scenerios you are testing for.
It's not clear

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -154,6 +153,30 @@ def test_raises_while_plugins_updates_are_active(self):
                 '{0}'.format(plugins_update.id)):
             self.client.plugins_update.update_plugins('hello_world')
 
+    def test_doesnt_raise_when_plugins_updates_are_not_active(self):
```
tyacbovi: not a good name, not clear what it's testing for

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
tyacbovi: I mean try to plugin update twice when every time you will get nothing todo

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
```diff
@@ -118,12 +118,11 @@ def test_raises_with_nonexisting_blueprint(self):
             self.client.plugins_update.update_plugins("non_existing")
             self.assertEqual(404, e.exception.status_code)
 
-    def test_raises_blueprint_has_no_deployments(self):
+    def test_no_op_blueprint_has_no_deployments(self):
```
tyacbovi: not clear of a name
=> test_basic_noting_to_plugin_update

rest-service/manager_rest/resource_manager.py
tyacbovi: I think you just need to use the data model and write it, no?

rest-service/manager_rest/resource_manager.py
```diff
@@ -294,10 +294,11 @@ def install_plugin(self, plugin):
             verify_no_executions=False,
             timeout=300)
 
-    def update_plugins(self, plugins_update):
+    def update_plugins(self, plugins_update, no_op=False):
```
tyacbovi: still no_op

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: The plugin update execution doesn't even execute. I've added a state check to the integration tests.

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: I split the tests

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: Split and changed the names

rest-service/manager_rest/test/endpoints/test_plugins_updates.py
davidginzbourg: Changed to `test_plugins_update_does_nothing_when_no_deployments_to_update`

rest-service/manager_rest/resource_manager.py
davidginzbourg: I could, just wanted the execution to have as many 'real' parameters as it can have so I won't break inconsistencies in the DB.

rest-service/manager_rest/resource_manager.py
davidginzbourg: Changed to `no_changes_required`

PR #2000, opened 2019-11-19
CY-1700 Add cluster-status endpoint

inbala:


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,237 @@
...
+from manager_rest.rest.rest_utils import parse_datetime_string
+
+
+DB_SERVICE_KEY = 'PostgreSQL'
+BROKER_SERVICE_KEY = 'RabbitMQ'
+UNINITIALIZED_STATUS = 'Uninitialized'
+CLUSTER_STATUS_PATH = '/opt/manager/resources/cluster_status'
+
+
+class ClusterStatusManager(object):
```
tehasdf: in general i don't see a reason for this to be a class, and passing through things between methods as attributes on an instance (and there is ever only one instance at a time anyway)

i think this really wants to just be a function, and all its methods would be just functions too, and they would simply take arguments instead of accessing instance attributes

the problem is that this is harder to read because looking at a method eg. `_handle_missing_non_manager_files`, it accesses `self.missing_status_files`, and now i need to examine the rest of the class to find where is `.missing_status_files` set, instead of just looking at the caller

so all in all i prefer more structure, by passing arguments, instead of a free-for-all accessing anything anywhere

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,237 @@
...
+            if not self._is_report_valid(node_status):
+                self._generate_node_status(node, formatted_nodes,
+                                           status=ServiceStatus.FAIL)
+                continue
+
+            report = node_status['report']
+            self._generate_node_status(node, formatted_nodes,
+                                       report['services'], report['status'])
+
+        is_external = service_nodes[0].to_dict().get('is_external', False)
```
tehasdf: no real reason for the `.to_dict`? vs `service_nodes[0].is_external`

tehasdf: same for other usages of `.to_dict` in the file

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,237 @@
...
+        return (self.cluster_structure[ServiceType.DB][0].node_id ==
+                self.cluster_structure[ServiceType.BROKER][0].node_id ==
+                self.cluster_structure[ServiceType.MANAGER][0].node_id)
+
+    def _get_service_status(self, formatted_nodes):
+        # A set of the service node statuses
+        service_statuses = {node['status'] for name, node in
+                            formatted_nodes.items()}
+
+        # If there is only one type of status it is it, else Degraded
```
tehasdf: this comment doesn't really help, because it just says the same thing that the code says

instead if you need a comment here, it should talk about why is it degraded if there's not exactly 1 node: i'd think having 3 nodes isn't degraded? we want to have 3 db nodes, 3 rabbit nodes...

rest-service/manager_rest/rest/resources_v3_1/cluster_status.py
```diff
@@ -68,15 +72,37 @@ def put(self, node_id, model, node_type):
...
-            self._verify_report_newer_than_current(report_time, path)
+        status_path = '{path}/{node_type}_{node_id}.json'.format(
+            path=CLUSTER_STATUS_PATH, node_type=node_type, node_id=node_id)
+        if os.path.exists(CLUSTER_STATUS_PATH):
+            self._verify_report_newer_than_current(report_time, status_path)
         else:
-            os.makedirs(STATUSES_PATH)
-        with open(path, 'w') as report_file:
+            os.makedirs(CLUSTER_STATUS_PATH)
+        with open(status_path, 'w') as report_file:
```
tehasdf: the writing should also probably be implemented in the `manager` module, where the reading is, instead of here

rest-service/manager_rest/cluster_status_manager.py
inbala: It is for cases it doesn't have this column, for example the managers table in the DB doesn't have the `is_external` column and the rabbitmq_brokers and db_nodes don't have the `public_ip` column.

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,240 @@
...
+
+        manager_nodes = manager_service['nodes']
+        for manager in missing_managers:
+            manager_nodes[manager.hostname]['status'] = ServiceStatus.FAIL
+        manager_service['status'] = self._get_service_status(manager_nodes)
+
+    def _handle_missing_non_manager_files(self, service_type, service_key,
+                                          managers):
+        """
+        Missing status file because it is an external service (user-provided)
```
tyacbovi: =>A node's status file could be missing, due to it being an external service (user-provided)
        or all-in-one manager or it's status reporter didn't send a report ey

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        Time since reported should be maximum twice the frequency interval
+        (Nyquist sampling theorem)
+        """
+        reporting_delta = node_status['reporting_freq'] * 2
+        report_time = parse_datetime_string(node_status['timestamp'])
+        return (report_time + timedelta(seconds=reporting_delta) >
+                datetime.utcnow())
+
+    def _keys_in_dict(self, dictionary, keys):
```
tyacbovi: _keys_in_dict => _are_keys_in_dict

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,240 @@
...
+                                     NodeServiceStatus)
+
+from manager_rest.storage import models, get_storage_manager
+from manager_rest.rest.rest_utils import parse_datetime_string
+
+
+DB_SERVICE_KEY = 'PostgreSQL'
+BROKER_SERVICE_KEY = 'RabbitMQ'
+UNINITIALIZED_STATUS = 'Uninitialized'
+CLUSTER_STATUS_PATH = '/opt/manager/resources/cluster_status'
```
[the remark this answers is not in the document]
inbala: They are using the same one.

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,240 @@
...
+        Generate the cluster status using:
+        1. The DB tables (managers, rabbitmq_brokers, db_nodes) for the
+           structure of the cluster.
+        2. The status files each reporter sends with the most updated status
+           of the node.
+        """
+        self.cluster_status = {'status': ServiceStatus.DEGRADED,
+                               'services': {}}
+
+        if not path.isdir(CLUSTER_STATUS_PATH):
```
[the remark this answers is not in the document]
inbala: We talked about it with Ofer and Barak that it should be Degraded if the reporters haven't started yet.

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,240 @@
...
+        if len(service_statuses) == 1:
+            return service_statuses.pop()
+        return ServiceStatus.DEGRADED
+
+    def _is_report_updated(self, node_status):
+        """
+        Time since reported should be maximum twice the frequency interval
+        (Nyquist sampling theorem)
+        """
+        reporting_delta = node_status['reporting_freq'] * 2
```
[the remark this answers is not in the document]
inbala: Oh in my status files it is int, you think it could be a string?

rest-service/manager_rest/cluster_status_manager.py
inbala: I agree, in the beginning I thought I'll use more the instance variables but now I see you are right and I changed it to be just a function.

rest-service/manager_rest/cluster_status_manager.py
inbala: I added the `CloudifyNodeMixin` and now I don't need the `to_dict`

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,240 @@
...
+        is_external = service_nodes[0].to_dict().get('is_external', False)
+        self.cluster_status['services'][service_type] = {
+            'status': self._get_service_status(formatted_nodes),
+            'is_external': is_external,
+            'nodes': formatted_nodes
+        }
+
+    def _generate_node_status(self, node, formatted_nodes, node_services=None,
+                              status=UNINITIALIZED_STATUS):
+        node_dict = node.to_dict()
```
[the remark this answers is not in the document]
inbala: I added `CloudifyNodeMixin` so now I don't need this

rest-service/manager_rest/cluster_status_manager.py
inbala: It checks the different statuses type, not matter how many node there are.
If all the node statuses are the same = OK or Fail this will be the service status.
Else it means some Fail and some OK and the service status is Degraded.

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -0,0 +1,240 @@
...
+                            formatted_nodes.items()}
+
+        # If there is only one type of status it is it, else Degraded
+        if len(service_statuses) == 1:
+            return service_statuses.pop()
+        return ServiceStatus.DEGRADED
+
+    def _is_report_updated(self, node_status):
+        """
+        Time since reported should be maximum twice the frequency interval
```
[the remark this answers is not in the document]
inbala: Is
```
    The report is considered updated, if the time passed since the report 
    was sent is maximum twice the frequency interval (Nyquist sampling theorem)
```
ok?

rest-service/manager_rest/cluster_status_manager.py
tyacbovi: Only the content of the comment was not clear

rest-service/manager_rest/storage/management_models.py
```diff
@@ -588,7 +588,27 @@ def write_ca_cert(self):
         return f.name
 
 
-class Manager(_WithCACert, SQLModelBase):
+class CloudifyNodeMixin(object):
+    """A mixin for Cloudify cluster node objects."""
+
+    @property
+    def name(self):
+        return self.name
```
tehasdf: those properties by themselves seem useless because if you do access `.name` then of course it just returns `self.name` so that's just unlimited recursion;  it's only not unlimited where subclasses override this, so i suppose it would make more sense for these methods to just return none

inbala: Good point, I'll fix it

PR #2002, opened 2019-11-19
CY-1918: switch to using http client

tyacbovi:


status-reporter/status_reporter/status_reporter.py
```diff
@@ -138,18 +135,21 @@ def _report(self):
...
         # let's try to avoid using the same manager always.
         random.shuffle(self._managers_ips)
+        reporting_result = False
         for manager_ip in self._managers_ips:
             try:
-                self._report_status(manager_ip, report)
-                self._update_managers_ips_list(manager_ip)
-                return
+                client = self._get_cloudify_http_client(manager_ip)
+                reporting_result = self._report_status(client, report)
```
inbala: Remember the break :)

status-reporter/status_reporter/status_reporter.py
```diff
@@ -138,18 +135,21 @@ def _report(self):
...
             except Exception as e:
                 logger.debug('Error had occurred while trying to report '
-                             'status and update active managers-ips list: {}'
+                             'status: {0}'
                              .format(e))
-        logger.error('Could not find an active manager to '
-                     'report the current status,'
-                     ' tried %s', ','.join(self._managers_ips))
+        if reporting_result:
+            self._update_managers_ips_list(client)
```
inbala: Maybe move it after `self._report_status(client, report)` and then the `reporting_result` boolean it redundant.

status-reporter/status_reporter/status_reporter.py
```diff
@@ -138,18 +135,21 @@ def _report(self):
...
+        else:
+            logger.error('Could not find an active manager to '
+                         'report the current status,'
+                         ' tried %s', ','.join(self._managers_ips))
```
inbala: And then for this error use the `else` clause of for loop like:
```
for item in container:
    if search_something(item):
        # Found it!
        process(item)
        break
else:
    # Didn't find anything..
    not_found_in_container()
```

status-reporter/status_reporter/status_reporter.py
tyacbovi: Not really, it will be the same like before

PR #2004, opened 2019-11-19
mgmtworker rpm build

tyacbovi:
aligning the solution for jinja2 affecting the rpm build


packaging/cloudify-mgmtworker.spec
```diff
@@ -38,6 +37,11 @@ STUBS=${RPM_SOURCE_DIR}/packaging/mgmtworker/stub_packages
 
 %{PIP_INSTALL} --upgrade kerberos==1.3.0
 
+# Jinja2 includes 2 files which will only be imported if async is available,
+# but rpmbuild's brp-python-bytecompile falls over when it finds them. Here
+# we remove them.
+rm -f %_manager_env/lib/python2.7/site-packages/jinja2/async*.py
```
tehasdf: not `%_manager_env`, that doesnt exist in this spec

PR #2005, opened 2019-11-21
cluster-status: small refactor

tyacbovi:
- added missing edge case
- small style refactors


rest-service/manager_rest/rest/resources_v3_1/cluster_status.py
```diff
@@ -42,6 +44,9 @@ def _get_request_dict():
 
     @staticmethod
     def _verify_report_newer_than_current(report_time, path):
+        if not (os.path.exists(path) and os.path.isfile(path)):
+            # Nothing to do if the file does not exists
+            return
         with open(path) as current_report_file:
             current_report = json.load(current_report_file)
         if report_time < parse_datetime_string(current_report['timestamp']):
```
YoniItzhak: Should be: 'The new report timestamp `{**0**}` is before  the current report timestamp'.format(report_time)
Same as below.

PR #2012, opened 2019-11-28
CY-1718: implementing manager status reporter

tyacbovi:
- implementing manager status reporter
- fixing missing timestamp in the status report
- fixing node id missing usage 
- fixing the token usage to work
- fixing cluster status endpoint for writing reports

### DIFF
--- a/rest-service/manager_rest/cluster_status_manager.py
+++ b/rest-service/manager_rest/cluster_status_manager.py
@@ -46,6 +46,7 @@ def _are_keys_in_dict(dictionary, keys):
 
 # region Get Cluster Status Helpers
 
+
 def _generate_cluster_status_structure():
     storage_manager = get_storage_manager()
     return {
@@ -178,7 +179,8 @@ def _is_report_content_valid(report):
 
 
 def _is_report_valid(report):
-    return _is_status_report_updated(report) and _is_report_content_valid(report)
+    return (_is_status_report_updated(report) and
+            _is_report_content_valid(report))
 
 
 def _handle_missing_status_reports(missing_status_reports, cluster_status,
--- a/rest-service/manager_rest/cluster_status_manager.py
+++ b/rest-service/manager_rest/cluster_status_manager.py
@@ -40,8 +40,11 @@ def get_report_path(node_type, node_id):
         status_path=CLUSTER_STATUS_PATH, node_type=node_type, node_id=node_id
     )
 
-# region Get Cluster Status Helpers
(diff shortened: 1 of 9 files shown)


status-reporter/status_reporter/status_reporter.py
```diff
@@ -95,35 +98,43 @@ def __init__(self, sampler, node_type):
             raise InitializationError('Failed initialization of status '
                                       'reporter due to:\n {issues}'.
                                       format(issues='\n'.join(issues)))
+        self._node_id = self._config.get('node_id', None)
```
inbala: Don't we want to add the `_node_id` to the required check?

status-reporter/status_reporter/status_reporter.py
```diff
@@ -95,35 +98,43 @@ def __init__(self, sampler, node_type):
...
+        self._reporter_credentials = {
+            'username': self._cloudify_user_name,
```
inbala: Do we need the user_name?

status-reporter/status_reporter/status_reporter.py
tyacbovi: The field is written in the install process

status-reporter/status_reporter/status_reporter.py
tyacbovi: it requires that

status-reporter/status_reporter/status_reporter.py
inbala: Like managers_ips, cloudify_user_name and token, no?

tyacbovi: Well I will remove the 'cloudify_user_name' check after we will write that field automaticaly on install

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -292,23 +275,36 @@ def get_cluster_status():
...
 
-def _verify_report_newer_than_current(report_time, status_path):
+def _are_keys_in_dict(dictionary, keys):
+    return all(key in dictionary for key in keys)
+
+
+def _verify_status_report_schema(node_id, report):
+    if (_are_keys_in_dict(report['report'], ['status', 'services'])
+        and report['report']['status'] in [ServiceStatus.HEALTHY,
+                                           ServiceStatus.FAIL]):
```
inbala: Wrong logic, it is valid if this 2 conditions are true

PR #2019, opened 2019-12-04
CY-1717 Get the private_ip from the reporter config

inbala:
Also, remove get_cloudify_config function.

### DIFF
--- a/status-reporter/status_reporter/postgresql_reporter.py
+++ b/status-reporter/status_reporter/postgresql_reporter.py
@@ -19,7 +19,8 @@ from requests.exceptions import RequestException
 from cloudify.cluster_status import CloudifyNodeType, NodeServiceStatus
 
 from .status_reporter import Reporter, logger
-from .utils import get_systemd_services, get_cloudify_config, get_node_status
+from .constants import STATUS_REPORTER_CONFIG_KEY
+from .utils import get_systemd_services, get_node_status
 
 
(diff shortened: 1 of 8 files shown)


status-reporter/status_reporter/postgresql_reporter.py
```diff
@@ -45,11 +45,12 @@ def _collect_status(self):
         return status, services
 
     def _get_patroni_status(self):
-        config = get_cloudify_config(logger)
-        if not config:
+        extra_config = self._config.get('extra_config', {})
```
tyacbovi: There is not const for 'extra_config'?

status-reporter/status_reporter/status_reporter.py
```diff
@@ -152,8 +151,8 @@ def _collect_status(self):
 
     def _report(self):
         status, services = self._collect_status()
-        if not isinstance(status, dict):
-            logger.error('Ignoring status: expected a report in dict format,'
+        if not isinstance(services, dict):
```
tyacbovi: in Yoni's PR there is the range validation for status?

inbala: Yes Yoni will update it in his PR

status-reporter/status_reporter/postgresql_reporter.py
inbala: There is now :)
There was in the manager-install repo

PR #2027, opened 2019-12-09
CY-2068 Add Postgres cluster-status logic

inbala:
* Every node reports it's own status and we want to check
  the status of the DB cluster.
* We make sure the master has at least one sync replication
  with streaming status (or the cluster is in Fail).
* If not all the replications are in streaming status the cluster
  status is degraded.

### DIFF
--- a/rest-service/manager_rest/cluster_status_manager.py
+++ b/rest-service/manager_rest/cluster_status_manager.py
@@ -29,8 +29,12 @@ from manager_rest.storage import models, get_storage_manager
 from manager_rest.rest.rest_utils import parse_datetime_string
 
 
+STATUS = 'status'
+SERVICES = 'services'
+EXTRA_INFO = 'extra_info'
 DB_SERVICE_KEY = 'PostgreSQL'
 BROKER_SERVICE_KEY = 'RabbitMQ'
+PATRONI_SERVICE_KEY = 'Patroni'
 UNINITIALIZED_STATUS = 'Uninitialized'
 CLUSTER_STATUS_PATH = '/opt/manager/cluster_statuses'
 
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -258,6 +259,55 @@ def _handle_missing_non_manager_report(service_type, service_key, managers,
...
+    if not sync_replica:
+        return ServiceStatus.FAIL
+
+    if not all_replicas_streaming:
+        return ServiceStatus.DEGRADED
+
+    return db_service['status']
+
+
+def _update_db_cluster_status(cluster_status):
```
tyacbovi: I think you should return the value and not update that there

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -258,6 +259,55 @@ def _handle_missing_non_manager_report(service_type, service_key, managers,
...
+    db_service = cluster_status['services'][CloudifyNodeType.DB]
+    if db_service['status'] == ServiceStatus.FAIL or db_service['is_external']:
```
tyacbovi: const 'status'

status-reporter/status_reporter/postgresql_reporter.py
```diff
@@ -55,16 +55,20 @@ def _get_patroni_status(self):
         if not detailed_status:
             return NodeServiceStatus.INACTIVE, {}
 
-        if detailed_status['state'] != 'running':
-            return NodeServiceStatus.INACTIVE, detailed_status
+        replication = detailed_status.get('replication', [])
+        replication = [{'sync_state': rep['sync_state'],
```
tyacbovi: 'replication' seems to contain data about replications, so I think it needs a better name

PR #2031, opened 2019-12-11
CY-2065 add rabbitmq cluster status logic

YoniItzhak:


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -258,6 +258,51 @@ def _handle_missing_non_manager_report(service_type, service_key, managers,
...
+    broker_nodes = broker_service['nodes']
+    active_broker_nodes = {name: node for name, node in broker_nodes.items()
+                           if node['status'] == ServiceStatus.HEALTHY}
+    if not _is_broker_cluster_status_the_same(active_broker_nodes):
+        return ServiceStatus.FAIL
+
+    return broker_service['status']
+
+
+def _is_broker_cluster_status_the_same(broker_nodes):
```
inbala: Maybe call it `_verify_broker_cluster_status` or something similar?

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -258,6 +258,51 @@ def _handle_missing_non_manager_report(service_type, service_key, managers,
                                             'services': node_service})
     service['status'] = _get_service_status(service['nodes'])
 
+
+def _update_broker_cluster_status(cluster_status):
```
inbala: As we said, we should create one function for the db and broker

rest-service/manager_rest/cluster_status_manager.py
inbala: Also maybe document this function

status-reporter/status_reporter/rabbitmq_reporter.py
```diff
@@ -25,7 +25,7 @@
 from .utils import get_systemd_services, determine_node_status
 
 NODES_API = 'nodes'
-VHOSTS_API = 'vhosts'
+OVERVIEW_API = 'overview'
```
tyacbovi: You don't have to const everything, if you use it in one place you don't need this

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -303,6 +304,56 @@ def _get_db_cluster_status(db_service, expected_nodes_number):
     return db_service[STATUS]
 
 
+def _should_not_validate_cluster(service):
+    return service[STATUS] == ServiceStatus.FAIL or service[IS_EXTERNAL]
+
+
+def _get_broker_cluster_status(broker_service, expected_nodes_number):
+    """
+    Get the status of the broker cluster. The majority of nodes should be
```
tyacbovi: I will not document in comments, because this could change without updating these docs and the code should be clear enough

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -303,6 +304,56 @@ def _get_db_cluster_status(db_service, expected_nodes_number):
...
+        broker_service = node[SERVICES][BROKER_SERVICE_KEY]
+        cluster_status = broker_service[EXTRA_INFO]['cluster_status']
+        cluster_nodes_status = cluster_status['cluster_nodes_status']
+        node_name = cluster_status['node_name']
+        if not first_node_cluster_status:
+            first_node_cluster_status = cluster_nodes_status
+            first_node_name = node_name
+            continue
+        if cluster_nodes_status != first_node_cluster_status:
+            current_app.logger.info('{0} is not in the same cluster as '
```
tyacbovi: error lvl

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -303,6 +304,56 @@ def _get_db_cluster_status(db_service, expected_nodes_number):
...
+                                    '{1}'.format(first_node_name, node_name))
+            return False
```
tyacbovi: let's run over all nodes, and return the failure at the end with the full report on what went wrong

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -303,6 +304,56 @@ def _get_db_cluster_status(db_service, expected_nodes_number):
...
+    """
+    first_node_name = ''
+    first_node_cluster_status = {}
+
+    for node in active_broker_nodes.values():
+        broker_service = node[SERVICES][BROKER_SERVICE_KEY]
+        cluster_status = broker_service[EXTRA_INFO]['cluster_status']
+        cluster_nodes_status = cluster_status['cluster_nodes_status']
+        node_name = cluster_status['node_name']
+        if not first_node_cluster_status:
```
tyacbovi: why not compare with the cluster structure from the DB?

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -303,6 +304,56 @@ def _get_db_cluster_status(db_service, expected_nodes_number):
...
+        return broker_service[STATUS]
+
+    broker_nodes = broker_service['nodes']
+    active_broker_nodes = {name: node for name, node in broker_nodes.items()
+                           if node[STATUS] == ServiceStatus.HEALTHY}
+
+    if not _verify_broker_cluster_status(active_broker_nodes):
+        return ServiceStatus.FAIL
+
+    if expected_nodes_number != len(active_broker_nodes):
```
tyacbovi: let's add logging to every failure point

tyacbovi: add a comment here that this should only happen on broker setup

rest-service/manager_rest/cluster_status_manager.py
YoniItzhak: Because we want to check the cluster status of the currently active nodes.

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -303,6 +308,59 @@ def _get_db_cluster_status(db_service, expected_nodes_number):
...
+        current_app.logger.error(BROKER_NODES_DIFFERENT_CLUSTER_STATUS.
+                                 format(node_name_a=node_name_a,
+                                        cluster_status_a=cluster_status_a,
+                                        node_name_b=node_name_b,
+                                        cluster_status_b=cluster_status_b))
+        return False
+    return True
+
+
+def _verify_identical_broker_cluster_status(active_nodes):
```
tyacbovi: well this is a bit overkill, you just need one source of truth and compare to it

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -303,6 +303,69 @@ def _get_db_cluster_status(db_service, expected_nodes_number):
...
+                             'recognizes the cluster {cluster_status_b}.'.
+                             format(node_name_a=node_name_a,
+                                    cluster_status_a=cluster_status_a,
+                                    node_name_b=node_name_b,
+                                    cluster_status_b=cluster_status_b))
+
+
+def _verify_identical_broker_cluster_status(active_nodes):
+    are_cluster_statuses_identical = True
+    first_node_name = ''
```
tyacbovi: you can get the first by poping the active_nodes

PR #2061, opened 2019-12-26
CY-2131: better handling of CA cert

tyacbovi:


status-reporter/status_reporter/status_reporter.py
```diff
@@ -59,6 +59,9 @@ class InitializationError(Exception):
     pass
 
 
+CA_DEFAULT_PATH = '/etc/cloudify/status_reporter_cert.pem'
```
inbala: Why not in /etc/cloudify/ssl as the rest of the certificates?

PR #2062, opened 2019-12-29
CY-2139 Fix cluster-status when one manager is down

inbala:
* When one manager is down, there is no updated status report,
  so there is no info on it's services.

* Added defaults for this case when one manager is down and
  there is no info on the rabbit or db services.


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -233,10 +233,13 @@ def _handle_missing_non_manager_report(service_type,
     service = cluster_services[service_type]
     for missing_node in missing_nodes:
         node_status = ServiceStatus.FAIL
+        default_status = NodeServiceStatus.INACTIVE
```
tyacbovi: May do we need this var 'default_status'?

PR #2063, opened 2019-12-30
CY-2150 Don't update execution token

tehasdf:
It's not stored, so no need to update it

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -550,12 +550,6 @@ class ResourceManager(object):
                     continue
 
                 update_operation = False
-                kwargs = operation.parameters.get('task_kwargs', {}) \
-                    .get('kwargs')
-                if kwargs:
-                    kwargs['__cloudify_context']['execution_token'] = \
-                        execution_token
-                    update_operation = True
 
                 if operation.state in from_states:
                     operation.state = cloudify_tasks.TASK_PENDING
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -549,14 +549,9 @@ class ResourceManager(object):
                 if operation.id in retried_operations:
                     continue
 
-                update_operation = False
-
                 if operation.state in from_states:
                     operation.state = cloudify_tasks.TASK_PENDING
                     operation.parameters['current_retries'] = 0
-                    update_operation = True
-
-                if update_operation:
                     self.sm.update(operation,
                                    modified_attrs=('parameters', 'state'))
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -550,12 +550,6 @@ def _reset_operations(self, execution, execution_token, from_states=None):
                     continue
 
                 update_operation = False
```
tyacbovi: We can change it back to how this code was before the last commit here (update_operation is always false now)

tests/integration_tests/tests/agentless_tests/test_resume.py
```diff
@@ -335,6 +335,7 @@ def test_force_resume_cancelled_nonresumable(self):
         self.assertEqual(execution.status, 'cancelled')
 
         self._unlock_operation('interface1.op_nonresumable')
+        time.sleep(5)
```
tyacbovi: well it's just for we will not see the unauthorized error in the log?

rest-service/manager_rest/resource_manager.py
inbala: `update_operation` can be removed and the `self.sm.update` can be back under the `if operation.state in from_states`

rest-service/manager_rest/resource_manager.py
```diff
@@ -550,12 +550,6 @@ def _reset_operations(self, execution, execution_token, from_states=None):
...
-                kwargs = operation.parameters.get('task_kwargs', {}) \
```
inbala: Also, `update_operation` can be removed and the `self.sm.update` can be back under the `if operation.state in from_states` like it was before my change.

tests/integration_tests/tests/agentless_tests/test_resume.py
inbala: I think it happens in more tests :/

rest-service/manager_rest/resource_manager.py
tehasdf: it's not always false (also set by the other thin) but you're right, changed :P

tests/integration_tests/tests/agentless_tests/test_resume.py
tehasdf: pretty much

we note that also one test already had a sleep after unlock_operation, and tbh every one should do that, so unlock_operation itself should do the sleep

tehasdf: yep, every test that uses _unlock_operation, so i put the sleep there

(one test already had it)

PR #2065, opened 2019-12-31
CY-2153 Add a request timeout when creating a client

inbala:


status-reporter/status_reporter/status_reporter.py
```diff
@@ -99,6 +99,7 @@ def __init__(self, node_type):
             self._ca_cert_valid = False
 
         self._current_reporting_freq = self._config.get('reporting_freq')
+        self._request_timeout = self._config.get('request_timeout', 1)
```
tyacbovi: I will set the 1 to the conf as default

PR #2067, opened 2020-01-01
CY-2155 Get DB nodes list

YoniItzhak:


rest-service/manager_rest/rest/resources_v3_1/manager.py
```diff
@@ -175,3 +175,12 @@ def get(self, pagination=None):
                 broker.username = None
                 broker.password = None
         return brokers
+
+
+class DBNodes(SecuredResource):
+    @marshal_with(models.DBNodes)
+    @paginate
+    @authorize('db_node_get')
```
tyacbovi: db_node s _get

PR #2068, opened 2020-01-03
adding capabilities of Component on the node's runtime props

tyacbovi:


cloudify_types/cloudify_types/component/tests/test_execute.py
```diff
@@ -15,8 +15,11 @@
 import mock
 
 from cloudify.exceptions import NonRecoverableError
+from cloudify_rest_client.exceptions import CloudifyClientError
 
 from ..operations import execute_start
+from ..constants import CAPABILITIES
```
inbala: Should be first

PR #2073, opened 2020-01-05
CY-2154 Add last_see column to the managers table

inbala:
Also, fix the migration when running upgrade node_id cannot be null.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/cluster_status.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/cluster_status.py
@@ -20,9 +20,10 @@ from flask_restful_swagger import swagger
 from cloudify.cluster_status import CloudifyNodeType
 
 from manager_rest.rest import responses
-from manager_rest.storage import models
+from manager_rest.utils import get_formatted_timestamp
 from manager_rest.security.authorization import authorize
 from manager_rest.rest.rest_decorators import marshal_with
+from manager_rest.storage import models, get_storage_manager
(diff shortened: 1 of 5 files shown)


rest-service/manager_rest/rest/resources_v3_1/cluster_status.py
```diff
@@ -73,6 +74,17 @@ def put(self, node_id):
...
                            models.Manager,
                            CloudifyNodeType.MANAGER)
+        storage_manager = get_storage_manager()
+        managers = storage_manager.list(models.Manager,
+                                        filters={'node_id': node_id})
+        if not managers:
+            return
+
+        # Update the manager's last_seen
+        manager = managers[0]
```
tyacbovi: Yoni added something for this, no? getting one value from a filter

rest-service/manager_rest/storage/management_models.py
```diff
@@ -628,6 +628,8 @@ class Manager(_WithCACert, SQLModelBase, CloudifyNodeMixin):
     fs_sync_node_id = db.Column(db.Text, nullable=True)
     networks = db.Column(JSONString)
     node_id = db.Column(db.Text, unique=True, nullable=False)
+    last_seen = db.Column(UTCDateTime, nullable=False, index=True)
+    status_report_frequency = db.Column(db.Integer, nullable=False, default=5)
```
tyacbovi: I would not assign a value here, because it's a must value

PR #2081, opened 2020-01-13
CY-2194: fixing status reporters users to a range

tyacbovi:
- For supporting snap res from version without those users


workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -34,6 +37,12 @@
 _STATUS_REPORTERS_QUERY_TUPLE = ', '.join(
     "'{0}'".format(reporter) for reporter in STATUS_REPORTER_USERS)
 
+STATUS_REPORTERS_ROLES = {
+    MANAGER_STATUS_REPORTER_ID: 7,
```
tehasdf: that 7, 8, 9 has constants, no?

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -205,25 +214,57 @@ def _get_status_reporters_update_query(self):
...
               '{0}',
               '{1}',
-              '{2}'
+              '{2}',
+              TRUE,
+              {3}
            )
         ON CONFLICT (username) DO
-        UPDATE SET password='{1}', api_token_key='{2}'
+        UPDATE SET password='{1}', api_token_key='{2}', active=TRUE, id={3}
```
inbala: Dose it work? the `TRUE`? I think it is usually `t`

tyacbovi: It works 👍

inbala: nice :)

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -200,30 +203,76 @@ def _get_admin_user_update_query(self):
         return (base_query.format(username, password),
                 base_query.format('*'*8, '*'*8))
 
+    @staticmethod
+    def _find_reporter_role(roles_mapping, reporter_id):
+        reporter_role = [role_id for user_id, role_id in roles_mapping
+                         if user_id == reporter_id]
+        if reporter_role:
+            return role_id
```
tehasdf: let's not do hacks like this (also it won't work in py3)

`roles_mapping` isn't really a mapping right now, so let's make it a mapping by doing `return dict(response['all'])` in `_get_status_reporters_roles`. Then this method will be just `try: return roles_mapping[reporter_id]; except KeyError: ...`

PR #2084, opened 2020-01-15
CY-1959 Operation state update: allow only specific transitions

tehasdf:
Only allow specific state transitions.
This is so that going back on an operation state is disallowed
(eg. can't move operation from "succeeded" to "pending"), and also
that updating an operation to be the same state that it already was,
is a 409 as well.

The latter will be thrown if we try to change an operation from
"started" to "started" again, in a resume, which will allow the
executor to understand that it's running a resume.


tehasdf:
obsoleted by #2088

PR #2086, opened 2020-01-15
CY-2208 Check syncthing status before updating the status report

inbala:


rest-service/manager_rest/rest/resources_v3_1/cluster_status.py
```diff
@@ -71,11 +71,18 @@ def get(self):
...
                            models.Manager,
                            CloudifyNodeType.MANAGER)
 
-        # Update the manager's last_seen
+    @staticmethod
+    def _update_manager_last_seen(node_id):
+        report = request.json.get('report', {})
+
+        # Update only if the manager is healthy
+        if report.get('status') != ServiceStatus.HEALTHY:
```
tyacbovi: The logging we talked about needs to be added

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -455,12 +540,24 @@ def _save_report(report_path, report_dict):
...
         json.dump(report_dict, report_file)
 
+
+def _verify_syncthing_status():
+    if not (ha_utils and ha_utils.is_clustered()):
+        return
+    syncthing_status, _ = get_syncthing_status()
+    if syncthing_status == NodeServiceStatus.INACTIVE:
+        raise manager_exceptions.FileSyncServiceError(
+            "Can't update the status report when Syncthing is not healthy"
```
tyacbovi: Well syncthing is the way we create the communication line between the managers, but the issue is that this manager is not healthy member of the cluster

tyacbovi: So not sure about the name

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -39,6 +47,83 @@
...
+    for device_id, stats in device_stats.items():
+        last_seen = parse_datetime_string(stats['lastSeen'])
+
+        # Syncthing is valid when at least one device was seen recently
+        if last_seen > min_last_seen:
+            return True
+    return False
+
+
+def _is_syncthing_valid(syncthing_config, device_stats):
```
tyacbovi: please add logging also to here

tyacbovi: _is_syncthing_valid => _is_manager_cluster_structure_valid

Or something like that, because syncthing is the way we check the manager cluster structure

inbala: I disagree, the intention of the check is to validate syncthing is working properly and sync the file system of all the active managers.

rest-service/manager_rest/cluster_status_manager.py
inbala: You are not sure about what name? The exception name or the message?

tyacbovi: FileSyncServiceError name

PR #2091, opened 2020-01-20
CY-2213: changing the snapshot restore for status reporters users

tyacbovi:
A problem with the current impl occurred because the handling of the restore was in mid db migration

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/postgres.py
+++ b/workflows/cloudify_system_workflows/snapshots/postgres.py
@@ -315,7 +315,7 @@ class Postgres(object):
         # at the moment, so admin must be the first user (ID=0)
         query = (
             'select row_to_json(row) from ('
-            'select * from users where id in {0}'
+            'select * from users where id in ({0})'
             ') row;'.format(reporter_ids)
         )
         command.extend([
@@ -340,7 +340,7 @@ class Postgres(object):
         # at the moment, so admin must be the first user (ID=0)
         query = (
             'select row_to_json(row) from ('
-            'select * from users_roles where user_id in {0}'
+            'select * from users_roles where user_id in ({0})'
             ') row;'.format(reporter_ids)
         )
         command.extend([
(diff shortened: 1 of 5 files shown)


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -566,6 +571,23 @@ def _restore_db(self, postgres, schema_revision, stage_revision):
         # user depending on whether we have the hash salt
         return admin_user_update_command
 
+    @staticmethod
+    def _restore_status_reporters(postgres,
```
inbala: Why this function and _dump_status_reporters are not in the postgres.py?

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -313,6 +314,73 @@ def restore_config_tables(self, config_path):
...
+        run_shell(command)
+        ctx.logger.debug('Dumped status reporter roles to {}'.format(path))
+        return path
+
+    @staticmethod
+    def _restore_json_dump_file(dump_path):
+        with open(dump_path) as dump_file:
+            return json.load(dump_file)
+
+    def restore_status_reporter_users(self, users_dump_path):
```
inbala: I think restore_status_reporter_users and restore_status_reporter_roles functions are redundant and instead we can just call _restore_json_dump_file with the appropriate path

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -313,6 +314,73 @@ def restore_config_tables(self, config_path):
...
+                                 status_reporter_users_path):
+        users = postgres.restore_status_reporter_users(
+            status_reporter_users_path)
+        roles = postgres.restore_status_reporter_roles(
+            status_reporter_roles_path)
+        postgres.upsert_status_reporters_users(users, roles)
+
+    @staticmethod
+    def dump_status_reporters(postgres, tempdir):
+        status_reporter_users_path = postgres.dump_status_reporter_users(
```
inbala: Should be `self` instead of `postgres`

PR #2094, opened 2020-01-23
CY-2255 Remove all integration tests related to 4.2

glukhman:


tests/setup.py
```diff
@@ -17,7 +17,7 @@
 
 setup(
     name='cloudify-integration-tests',
-    version='4.2',
+    version='4.3.1',
```
tyacbovi: Not relevant, don't change this

workflows/cloudify_system_workflows/snapshots/constants.py
```diff
@@ -59,9 +59,6 @@
 SECURITY_FILE_LOCATION = join('/opt/manager/', SECURITY_FILENAME)
 
 
-V_4_0_0 = ManagerVersion('4.0.0')
```
tyacbovi: Not related to the tests, don't change

PR #2099, opened 2020-01-27
CY-2269 Make status reporter keep syncing

geokala:
Without this, an interruption to the ability to write the files seems to
upset syncthing.


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -540,6 +541,7 @@ def _create_statues_folder_if_needed():
 
 
 def _save_report(report_path, report_dict):
+    subprocess.check_call(['rm', '-f', report_path])
```
tehasdf: `os.unlink`? :P

geokala:
POC not needed currently, proper solution under discussion.

PR #2102, opened 2020-02-05
update docker py version

EarthmanT:
update to look for wagon before building wagon. That way we can test specific wagon without building. shortens tests.

### DIFF
--- a/tests/integration_tests/framework/wagon_build.py
+++ b/tests/integration_tests/framework/wagon_build.py
@@ -126,7 +126,7 @@ class WagonBuilderMixin(DockerInterface):
             detach=True)
         response = container.wait(timeout=self.wagon_build_time_limit)
         for msg in container.logs(stream=True):
-            logger.info(msg)
+            logger.info('Wagon build container log: {0}'.format(msg))
         if response['StatusCode'] != 0:
             raise WagonBuildError(response)
         container.stop()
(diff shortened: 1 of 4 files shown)


tests/integration_tests/tests/test_cases.py
```diff
@@ -788,11 +788,14 @@ def tearDown(self):
         self.examples.cleanup()
         super(PluginsTest, self).tearDown()
 
-    @staticmethod
-    def get_wagon_path(plugin_path):
+    # @staticmethod
+    def get_wagon_path(self, plugin_path):
```
tyacbovi: it returns a list now, why that can happen?

tests/integration_tests/framework/wagon_build.py
```diff
@@ -117,13 +118,15 @@ def prepare_docker_image(self):
...
 
-    def build_wagon(self):
+    def build_wagon(self, logger):
         self.prepare_docker_image()
         container = self.run_container(
             self.docker_image_name_with_repo,
             volumes=self.docker_volume_mapping,
             detach=True)
         response = container.wait(timeout=self.wagon_build_time_limit)
+        for msg in container.logs(stream=True):
```
tyacbovi: maybe add some message before and after for readability

tests/integration_tests/tests/test_cases.py
EarthmanT: We have wagons for more than one platform (`Centos`, `RedHat`).

PR #2110, opened 2020-02-10
restservice: pin virtualenv to 15.1

tehasdf:
This is the same version we use in cloudify-agent


tehasdf:
maybe revert after we release wagon 0.6.4

mabuaisha:
@tehasdf I think for `0.6.4` we should make it virtualenv < 20.0.0 as the changes happened on 20.0.0

tehasdf:
this change is only relevant for builds, which is green in CI; the tests are also red on master, so I'll ignore that here

PR #2120, opened 2020-03-02
CY-2366 Update cloudify context with host ip

mabuaisha:


rest-service/manager_rest/workflow_executor.py
```diff
@@ -200,6 +199,10 @@ def _send_task_to_dlx(message, message_ttl, routing_key='workflow'):
 
 def _execute_task(execution_id, execution_parameters,
                   context, execution_creator, scheduled_time=None):
+    # Get the host ip info and return them
+    sm = get_storage_manager()
+    managers = sm.list(models.Manager)
+    context['rest_host'] = [manager.private_ip for manager in managers]
```
tehasdf: nobody really knows if this should be .private_ip or .networks['default']

mabuaisha: Yeah, in some places there is a reference for `networks['default']`

PR #2121, opened 2020-03-03
CY-2363 Add a "validate blueprint" REST endpoint

mateumann:
Add a common blueprint validation endpoint in the restservice so that the composer and the CLI can call to that (instead of using their own code for validating blueprints).


rest-service/dev-requirements.txt
```diff
@@ -1,4 +1,6 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==master
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-2363-add-bluepoint-validate#egg=cloudify-common==CY-2363-add-bluepoint-validate
 
 # For dealing with the binary leftovers of psycopg2 in the 2.7.x version
-psycopg2==2.7.4 --no-binary psycopg2
\ No newline at end of file
+psycopg2==2.7.4 --no-binary psycopg2
+
```
geokala: Why the extra newlines?

rest-service/manager_rest/test/endpoints/test_blueprints.py
```diff
@@ -465,3 +465,46 @@ def test_not_listing_hidden_blueprints(self):
...
+          client_max_version=base_test.LATEST_API_VERSION)
+    def test_blueprint_validate_invalid_id(self):
+        blueprint_id = 'invalid blueprint id'
+        blueprint_file = 'blueprint.yaml'
+        blueprint_path = os.path.join(
+            self.get_blueprint_path('mock_blueprint'),
+            blueprint_file)
+        with self.assertRaises(exceptions.CloudifyClientError) as context:
+            self.client.blueprints.validate(blueprint_path, blueprint_id)
+        # import pdb; pdb.set_trace()
```
tehasdf: remove

rest-service/manager_rest/test/endpoints/test_blueprints.py
```diff
@@ -465,3 +465,46 @@ def test_not_listing_hidden_blueprints(self):
...
+    def test_blueprint_validate_invalid_blueprint(self):
+        blueprint_id = 'blueprint_main_file_name'
+        blueprint_file = 'invalid_blueprint.yaml'
+        blueprint_path = os.path.join(
+            self.get_blueprint_path('mock_blueprint'),
+            blueprint_file)
+        self.assertEquals(0, len(self.client.blueprints.list()))
+        with self.assertRaises(exceptions.CloudifyClientError) as context:
+            self.client.blueprints.validate(blueprint_path, blueprint_id)
+        # import pdb; pdb.set_trace()
```
tehasdf: remove

rest-service/manager_rest/test/endpoints/test_blueprints.py
geokala: Ditto

rest-service/manager_rest/test/endpoints/test_blueprints.py
geokala: Please remove the commented code

geokala: Side note: if you run pytest with --pdb it'll give you a pdb on failure.

rest-service/manager_rest/upload_manager.py
```diff
@@ -606,6 +606,56 @@ def _extract_application_file(cls,
...
+                    resource_target_path,
+                    self._get_data_url_key(),
+                    self._get_kind())
+            self._prepare_and_process_doc(
+                data_id,
+                file_server_root,
+                resource_target_path,
+                additional_inputs=additional_inputs,
+                **kwargs)
+            return {}, 204
```
tehasdf: should probably be something like `return "", 204`, and maybe no `@marshal_with`? i think like this, it will actually return a response, but with 204, there is supposed to be actually no response body at all

rest-service/manager_rest/upload_manager.py
```diff
@@ -606,6 +606,56 @@ def _extract_application_file(cls,
...
+                app_dir,
+                app_file_name,
+                file_server_root
+            )
+        except manager_exceptions.DslParseException, ex:
+            raise manager_exceptions.InvalidBlueprintError(
+                'Invalid blueprint - {0}'.format(ex.message))
+        finally:
+            # remove uploaded data that has been validated
+            shutil.rmtree(os.path.join(file_server_root, app_dir))
```
tehasdf: maybe `remove()` as well (same as you use in the finally in the other method)

rest-service/manager_rest/upload_manager.py
```diff
@@ -606,6 +606,56 @@ def _extract_application_file(cls,
...
+            file_server_root, app_dir, args.application_file_name)
+
+        # add to blueprints manager (will also dsl_parse it)
+        try:
+            get_resource_manager().validate_blueprint(
+                app_dir,
+                app_file_name,
+                file_server_root
+            )
+        except manager_exceptions.DslParseException, ex:
```
tehasdf: this syntax is deprecated in 2.7 and removed in 3, please get used to writing this instead: `except manager_exceptions.DslParseException as ex:` (`as` instead of comma)

i realize there's other cases in this file of using this syntax, and there's no rush to change this now, because an upcoming py3 translation for cloudify-manager will handle this case automatically, just keep in mind that in the future this syntax won't work

rest-service/dev-requirements.txt
```diff
@@ -1,4 +1,5 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common==master
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-2363-add-bluepoint-validate#egg=cloudify-common==CY-2363-add-bluepoint-validate
```
tehasdf: remember to revert

PR #2124, opened 2020-03-11
CY-2435 Restore scheduled tasks from snapshot

mateumann:
Properly restore scheduled executions from snapshots.

Set status of overdue executions as FAILED, requeue executions that are
due in the future.


rest-service/manager_rest/resource_manager.py
```diff
@@ -621,6 +622,39 @@ def resume_execution(self, execution_id, force=False):
 
         return execution
 
+    def requeue_execution(self, execution_id, force=False):
+        execution = self.sm.get(models.Execution, execution_id)
+        execution_token = generate_execution_token(execution_id)
+        if execution.status != ExecutionState.SCHEDULED:
```
geokala: Shouldn't this happen before generating the token?

tests/integration_tests/tests/agentless_tests/test_snapshot.py
```diff
@@ -494,3 +494,37 @@ def _restore_security_config(self):
...
+
+        failed_executions = self.client.executions.list(
+            status=Execution.FAILED
+        )
+        self.assertEqual(len(failed_executions), 2)
+
+        executions = self.client.executions.list(status=Execution.SCHEDULED)
+        self.assertEqual(len(executions), 2)
+
+        for execution in filter(lambda e: e.workflow_id == 'install',
```
tehasdf: plz no filter-lambda, that's ugly, slow, and will be replaced by futurize in py3 porting anyway

instead of `filter(lambda x: x == 'install', somelist)`, do `[x for x in somelist if x == 'install']`

in this case, actually best do:
```python
for execution in executions:
    if execution.workflow_id == 'install':
       ...
    if execution.workflow_id == 'uninstall':
       ...
```

workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -449,3 +452,30 @@ def composer_db_schema_get_current_revision():
...
+            ' valid date. \nExpected format: YYYYMMDDHHMM+HHMM or'
+            ' YYYYMMDDHHMM-HHMM i.e: 201801012230-0500'
+            ' (Jan-01-18 10:30pm EST)'.format(datetime_str))
+    # Date is in UTC, tzinfo is not necessary
+    return utc_date.replace(tzinfo=None)
+
+
+def is_later_than_now(datetime_str):
+    datetime_ts = parse_datetime_string(datetime_str)
+    return datetime.datetime.now(tz=None) < datetime_ts
```
tehasdf: utcnow?

PR #2136, opened 2020-03-30
CY-2521 Py3 porting: stage2 changes

tehasdf:
Stage 2 changes, including fix_dict but no unicode changes yet


rest-service/manager_rest/deployment_update/handlers.py
```diff
@@ -595,8 +595,8 @@ def handle(self, dep_update, updated_instances):
         """
         # create node instance relationship ordering
         modified_instances = {k: {}
-                              for k, _ in self._handlers_mapper.iteritems()}
-        for change_type, handler in self._handlers_mapper.iteritems():
+                              for k, _ in self._handlers_mapper.items()}
```
mateumann: could you use `.keys()` instead?

PR #2147, opened 2020-04-07
CY-2592 Add unique indexes to SQL database

mateumann:
So far there were hardly any constraints for plugins stored in database.
All validation was handled in Python code.

This patch adds a constraint (and an index as well) for `package_name`
and `package_version` in `plugins` table to be unique.


tehasdf:
good idea
note that any db change also requires a change in migrations to actually be able to create a db schema with this

PR #2155, opened 2020-04-13
CY-2488 Fetch rest host from manager networks

mabuaisha:


mabuaisha:
@tehasdf will close this once finish testing for sure

mabuaisha:
This PR will be closed as its needed and was created for sake of testing and generating new rpm and also it truns out that we do not need to pass all ips networks to the rest host as we only now pass the relevant ips for agent to connect to the manager

PR #2160, opened 2020-04-16
CY-2609 verify_and_convert_bool: use bool literals

tehasdf:
verify_and_convert_bool has a isinstance(.., text_type) - while
data coming in from the request is indeed text (ie. unicode),
literals in our code aren't necessarily (bare strings in py2, being
bytes).

Instead, just pass bool literals. No reason to pass strings,
be it text or bytes.


tehasdf:
fwiw `verify_and_convert_bool` is at https://github.com/cloudify-cosmo/cloudify-manager/blob/eed9baf6ae906f4ba82b6112ae4bbd85c8fd112e/rest-service/manager_rest/rest/rest_utils.py#L89

PR #2166, opened 2020-04-20
CY-2632 Run restservice on py3

tehasdf:
Use python3 to make the restservice virtualenv, so it'll use that
python to actually run it.

This also makes amqp-postgres run py3, because it's using the
same virtualenv as restservice


tehasdf:
fwiw a cloudify-manager-install CI run using this branch is https://circleci.com/workflow-run/ac5f3925-843d-49c5-90eb-1973f1a720ea

PR #2168, opened 2020-04-21
Cy 2617 component exec cancel bug

glukhman:


rest-service/manager_rest/resource_manager.py
```diff
@@ -1168,11 +1168,13 @@ def cancel_execution(self, execution_id, force=False, kill=False):
         if kill:
             workflow_executor.cancel_execution(execution_id)
 
-        # Dealing with the inner Component-s deployments
+        # Dealing with the inner Component's deployments
```
YoniItzhak, line 1171: I think it should be Components**'** since there can be multiple?

glukhman: right...

PR #2171, opened 2020-04-28
CY-2688 Add test_modify_inputs_ops_order

mateumann:
A test to verify if update and install steps are executed as they should be
during the update workflow.


tests/integration_tests/tests/agentless_tests/deployment_update/test_deployment_update_modification.py
```diff
@@ -388,3 +390,45 @@ def test_modify_description(self):
...
+        execution_ids = [en.id for en in self.client.executions.list(
+            deployment_id=deployment.id,
+            workflow_id='update',
+            status='terminated',
+        )]
+        self.assertEquals(len(execution_ids), 1)
+
+        # verify if inputs have been updated
+        deployment = self.client.deployments.get(dep_update.deployment_id)
+        self.assertEquals(deployment['inputs'], {
```
tehasdf, line 418: plz `assertEqual`, sans `s` - the old "equals" alias logs a warning at runtime :(

PR #2173, opened 2020-05-06
CY-2743 Mgmtworker: don't install stub packages anymore

tehasdf:
No need.


tehasdf:
test_restservice is unrelated to this build, so I'm not going to care about the failure there

PR #2180, opened 2020-05-14
CY-2702 update IDD table during upgrade

YoniItzhak:


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -352,6 +355,62 @@ def _restore_deployment_envs(self, postgres):
...
+                        tenant_client.inter_deployment_dependencies.restore(
+                            deployment_id, update_service_composition)
+                    except RuntimeError as err:
+                        failed_deployments_queue.put((deployment_id, tenant))
+                        ctx.logger.info('Failed creating inter deployment '
+                                        'dependencies for deployment {0} from'
+                                        ' tenant {1}. {2}'
+                                        .format(deployment_id, tenant, err))
+
+        threads = list()
```
tehasdf, line 394:
```suggestion
        threads = []
```
`[]` is just shorter to write

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -352,6 +355,62 @@ def _restore_deployment_envs(self, postgres):
...
+        deployments_queue = queue.Queue()
+        failed_deployments_queue = queue.Queue()
+        for tenant_name, deployments in deployment_contexts:
+            for dep_id in deployments:
+                deployments_queue.put((tenant_name, dep_id))
+
+        wf_context = current_workflow_ctx.get_ctx()
+        context_params = current_workflow_ctx.get_parameters()
+
+        def _create_inter_deployment_dependencies():
```
tehasdf, line 375: i would prefer this function to take both queues as arguments, instead of by closure (but it would also need wf_context and params then)

the reason is that this is going to run as a thread, and with threads, you want to really be sure that you don't have unsynchronized (unlocked) access to random things. If this function didn't take anything by closure, but instead took everything by arguments, then it would even be possible to move it outside, to module level - and then it would be very obvious that it doesn't access anything it shouldn't

accessing things by closure like this, just makes me look at this function very closely to examine if it doesn't touch anything that it isn't supposed to

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -352,6 +355,62 @@ def _restore_deployment_envs(self, postgres):
...
+                                        'dependencies for deployment {0} from'
+                                        ' tenant {1}. {2}'
+                                        .format(deployment_id, tenant, err))
+
+        threads = list()
+        for i in range(min(self._config.snapshot_restore_threads,
+                           deployments_queue.qsize())):
+            t = threading.Thread(target=_create_inter_deployment_dependencies)
+            threads.append(t)
+            t.setDaemon(True)
```
tehasdf, line 399: the `.setDaemon` does mostly tnothing here: daemon threads will not keep the process open, if the main thread exits.
But here, the main threads calls `.join()` on them anyway, so the main thread will not exit as long as those thread live anyway.

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -352,6 +355,62 @@ def _restore_deployment_envs(self, postgres):
...
+            t.start()
+
+        for t in threads:
+            t.join()
+
+        if not failed_deployments_queue.empty():
+            deployments = list(failed_deployments_queue.queue)
```
tehasdf, line 406: accessing `.queue` on a Queue object is not great, because
1. it sidesteps the locks (here, it's easy to see that by this point, there's no concurrent access to the queue, so it should be safe
2. it's not a documented attribute and so should/could be considered private

but in this case, there isn't a very easy way of getting the queue contents. And just for displaying, I'd allow it..

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -352,6 +355,62 @@ def _restore_deployment_envs(self, postgres):
...
+                    try:
+                        tenant_client = get_rest_client(tenant=tenant)
+                        tenant_client.inter_deployment_dependencies.restore(
+                            deployment_id, update_service_composition)
+                    except RuntimeError as err:
+                        failed_deployments_queue.put((deployment_id, tenant))
+                        ctx.logger.info('Failed creating inter deployment '
+                                        'dependencies for deployment {0} from'
+                                        ' tenant {1}. {2}'
+                                        .format(deployment_id, tenant, err))
```
tehasdf, line 392:
```suggestion
                        ctx.logger.info('Failed creating inter deployment '
                                        'dependencies for deployment %s from'
                                        ' tenant %s. %s',
                                        deployment_id, tenant, err)
```

the difference is that `logger.info("xx {0}".format(x))` does the string formatting right there, while `logger.info("xx %s", x)` (note: separate arguments, NOT `"%s" % x`) only does the formatting when the log is emitted. So it won't do it, if the log is not going to be emitted because of logging level, so that might be better for performance

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -352,6 +355,62 @@ def _restore_deployment_envs(self, postgres):
 
         self._log_final_information(deps_with_failed_plugins)
 
+    def _restore_inter_deployment_dependencies(self):
+        if (self._snapshot_version < V_4_6_0 or
+                self._snapshot_version > V_5_0_5):
+            return
```
tehasdf, line 361: this could use a comment as to _why_ this check is here ("because managers older than 4.6.0 didn't support components, and managers newer than 5.0.5 have the dependencies as part of the database dump")

tests/integration_tests/tests/agentless_tests/test_snapshot.py
```diff
@@ -238,11 +244,58 @@ def test_v_5_0_5_restore_snapshot_with_updated_deployment(self):
...
+
+        assert (self._openstack_inter_deployment_dependencies()).issubset(
+            set(inter_deployment_dependencies))
+
+        if not self._assert_component_listed(inter_deployment_dependencies):
+            raise AssertionError('component.infrastructure does not exist '
+                                 'as a dependency creator.')
+
+    @staticmethod
+    def _assert_component_listed(inter_deployment_dependencies):
```
tehasdf, line 265: this one is named `assert` but it doesn't actually assert, just returns a bool

tests/integration_tests/tests/agentless_tests/test_snapshot.py
```diff
@@ -238,11 +244,58 @@ def test_v_5_0_5_restore_snapshot_with_updated_deployment(self):
...
+        for dependency in inter_deployment_dependencies:
+            if dependency[2].startswith('component.infrastructure'):
+                return True
+
+        return False
```
tehasdf, line 270: could be written as `return any(d[2].startswith(..) for d in inter_dep_deps)`

tests/integration_tests/tests/agentless_tests/test_snapshot.py
```diff
@@ -238,11 +244,58 @@ def test_v_5_0_5_restore_snapshot_with_updated_deployment(self):
...
+        inter_deployment_dependencies = postgresql.run_query(
+            "SELECT _source_deployment, _target_deployment,  "
+            "dependency_creator FROM inter_deployment_dependencies;"
+        )['all']
+
+        assert (self._openstack_inter_deployment_dependencies()).issubset(
+            set(inter_deployment_dependencies))
+
+        if not self._assert_component_listed(inter_deployment_dependencies):
+            raise AssertionError('component.infrastructure does not exist '
```
tehasdf, line 261: usually you don't raise AssertionError directly, but there's a few things you can choose from instead:
1. `self.fail('blabla')` (unittest-compatible)
2. `pytest.fail('blabla')` (pytest, of course)
3. `self.assertTrue(self._blabla(),  "fail message here")`
4. `assert self._blabla, "fail message here"`
5. throwing that assertion inside of `self._assert_component_listed`

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -324,3 +333,82 @@ def get(self,
...
+        parsed_deployment = rest_utils.get_parsed_deployment(
+            blueprint, app_dir, app_blueprint)
+        deployment_plan = rest_utils.get_deployment_plan(
+            parsed_deployment, deployment.inputs,  runtime_only_evaluation)
+        rest_utils.update_deployment_dependencies_from_plan(
+            deployment_id, deployment_plan, sm, lambda *_: True)
+        if data.get('update_service_composition'):
+            self._create_service_composition_dependencies(deployment_plan,
+                                                          deployment,
+                                                          sm)
```
tehasdf, line 365: doesn't this function need to return something?

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -324,3 +333,82 @@ def get(self,
                 substr_filters=search
             )
         return inter_deployment_dependencies
+
+
+class InterDeploymentDependenciesRestore(SecuredResource):
+    @authorize('inter_deployment_dependency_create')
+    def post(self):
+        """
```
tehasdf, line 341: PEP 257 is the style document for docstrings (like PEP8 for code), and it prefers the style:
```python
"""Short summary, just one line

Longer text longer text
probably even multiline

:param: maybe some param declarations here
"""
```

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -88,58 +87,30 @@ def stage_deployment_update(self,
...
+        parsed_deployment = get_parsed_deployment(old_blueprint,
+                                                  app_dir,
+                                                  app_blueprint)
 
         # Updating the new inputs with the deployment inputs
         # (overriding old values and adding new ones)
         old_inputs = copy.deepcopy(deployment.inputs)
         new_inputs = {k: old_inputs[k]
-                      for k in plan.inputs.keys() if k in old_inputs}
+                      for k in parsed_deployment.inputs.keys()
```
tehasdf, line 100: i usually prefer to drop `.keys()` from dict iteration calls, because iterating over the dict itself gives you keys, so no real need to create an additional object (and in py2, it'll be a full list of keys - so if the dict was a million entries, you're now creating an additional list of a million items as well)

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
YoniItzhak: Makes a lot of sense, changed it

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
YoniItzhak: Correct, removed

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
YoniItzhak: Thanks, I'll remember this comment for next time

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
YoniItzhak: Nice! Changed it

workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
YoniItzhak: Added

tests/integration_tests/tests/agentless_tests/test_snapshot.py
YoniItzhak: Correct, I'll move the exception from the main function in here

tests/integration_tests/tests/agentless_tests/test_snapshot.py
YoniItzhak: Oh nice, done

tests/integration_tests/tests/agentless_tests/test_snapshot.py
YoniItzhak: Picked 5, and I'll remember it for future cases

rest-service/manager_rest/rest/resources_v3_1/deployments.py
YoniItzhak: Great question, it basically just updates the inter deployment dependencies in the DB, so I figured it doesn't need to return anything.

rest-service/manager_rest/rest/resources_v3_1/deployments.py
YoniItzhak: Nice, made me think also about the parameter 'runtime_only_evaluation' and I realized there is no need in it so I deleted it.

rest-service/manager_rest/deployment_update/manager.py
YoniItzhak: Great call, deleted

PR #2186, opened 2020-05-18
Cybl 1017 scale failure

geokala:


geokala:
CI is sad because it wants only the latest and greatest (and because of recent flake8 changes). It doesn't make sense to apply all the changes CI wants for a patch for a previous version.

PR #2187, opened 2020-05-18
CYBL-1017 Correctly store pre-scale state

geokala:


geokala:
To be merged when related patchify commit (not yet created) is merged.

PR #2197, opened 2020-05-22
snapshots' utils run(): decode stdout/stderr [ci skip]

tehasdf:
Decode on the fringes, in a "sandwich" way.

### DIFF
--- a/workflows/cloudify_system_workflows/snapshots/utils.py
+++ b/workflows/cloudify_system_workflows/snapshots/utils.py
@@ -254,6 +254,10 @@ def run(command, ignore_failures=False, redirect_output_path=None, cwd=None):
         proc = subprocess.Popen(command, stdout=stdout, stderr=stderr, cwd=cwd)
 
     proc.aggr_stdout, proc.aggr_stderr = proc.communicate()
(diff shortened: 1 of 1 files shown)


workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -254,6 +254,10 @@ def run(command, ignore_failures=False, redirect_output_path=None, cwd=None):
         proc = subprocess.Popen(command, stdout=stdout, stderr=stderr, cwd=cwd)
 
     proc.aggr_stdout, proc.aggr_stderr = proc.communicate()
+    if proc.aggr_stdout is not None:
+        proc.aggr_stdout = proc.aggr_stdout.decode('utf-8')
+    if proc.aggr_stderr is not None:
+        proc.aggr_stderr = proc.aggr_stderr.decode('utf-8')
```
tehasdf, line 260: this is the same as we do in other places where we have a `run()` utility, eg. https://github.com/cloudify-cosmo/cloudify-manager-install/blob/cdb9306798100f337cf183edf87ea97a25737fd5/cfy_manager/utils/common.py#L69-L72

PR #2202, opened 2020-05-26
Display error in uninst/stop/update

glukhman:
not finished yet

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -482,14 +482,36 @@ class ResourceManager(object):
         deployment = self.sm.get(models.Deployment, deployment_id)
 
         # Validate there are no running executions for this deployment
-        deplyment_id_filter = self.create_filters_dict(
+        deployment_id_filter = self.create_filters_dict(
             deployment_id=deployment_id,
             status=ExecutionState.ACTIVE_STATES + ExecutionState.QUEUED_STATE
         )
         executions = self.sm.list(
             models.Execution,
-            filters=deplyment_id_filter
+            filters=deployment_id_filter
         )
+
+        # Verify deleting the deployment won't affect dependent deployments
+        dep_graph = RecursiveDeploymentDependencies(self.sm)
+        excluded_id = self._excluded_component_creator_id(deployment)
+        deployment_dependencies = \
+            dep_graph.retrieve_and_display_dependencies(
+                deployment.id,
+                excluded_component_creator_id=excluded_id)
+        if deployment_dependencies:
+            if ignore_live_nodes:
+                current_app.logger.warning(
+                    "Force-deleting deployment {0} despite having the "
+                    "following existing dependent installations\n{1}".format(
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -482,15 +482,37 @@ def delete_deployment(self,
...
         # Validate there are no running executions for this deployment
-        deplyment_id_filter = self.create_filters_dict(
+        deployment_id_filter = self.create_filters_dict(
             deployment_id=deployment_id,
             status=ExecutionState.ACTIVE_STATES + ExecutionState.QUEUED_STATE
         )
         executions = self.sm.list(
             models.Execution,
-            filters=deplyment_id_filter
+            filters=deployment_id_filter
```
YoniItzhak, line 491: Nice catch, can you change it also on line 504 and 508?

rest-service/manager_rest/resource_manager.py
```diff
@@ -482,15 +482,37 @@ def delete_deployment(self,
...
 
+        # Verify deleting the deployment won't affect dependent deployments
+        dep_graph = RecursiveDeploymentDependencies(get_storage_manager())
+        excluded_id = self._excluded_component_creator_id(deployment)
+        deployment_dependencies = \
+            dep_graph.retrieve_and_display_dependencies(
+                deployment.id,
+                excluded_component_creator_id=excluded_id)
+        if deployment_dependencies:
+            if ignore_live_nodes:
```
YoniItzhak, line 502: Maybe we should change the var name: `ignore_live_nodes` to something like `force_delete` or anything else? I guess `ignore_live_nodes` refer to the fact that the deployment itself has live nodes that it deployed.

rest-service/manager_rest/resource_manager.py
```diff
@@ -482,15 +482,37 @@ def delete_deployment(self,
...
         )
         executions = self.sm.list(
             models.Execution,
-            filters=deplyment_id_filter
+            filters=deployment_id_filter
         )
 
+        # Verify deleting the deployment won't affect dependent deployments
+        dep_graph = RecursiveDeploymentDependencies(get_storage_manager())
+        excluded_id = self._excluded_component_creator_id(deployment)
```
YoniItzhak, line 496: What does this line do?

rest-service/manager_rest/resource_manager.py
```diff
@@ -482,15 +482,37 @@ def delete_deployment(self,
...
             status=ExecutionState.ACTIVE_STATES + ExecutionState.QUEUED_STATE
         )
         executions = self.sm.list(
             models.Execution,
-            filters=deplyment_id_filter
+            filters=deployment_id_filter
         )
 
+        # Verify deleting the deployment won't affect dependent deployments
+        dep_graph = RecursiveDeploymentDependencies(get_storage_manager())
```
YoniItzhak, line 495: We can use `self.sm` for the storage_manager?

PR #2207, opened 2020-06-04
CY-2915 Update service status management for supervisord

mabuaisha:


rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -63,14 +63,19 @@
...
-SUPERVISORD_SERVICES = {
-    'AMQP-Postgres': 'cloudify-amqp-postgres',
-    'Cloudify Composer': 'cloudify-composer',
-    'Cloudify Console': 'cloudify-stage',
-    'Management Worker': 'cloudify-mgmtworker',
-    'Webserver': 'cloudify-nginx',
-    'Manager Rest-Service': 'cloudify-restservice',
-    'File Sync Service': 'cloudify-syncthing'
+SUPERVISORD_BASE_SERVICES = {
+    'cloudify-amqp-postgres': 'AMQP-Postgres',
```
geokala, line 67: Aside from the .service extension, these are the same as the above- and I think systemd works without the .service extension so maybe we should remove it and fold them into one?

rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -136,34 +141,44 @@ def get(self):
         return {'status': status, 'services': services}
 
     def _check_supervisord_services(self, services):
+        statuses = []
+        supervisord_services = get_system_manager_services(
+            SUPERVISORD_BASE_SERVICES,
+            SUPERVISORD_OPTIONAL_SERVICES
+        )
+        for name, display_name in supervisord_services.items():
```
geokala, line 149: If we can collapse the service listings together then we can presumably refactor this with the _check_systemd_services since the logic is basically the same?

rest-service/manager_rest/rest/resources_v3_1/status.py
mabuaisha: If we can drop the `.service` extension then yeah we can fold them into one instead of having two extra dicts. Will check and see

mabuaisha: I think this is not going to only drop it from this dict but need also to do changes for the file names as well and for example `nginx.service` is something that we are not configure/handle. the same thing applies for database service name. So maybe we can do it as part of another jira in the future

rest-service/manager_rest/rest/resources_v3_1/status.py
mabuaisha: Based on my comment added above, the change requries to change service file name on the manager side as well as the services that we are not override/configure like nginx/postgresql

rest-service/manager_rest/rest/resources_v3_1/status.py
geokala: Do we use these to check file names then, or just systemctl status?

mabuaisha: @geokala we used them when configure each service by moving each service under `/usr/lib/systemd/system/`

rest-service/manager_rest/rest/resources_v3_1/status.py
mabuaisha: Did not combine the two logic into one method as it will increase complexity somehow as both supervisord and systemd used different interfaces for getting data about the services and some pre call is needed for systemd and not for supervisord.

PR #2212, opened 2020-06-04
CY-2642 Fix component execution resume

glukhman:


rest-service/manager_rest/resource_manager.py
```diff
@@ -639,11 +639,13 @@ def resume_execution(self, execution_id, force=False):
...
         )
 
-        # Dealing with the inner Component-s deployments
+        # Dealing with the inner Components' deployments
         components_executions = self._find_all_components_executions(
             execution.deployment_id)
         for exec_id in components_executions:
-            self.resume_execution(exec_id, force)
+            execution = self.sm.get(models.Execution, exec_id)
+            if execution.status not in ExecutionState.END_STATES:
```
geokala, line 647: This seems to be avoiding resuming in all situations where we might want to resume, since resuming only makes sense in cases of failed or cancelled (?) workflows. This won't allow that to happen, but will allow resuming in all cases where it doesn't make sense (pending, started) if they get this far?

PR #2216, opened 2020-06-09
CY-2957 Cancel cascade deletion in IDD table when uninstalling a comp…

glukhman:
…onent

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/deployments.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/deployments.py
@@ -212,9 +212,11 @@ class InterDeploymentDependencies(SecuredResource):
         return sm.put(deployment_dependency)
 
     @staticmethod
-    def _verify_and_get_source_and_target_deployments(sm,
-                                                      source_deployment_id,
-                                                      target_deployment_id):
+    def _verify_and_get_source_and_target_deployments(
+            sm,
+            source_deployment_id,
+            target_deployment_id,
+            is_component_deletion=False):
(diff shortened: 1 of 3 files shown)


rest-service/dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common[dispatcher]==master
+git+https://github.com/cloudify-cosmo/cloudify-common@CY-2957-Cancel-cascade-delete-IDD-for-component-uninstall#egg=cloudify-common[dispatcher]==CY-2957-Cancel-cascade-delete-IDD-for-component-uninstall
```
YoniItzhak, line 1: Reminder to revert.

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -212,9 +212,9 @@ def put(self):
         return sm.put(deployment_dependency)
 
     @staticmethod
-    def _verify_and_get_source_and_target_deployments(sm,
-                                                      source_deployment_id,
-                                                      target_deployment_id):
+    def _verify_and_get_source_and_target_deployments(
+            sm, source_deployment_id, target_deployment_id,
+            is_component_deletion=False):
```
YoniItzhak, line 217: Just in terms of styling, it would be great if you could make it look like a nice list (like the one before), I guess flake8 is the problem but there might be a solution :)

PR #2219, opened 2020-06-14
CY-2916 Update status reporters to work with supervisord

mabuaisha:


rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -188,8 +188,17 @@ def _lookup_supervisor_service_status(self, service_name):
         try:
             status_response = server.supervisor.getProcessInfo(service_name)
         except xmlrpclib.Fault as e:
-            if not (is_optional or e.faultCode == 10):
+            # If the error raise that means one of the two options:
```
geokala, line 191: If the error is raised

mabuaisha: Fixed

PR #2220, opened 2020-06-15
CY-1915 Integration tests: use docker, not docl

tehasdf:
This simplifies the tests, and is the first step towards a new, better architecture.

Also, now you can parametrize the image.

But now you can't mount the source code.


tests/integration_tests/framework/docker.py
```diff
@@ -227,28 +213,36 @@ def _set_container_id_and_ip(container_details):
 
 
 def _retry(func, exceptions, cleanup=None):
-    for _ in range(600):
+    err = None
+    for _ in range(2600):
```
tehasdf, line 217: this is to be changed but we currently made rabbitmq take a loooong time to start

tests/integration_tests/framework/docl.py
```diff
@@ -227,28 +213,36 @@ def _set_container_id_and_ip(container_details):
...
         try:
             res = func()
             if cleanup:
                 cleanup(res)
-            break
-        except exceptions:
+            return
+        except exceptions as e:
+            err = e
             time.sleep(0.1)
```
geokala, line 225: Nice, terraform style.

PR #2234, opened 2020-06-25
CY-3048 Update restore snapshot to start using supervisord

mabuaisha:


mabuaisha:
@tehasdf we cannot use the cfy_manager restart because it will raise permission errors

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -686,3 +688,15 @@ def get_manager_reporter_info(self):
         """.format(MANAGER_STATUS_REPORTER)
         result = self.run_query(query)
         return result['all'][0][0]
+
+    def get_service_management(self):
+        query = """
+        SELECT
+            json_build_object(
```
tehasdf, line 695: no need for json

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -686,3 +688,15 @@ def get_manager_reporter_info(self):
...
+                'value', value
+            )
+        FROM config
+        WHERE name = '{0}'
+        """.format(SERVICE_MANAGEMENT)
```
tehasdf, line 700: either pass this as an argument (`vars`) or just put it into the string
writing it like this just looks like it wants to be a sql injection but is actually a constant

PR #2235, opened 2020-06-30
Integration tests changes: move env to fixtures

tehasdf:
Moving all the docker/env stuff to fixtures. See each commit.


tests/integration_tests/framework/docker.py
```diff
@@ -161,7 +160,6 @@ def run_manager(image, resource_mapping=None):
     command += [image]
     manager_id = subprocess.check_output(command).strip()
     execute(manager_id, ['cfy_manager', 'wait-for-starter'])
-    upload_mock_license(manager_id)
```
geokala, line 164: Why are we not uploading the license here now?

tests/integration_tests/framework/docker.py
```diff
@@ -225,7 +223,7 @@ def _set_container_id_and_ip(container_details):
 
 def _retry(func, exceptions, cleanup=None):
     err = None
-    for _ in range(2600):
+    for _ in range(1200):
```
geokala, line 226: I know it's unrelated to the change you're making here, but is it worth migrating to the retrying library?

tests/integration_tests/framework/docker.py
tehasdf: i guess we could. It fits for me better to upload it in the other fixture, because that fixture is about keeping on-manager storage consistent, and i feel the license is part of that. While this function is just about starting the container.

I added a new commit that renames & documents the fixture i mean (`prepare_manager_storage`)

tests/integration_tests/framework/docker.py
tehasdf: Maybe. I don't really want to add a dependency for this one use. Also I feel that retrying, with its arcane argument names, is way less readable than just writing it out (note that we are doing non-basic stuff like a cleanup handler)

Anyway, now I noted that this function isn't even used anymore. Now, just the wait-for-starter is used, with its retries and timeouts. So I removed this function instead.

tests/integration_tests/framework/docker.py
geokala: Sounds reasonable.

tests/integration_tests/framework/docker.py
geokala: <thumbs up>

PR #2254, opened 2020-07-09
CY-3096 Fix failed authentication message

glukhman:


rest-service/manager_rest/security/authentication.py
```diff
@@ -31,6 +31,8 @@
 
 
 Authorization = namedtuple('Authorization', 'username password')
+Failed_auth_message = 'Authentication failed for <User username=`{0}`>. ' \
```
alexmolev, line 34: Is there a reason variable starts with capital?

glukhman: nope. i saw the one above it and followed suit. Change to lower case?

PR #2256, opened 2020-07-13
CY-2859 Make concurrent calls to remote Prometheuses

mateumann:


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -89,6 +91,72 @@ def _store_ca(cls, service_type, service_node):
...
+                self._out_queue.put((service_node_name,
+                                     prometheus_response,))
+
+    def append(self, service_type, service_node, service_node_name,
+               status_func):
+        self._in_queue.put((service_type, service_node, service_node_name,
+                            status_func,))
+        self._in_queue_len += 1
+        self._result_cache[service_node_name] = []
+        for thread in [t for t in self._threads if not t.is_alive()]:
```
tehasdf, line 134: doing this in `append` is a bit weird. We don't need the threads in append, really, we need them in get_responses, for getting the responses.
How about, in `__init__`, `self._threads = None` and then in `get_responses`,  `if self._threads is None:  self._threads = [...` and also start them right there

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -273,14 +341,14 @@ def get_service_node(sn_name):
...
+    status_getter = get_concurrent_status_checker()
     for service_node_name in remote_node_names:
         service_node = get_service_node(service_node_name)
-        creds = Credentials.get(service_type, service_node.private_ip)
-        prometheus_response = status_func(
-            'https://{ip}:53333/monitoring/'.format(
-                ip=service_node.private_ip),
-            auth=(creds.username, creds.password),
-            ca_path=creds.ca_path)
+        status_getter.append(
```
tehasdf, line 347: the `.append` + `.get_responses` calling approach seems too complex, can't it just be `.get_responses(remote_node_names)` or similar?

tehasdf: or of course `.get_responses(get_service_node(name) for name in remote_node_names)` or however you want to spell it :)

PR #2258, opened 2020-07-14
RD-664 Allow using alembic directly to migrate

tehasdf:
This allows using alembic directly, eg.:
```bash
cd /opt/manager/resources/cloudify/migrations/
/opt/manager/env/bin/alembic upgrade head
```
instead of having to go through flask-migrate.

Also, `alembic upgrade --sql head > file.sql` now works too!

The changes are about replacing ORM usage in the migrations, with
sql/expression-layer stuff instead, so eg. instead of `session.add(obj)`,
to insert things in the migration, we use `op.bulk_insert()` with a
list of dicts, etc.

### DIFF
--- a/resources/rest-service/cloudify/migrations/versions/387fcd049efb_5_0_5_to_5_1.py
+++ b/resources/rest-service/cloudify/migrations/versions/387fcd049efb_5_0_5_to_5_1.py
@@ -13,14 +13,12 @@ Create Date: 2020-03-30 06:27:26.747213
 
 from alembic import op
 import sqlalchemy as sa
-from sqlalchemy import orm, sql
+from sqlalchemy.sql import table, column, expression
 from manager_rest import storage
 from sqlalchemy.dialects import postgresql
-from sqlalchemy.ext.declarative import declarative_base
 
 from cloudify.models_states import VisibilityState
 
-from manager_rest.storage.models import User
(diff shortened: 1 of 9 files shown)


mateumann (dismissed):

resources/rest-service/cloudify/migrations/versions/387fcd049efb_5_0_5_to_5_1.py
```diff
@@ -35,62 +33,55 @@
                                   name='visibility_states',
                                   create_type=False)
 
-Base = declarative_base()
-
-
-class Config(Base):
```
tehasdf, line 41: note that this class is not used here to create the table, because the table is already created elsewhere. It is only used to insert entries with the ORM (which requires a model to do so).

Instead, we declare just a table, with a `config_table = table([columns])`, and we only declare the table for use with `op.bulk_insert`, again - not for creating it.

So we only there declare the columns that we are actually going to use, we don't need to declare all columns, and we don't need to declare constraints etc either.
Think of that table declaration as just telling sqlalchemy about the schema, to use in a `INSERT INTO config (x, y, z) VALUES (...)` <-- it needs to know the "x y z", but that statement doesn't need to know about unused columns or about constraints.

mateumann (approved):

PR #2264, opened 2020-07-21
CY-3094 Plugin force-install endpoints

tehasdf:
Add an endpoint to force-install a plugin, add a PluginState bookkeeping thing

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/plugins.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/plugins.py
@@ -17,9 +17,10 @@ from flask_restful.inputs import boolean
 from flask_restful_swagger import swagger
 from flask_restful.reqparse import Argument
 
+from cloudify._compat import text_type
 from cloudify.models_states import VisibilityState
-from cloudify.plugins.install_utils import remove_status_prefix
 
+from manager_rest import manager_exceptions
 from manager_rest.security import SecuredResource
 from manager_rest.plugins_update.constants import PHASES
 from manager_rest.security.authorization import authorize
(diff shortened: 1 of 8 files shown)


rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -161,18 +162,80 @@ def get(self,
...
+        This method is for internal use only.
+        """
+        request_dict = rest_utils.get_json_and_verify_params({
+            'agent': {'type': text_type, 'optional': True},
+            'manager': {'type': text_type, 'optional': True},
+            'state': {'type': text_type},
+            'error': {'type': text_type, 'optional': True},
+        })
         sm = get_storage_manager()
+        agent = sm.get(
```
geokala, line 207: Wouldn't it be better to gate checking the agent/manager listing on what was passed in in the request dict to avoid an unnecessary DB lookup on every manager-side install?

rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -161,18 +162,80 @@ def get(self,
 
 
 class PluginsId(resources_v2_1.PluginsId):
+    @authorize('plugin_upload')
+    @rest_decorators.marshal_with(models.Plugin)
+    def post(self, plugin_id, **kwargs):
+        """Force plugin installation on the given managers or agents.
+
+        This method is for internal use only.
```
geokala, line 170: Is this really for internal use? I thought it was going to be called by users to cause the plugins to be pre-warmed?

rest-service/manager_rest/workflow_executor.py
```diff
@@ -252,3 +252,54 @@ def cancel_execution(execution_id):
...
+def install_plugin_on_managers(plugin, target_names):
+    _send_mgmtworker_task(
+        _get_install_plugin_message(plugin, target_names=None),
+        routing_key='service')
+
+
+def install_plugin_on_agents(plugin, target_agents):
+    # all the agents are supposed to be on the same tenant, so we only need
+    # one amqp client, but let's check that. There should never be a way
+    # to get agents from multiple tenants into this function.
```
geokala, line 285: What about with deployments with visibility: shared?

rest-service/manager_rest/rest/resources_v3_1/plugins.py
tehasdf: internal use, as in only to be called by our code, and not documented as a rest endpoint. I don't want to make this public yet, I'm not absolutely sure this is as good as it can be (wrt. POST/PUT). But I don't have a better idea yet. Maybe we can make this public in 5.2.

rest-service/manager_rest/workflow_executor.py
tehasdf: good point. Well, disallowed for now. But I've pushed a commit that allows installing plugins on those, even if you need N requests to install on agents from N tenants.

rest-service/manager_rest/rest/resources_v3_1/plugins.py
tehasdf: yeah, rewritten this part

rest-service/manager_rest/rest/resources_v3_1/plugins.py
geokala: <thumbs up>

rest-service/manager_rest/workflow_executor.py
geokala: Sounds reasonable on this pass.

PR #2280, opened 2020-07-29
CY-3226 Fix issues with components test cases

mabuaisha:


tests/integration_tests/tests/agentless_tests/component/test_component_basics.py
```diff
@@ -84,8 +87,8 @@ def test_component_creation_with_secrets_and_plugins(self):
         self.assertTrue(self.client.deployments.get(self.component_name))
         self.assertEqual(self.client.secrets.get('secret1')['value'], 'test')
         plugins_list = self.client.plugins.list()
-        self.assertEqual(len(plugins_list), 1)
-        self.assertTrue(plugins_list[0]['package_name'],
+        self.assertEqual(len(plugins_list), 2)
+        self.assertTrue(plugins_list[1]['package_name'],
```
tehasdf, line 91: we can't assert on the plugin list like that because it used to be `1`, now it's `2`, but if you also have another test in this run which uploads another plugin before, then it'll be `3` or `4` or...

instead, do something like `assert any(p['package_name'] == 'cloudify-openstack-plugin' for p in plugins_list)`, or add a package_name filter to that restclient `.list()` call

tests/integration_tests/tests/agentless_tests/component/test_component_cascading.py
```diff
@@ -27,12 +29,17 @@
...
 
 
+@pytest.mark.usefixtures('testmockoperations_plugin')
 class ComponentCascadingCancelAndResume(AgentlessTestCase):
     def _wait_for_component_install(self, deployment_id):
         # waiting for component to create it's deployment
-        do_retries(verify_deployment_env_created,
-                   30,
-                   deployment_id=deployment_id)
+        do_retries(
```
tehasdf, line 36: this is here many times, package this into a method?

tests/integration_tests/tests/agentless_tests/shared_resource/test_relationship_running_workflow.py
```diff
@@ -11,24 +11,25 @@
...
 
+@pytest.mark.usefixtures('mock_workflows_plugin')
 class TestRelationshipRunningRemoteWorkflow(AgentlessTestCase):
     shared_resource_blueprint = """
 tosca_definitions_version: cloudify_dsl_1_3
 
 imports:
     - cloudify/types/types.yaml
-    - wf--blueprint:mock_workflows
+    - plugin:mock_workflows
```
tehasdf, line 26: hmm no, this makes it a different test, the point was using something from a different blueprint

tests/integration_tests/tests/agentless_tests/test_inter_deployment_dependencies.py
```diff
@@ -232,15 +238,17 @@ def _assert_dependencies_exist(self,
...
+        comp_target_id = COMP_DEPLOYMENT1
+        if not is_first_state:
+            shared_deployment_target = SR_DEPLOYMENT2
+            comp_target_id = COMP_DEPLOYMENT2
+
+        static_changed_to_static = shared_deployment_target
+        static_changed_to_runtime = shared_deployment_target
+        runtime_changed_to_runtime = shared_deployment_target
+        runtime_changed_to_static = shared_deployment_target
+        shared_resource_target_id = shared_deployment_target
```
tehasdf, line 251: i don't think we need 5 names for the same variable.. :D

tests/integration_tests/tests/agentless_tests/component/test_component_basics.py
mabuaisha: Done and updated

tests/integration_tests/tests/agentless_tests/component/test_component_cascading.py
mabuaisha: Done and updated

tests/integration_tests/tests/agentless_tests/shared_resource/test_relationship_running_workflow.py
mabuaisha: Done and updated

tests/integration_tests/tests/agentless_tests/test_inter_deployment_dependencies.py
mabuaisha: Done and updated

PR #2297, opened 2020-08-07
CY-3306 Add services section to nodes of cluster status

mateumann:


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -54,36 +54,45 @@
...
+    'cloudify-stage': CloudifyService(
+        name='Cloudify Console',
+        description='Cloudify Console service'),
+    'cloudify-syncthing': CloudifyService(
+        name='Syncthing',
+        description='Syncthing service'),
     'haproxy': CloudifyService(
         name='HA Proxy',
-        description='HAProxy Load Balancer'),
+        description='HAProxy Load Balancer service'),
```
geokala, line 83: Hmm... do we need all of the services to have 'service' at the end of their description? I'd probably have gone with removing it from the others rather than adding it to this (and maybe making the descriptions more meaningful such as syncthing's being 'File synchronisation between manager nodes' and console being 'Cloudify Web UI' or something?

PR #2298, opened 2020-08-09
CY-3314 Fix replacing certs on agents

YoniItzhak:


YoniItzhak:
I got an error while trying to call the endpoint PTACH /agents because apparently, in the cluster case, the rest call [has to return a response with json](https://github.com/cloudify-cosmo/cloudify-cli/blob/3cd1d8b7222dba068cd78342e6aaadb77b5db44c/cloudify_cli/env.py#L581), otherwise it's failing.

PR #2300, opened 2020-08-10
5.1 migration: add CASCADE to plugins_states fks

tehasdf:
Yeah I just wrote that migration by hand, before, I think.
Now, it's actually auto-generated, and shouldn't be missing things
anymore.
Whoops.


tehasdf:
the only real difference is the cascade, but i've taken the autogenerated one instead, so the style is different

PR #2301, opened 2020-08-10
Integration tests: cleanups, more mounts, drop PluginsTest

tehasdf:
Dropping the unused cruft, mostly


tests/integration_tests/conftest.py
```diff
@@ -45,7 +45,9 @@ def pytest_addoption(parser):
 # pairs of (source path, [list of target virtualenvs])
 # TODO fill this in as needed, when needed
 sources = [
+    ('cloudify-cli/cloudify_cli', ['/opt/cfy']),
     ('cloudify-common/cloudify', ['/opt/manager/env', '/opt/mgmtworker/env']),
```
geokala, line 49: I know it's not part of this change, but isn't this needed in /opt/manager/env too?

tehasdf: nonono, thankfully we don't install the cli into the restservice venv

geokala: I mean cloudify-common

geokala: Ignore me, I misread this line.

PR #2305, opened 2020-08-11
5.1 migration: remove duplicated drop

tehasdf:


tehasdf:
the other drop is at https://github.com/cloudify-cosmo/cloudify-manager/blob/b35cbe335662dcf7120998b6f964ecc3091d12e7/resources/rest-service/cloudify/migrations/versions/387fcd049efb_5_0_5_to_5_1.py#L98

PR #2308, opened 2020-08-13
CY-3277 Use dockercompute fixture

glukhman:


tests/integration_tests/resources/dsl/agent_tests/plugin-requires-old-package-blueprint/plugins/requires-old-package-plugin/setup.py
```diff
@@ -22,6 +22,6 @@
     packages=['requires_old_package_plugin'],
     install_requires=[
         # The old repo/package before we merged it to cloudify-common
-        "cloudify-plugins-common>=5.1.0.dev1"
+        "cloudify-common==5.0.0"
```
alexmolev, line 25: Why downgraded?

alexmolev (changes requested):

tests/integration_tests/resources/dsl/agent_tests/plugin-requires-old-package-blueprint/plugins/requires-old-package-plugin/setup.py
glukhman: Well, for one, cloudify-plugins-common has no version above 4.3.2.
cloudify-common, which we've been using since, has of course 5.1.0.dev1, but the test is meant to check whether a plugin which requires an *older* package can work. 5.0.0 is the earliest package which doesn't error for reasons such as changing how our MQ works.

tests/integration_tests/framework/amqp_events_printer.py
```diff
@@ -73,7 +73,7 @@ def callback(ch, method, properties, body):
                 ev = json.loads(body)
                 output = cloudify.logs.create_event_message_prefix(ev)
                 if output:
-                    sys.stdout.write('{0}\n'.format(output))
+                    sys.stdout.write('{0}\n'.format(output.encode('utf-8')))
```
tehasdf, line 76: this makes it print with a `b''` prefix of course, because we're formatting an .encoded bytes into a unicode (`{0}\n`)

but really, just make this `print(output)`

tests/integration_tests/framework/hello_world.py
```diff
@@ -70,16 +71,15 @@ def _webserver_request():
...
+        events = self.test_case.client.events.list(
             deployment_id=deployment_id)
         self.test_case.assertGreater(len(events.items), 0)
 
     def _prepare_hello_world(self):
         logger = self.test_case.logger
         repo_name = 'cloudify-hello-world-example'
-        branch = self.test_case.env.core_branch_name
-        workdir = self.test_case.env.test_working_dir
+        branch = os.environ.get(BRANCH_NAME_CORE, 'master')
```
tehasdf, line 81: oh nooo, it's the "random undocumented envvars" again, let's not do this

and actually, I don't see what does this test_hello_world test (which is split between 3 files - this, `agent_tests/test_hello_world.py`, and the base `test_cases.py`) actually test, that isn't tested by just the agent install test

drop it if you want, otherwise i'll drop it

tests/integration_tests/resources/dsl/agent_tests/dockercompute_timezone.yaml
```diff
@@ -16,7 +16,7 @@ node_templates:
   agent_host:
     type: cloudify.nodes.docker.Compute
     properties:
-      install_agent: true
+      install_agent: none
```
tehasdf, line 19: none, why?

tests/integration_tests/tests/agent_tests/test_plugin_update.py
```diff
@@ -274,23 +269,31 @@ def _perform_update(self, update_plugins=True):
...
+        plugins_dir = os.path.dirname(integration_tests_plugins.__file__)
+        plugin_source = os.path.join(plugins_dir, plugin_name)
+
+        wagon_paths = self._get_or_create_wagon(plugin_source)
+        for wagon_path in wagon_paths:
+            yaml_path = os.path.join(plugin_source, 'plugin.yaml')
+            with zip_files([wagon_path, yaml_path]) as zip_path:
+                self.client.plugins.upload(zip_path, visibility='global')
+        self.logger.info(
+            'Finished uploading {0}...'.format(plugin_name))
```
tehasdf, line 296: what's all this, we already have the fixture

tests/integration_tests_plugins/version_aware_v2/version_aware/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        yield data.setdefault(deployment_id, {})
+        with open(storage_file_path, 'w') as f:
+            json.dump(data, f, indent=2)
+            f.write(os.linesep)
+
+
+@contextmanager
+def storage_file_lock(storage_file_path):
+    with fasteners.InterProcessLock('{0}.lock'.format(storage_file_path)):
+        yield
```
tehasdf, line 68: hah did we make fasteners available for plugins by accident?

anyway, best replace this with just using runtime-props, as the other plugins did

tests/integration_tests/tests/agent_tests/test_plugin_update.py
glukhman: In _test_single_deployment_is_updated_ and _test_many_deployments_are_updated_ I need to upload version 2 of the plugin **after** the BP is installed. Don't see how can I use the fixture for that

tests/integration_tests/resources/dsl/agent_tests/dockercompute_timezone.yaml
glukhman: dropped this

tehasdf (approved):

alexmolev (approved):

PR #2317, opened 2020-08-21
CY-3361 Don't use dateutil for time comparison

geokala:
It's not in 5.0.5 and it's not needed anyway for what we're doing.


geokala:
Merged manually to the old 5.0.5 branch.

PR #2324, opened 2020-08-31
CY-3301 Separate node services per node type

mateumann:


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -100,6 +100,36 @@
...
 }
+SERVICE_ASSIGNMENTS = {
+    CloudifyNodeType.DB: [
+        'etcd',
+        'haproxy',
+        'nginx',
+        'node_exporter',
+        'patroni',
+        'postgres_exporter',
+        'postgresql-9.5',
```
geokala, line 111: This service shouldn't be on a DB node, only on an AIO node.

geokala: (though we do have other code that handles that, so not necessarily a major concern at this point)

geokala (approved):
Approved- I still feel like this part of the code is overcomplicated but I think cleaning it up is not a task to be done close to release.

rest-service/manager_rest/cluster_status_manager.py
mateumann: yes, exactly my attitude here

mateumann:
Yes it is too complicated, as is the returned data structure.  But (this comment concerns the latter) we should maintain backward compatibility.

geokala:
Agreed, back compat is important, but let's review the actual code structure at some point and see if we can make it easier to follow... somehow. <smiles>

PR #2334, opened 2020-09-09
CY-3420 Make DELETE endpoints return 204

tehasdf:
Deletes return 204, instead of including all the data and ALL the related rows.

Also, use cascades wherever possible.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v2/plugins.py
+++ b/rest-service/manager_rest/rest/resources_v2/plugins.py
@@ -186,8 +186,6 @@ class PluginsId(SecuredResource):
     @authorize('plugin_delete')
     @rest_decorators.marshal_with(models.Plugin)
     def delete(self, plugin_id, **kwargs):
-        """
-        Delete plugin by ID
-        """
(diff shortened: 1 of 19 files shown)


rest-service/manager_rest/rest/resources_v2/plugins.py
```diff
@@ -186,8 +186,5 @@ def get(self, plugin_id, _include=None, **kwargs):
     @authorize('plugin_delete')
     @rest_decorators.marshal_with(models.Plugin)
     def delete(self, plugin_id, **kwargs):
-        """
-        Delete plugin by ID
-        """
-        return get_resource_manager().remove_plugin(plugin_id=plugin_id,
-                                                    force=False)
+        """Delete plugin by ID"""
+        get_resource_manager().remove_plugin(plugin_id=plugin_id, force=False)
```
geokala, line 190: Doesn't this need to return a 204?

tehasdf: it does now

geokala (approved):

PR #2335, opened 2020-09-14
CY-3468 Add execution id to hook operation context

glukhman:


tehasdf (approved):

glukhman:
execution_id is added to the operation_context, so that _validate_not_cancelled, which gets the operation context of the hook task, has the execution id to validate on

PR #2337, opened 2020-09-15
CY-3440 Status reporter fixup

geokala:
Refactored to make the code easier to follow and to make less requests.

This should mean that if any of the checks for status reporter behaviour show it behaving incorrectly it'll be easier to track down the cause (hopefully!).

### DIFF
--- a/rest-service/manager_rest/cluster_status_manager.py
+++ b/rest-service/manager_rest/cluster_status_manager.py
@@ -1,25 +1,6 @@
-#########
-# Copyright (c) 2019 Cloudify Platform Ltd. All rights reserved
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#       http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-#  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#  * See the License for the specific language governing permissions and
-#  * limitations under the License.
-
-import re
-import os
-import atexit
 import threading
 import queue
-from collections import namedtuple
-from datetime import datetime, timedelta
+from datetime import datetime
 
 from flask import current_app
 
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/syncthing_status_manager.py
```diff
@@ -0,0 +1,90 @@
...
+    from cloudify_premium import syncthing_utils
+    from cloudify_premium.ha import utils as ha_utils
+except ImportError:
+    syncthing_utils = None
+    ha_utils = None
+
+
+def _last_manager_in_cluster():
+    storage_manager = get_storage_manager()
+    managers = storage_manager.list(models.Manager,
```
tehasdf, line 21: probably `.full_access_list` because we don't need the pagination or the tenant stuff here. Except that full_access_list of course doesn't have a `sort` so maybe add that. Or feel free to not sort.

rest-service/manager_rest/syncthing_status_manager.py
```diff
@@ -0,0 +1,90 @@
...
+
+from cloudify.cluster_status import (ServiceStatus,
+                                     NodeServiceStatus)
+
+from manager_rest.storage import models, get_storage_manager
+from manager_rest.rest.rest_utils import parse_datetime_string
+
+try:
+    from cloudify_premium import syncthing_utils
+    from cloudify_premium.ha import utils as ha_utils
```
tehasdf, line 13: unused?

rest-service/manager_rest/syncthing_status_manager.py
```diff
@@ -0,0 +1,90 @@
...
+
+    current_app.logger.debug(
+        'Inactive File Sync Service - no other devices were seen by it'
+    )
+    return False
+
+
+def get_syncthing_status():
+    try:
+        syncthing_config = syncthing_utils.config()
```
tehasdf, line 75: and what if syncthing_config is none?

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -236,511 +195,262 @@ def get_concurrent_status_checker():
...
 
-def _get_cloudify_service_description(metric_name):
-    if metric_name in SERVICE_DESCRIPTIONS:
-        return SERVICE_DESCRIPTIONS[metric_name]
-    elif metric_name.endswith('.service'):
-        return _get_cloudify_service_description(metric_name[:-8])
-    return None
+    for result in prometheus_results:
+        metric = result.get('metric', {})
+        value = result.get('value', [0, ''])
```
tehasdf, line 424: maybe `timestamp, is_healthy = ...` unpack here so that those things have a name? with just `[0, '']`, it's hard to tell what are we looking at

tehasdf (commented):
i'll read through `cluster_status_manager.py` later, lunchtime now

rest-service/manager_rest/syncthing_status_manager.py
```diff
@@ -0,0 +1,90 @@
...
+                                    sort={'last_seen': 'desc'})
+    active_managers = 0
+    for manager in managers:
+        # Probably new manager, first status report is yet to arrive
+        if manager.status_report_frequency is None:
+            active_managers += 1
+        else:
+            # The manager is considered active, if the time passed since
+            # it's last_seen is maximum twice the frequency interval
+            # (Nyquist sampling theorem)
```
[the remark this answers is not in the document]
geokala: Removing the nyquist line

rest-service/manager_rest/syncthing_status_manager.py
```diff
@@ -0,0 +1,90 @@
...
+    for device_id, stats in device_stats.items():
+        last_seen = parse_datetime_string(stats['lastSeen'])
+
+        # Syncthing is valid when at least one device was seen recently
+        if last_seen > min_last_seen:
+            return True
+    return False
+
+
+def _is_syncthing_valid(syncthing_config, device_stats):
```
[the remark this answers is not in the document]
geokala: This was a straight copy across. I'll add a docstring. It probably does want a better name, but I think the docstring will be a reasonable starting point.

rest-service/manager_rest/syncthing_status_manager.py
geokala: Hmm. I'm not sure this one is worth rewriting since pagination would only take effect if the manager count surpassed the page size so probably not worth mucking about with as this was just copied code?

rest-service/manager_rest/syncthing_status_manager.py
geokala: Removing

rest-service/manager_rest/syncthing_status_manager.py
geokala: I think addressing that would fit better under a different jira, this is the original code.

rest-service/manager_rest/cluster_status_manager.py
geokala: Makes sense, modified.

tehasdf (approved):

PR #2343, opened 2020-09-21
CY-3539 Error logging: make auth error logs DEBUG

tehasdf:
All kinds of bad auth can flood the logs with useless logs.
The big offender is the status checkers, as even the built-in one
generates one log every 15 seconds, and of course there can be
user-specific ones on top of that.

These logs are also useless for detecting attacks, as they don't
include any kind of identification with them (eg. an IP). Nginx logs
serve better for that.

Hide the auth logs by default, by making them debug-level.

### DIFF
--- a/rest-service/manager_rest/app_logging.py
+++ b/rest-service/manager_rest/app_logging.py
@@ -1,5 +1,6 @@
 import sys
 import logging
+import warnings
 
 from flask import current_app, request
 from logging.handlers import WatchedFileHandler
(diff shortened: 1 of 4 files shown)


alexmolev (dismissed):

tehasdf:
specialized the error, with a subclass, and deprecated the useless function. See each commit message for details.

rest-service/manager_rest/app_logging.py
```diff
@@ -75,6 +76,10 @@ def log_response(response):
 
 
 def raise_unauthorized_user_error(extra_info=None):
+    warnings.warn(
+        'raise_unauthorized_user_error is deprecated, raise an '
+        'UnauthorizedError directly instead',
+        DeprecationWarning)
```
tehasdf, line 82: deprecating this, and CY-3540 is about removing it

geokala (approved):

PR #2347, opened 2020-09-23
CY-3547 Generate mapping for plugins update

mateumann:
Add a script that produces a mapping file that can be used later on to update blueprints and/or blueprint plans.


rest-service/setup.py
```diff
@@ -58,6 +58,12 @@
     description='Cloudify manager rest service',
     zip_safe=False,
     install_requires=install_requires,
+    entry_points={
+        'console_scripts': [
+            'update-plugin-imports = '
+            'restmanager_rest.shell.update_plugin_imports:main',
```
tehasdf, line 64:
```suggestion
            'manager_rest.shell.update_plugin_imports:main',
```

PR #2366, opened 2020-09-30
RESTservice: use a request-local for the .parser_context

tehasdf:
We can't just set the parser settings on the app, because that can
get persisted across requests, and then multiple requests will
wrongly use the same parser settings (eg. same plugin_mappings).

Instead, use a per-request variable.

It would be good to not use these context variables at all, but
just pass the arguments through instead, but for now...


rest-service/manager_rest/test/infrastructure/test_parse_with_resolver.py
```diff
@@ -109,28 +109,25 @@ def mock_create_import_resolver(resolver_section):
         self.assertEqual(create_import_resolver_inputs[0]['rules'],
                          resolver_section['rules'])
 
-        self.assertEqual(self.app.application.parser_context['resolver'],
```
tehasdf, line 112: this assert checks that after request, the context stays; this is exactly the opposite of what we want to happen

mateumann (approved):

PR #2380, opened 2020-10-05
bump sqlalchemy to 1.3.19

glukhman:


tehasdf (approved):

glukhman:
The change is because SQLAlchemy 1.3.X made a new class AssociationProxyInstance which contains the actual data of AssociationProxy, so to make AssociationProxy stateless. As a result the objects we get which were before of type AssociationProxy, are now of various subtypes of AssociationProxyInstance, thus not having __is_attribute_ and _extension_type_ fields

PR #2382, opened 2020-10-05
CY-3615 Use lists for AMQP host/management host

tehasdf:
We can have multiple brokers, so let's use a list for storing
their addresses.

### DIFF
--- a/rest-service/manager_rest/config.py
+++ b/rest-service/manager_rest/config.py
@@ -184,11 +184,11 @@ class Config(object):
             setattr(self, conf_value.name, conf_value.value)
 
         stored_brokers = session.query(models.RabbitMQBroker).all()
-        broker_hosts = [broker.host for broker in stored_brokers]
+        self.amqp_host = [b.host for b in stored_brokers]
+        self.amqp_management_host = [b.management_host for b in stored_brokers]
         for broker in stored_brokers:
-            # currently, there's going to be only one rabbitmq
-            self.amqp_host = broker_hosts
-            self.amqp_management_host = broker.management_host
+            # all brokers will use the same credentials
+            # (rabbitmq replicates users)
             self.amqp_username = broker.username
             self.amqp_password = broker.password
             self.amqp_ca_path = broker.write_ca_cert()


rest-service/manager_rest/config.py
```diff
@@ -184,11 +184,11 @@ def load_from_db(self):
             setattr(self, conf_value.name, conf_value.value)
 
         stored_brokers = session.query(models.RabbitMQBroker).all()
-        broker_hosts = [broker.host for broker in stored_brokers]
+        self.amqp_host = [b.host for b in stored_brokers]
+        self.amqp_management_host = [b.management_host for b in stored_brokers]
         for broker in stored_brokers:
-            # currently, there's going to be only one rabbitmq
-            self.amqp_host = broker_hosts
```
tehasdf, line 190: oh look, this was already a list

the list is already handled by the amqp client, and it is even copied:
https://github.com/cloudify-cosmo/cloudify-common/blob/208d78287eacff09a3e160368136e99ae8eaeaf9/cloudify/amqp_client.py#L147-L150

rest-service/manager_rest/config.py
```diff
@@ -184,11 +184,11 @@ def load_from_db(self):
             setattr(self, conf_value.name, conf_value.value)
 
         stored_brokers = session.query(models.RabbitMQBroker).all()
-        broker_hosts = [broker.host for broker in stored_brokers]
+        self.amqp_host = [b.host for b in stored_brokers]
+        self.amqp_management_host = [b.management_host for b in stored_brokers]
```
tehasdf, line 188: this is now a list too, and the client copes with both a list and a single value,

https://github.com/cloudify-cosmo/cloudify-common/blob/e980d9222160c3c8562f497fe97e0045b70c4d7c/cloudify/rabbitmq_client.py#L32-L33

tehasdf: I've verified that all usages of AMQPClient inside of cloudify-manager do create a new client per request, so we're OK wrt. all the `.pop()` calls that the client does

glukhman (approved):

PR #2396, opened 2020-10-16
CY-3473 Remove proxy to fix issues with certs

geokala:


geokala:
Still needs at least:
1. Logging
2. Update to flask_utils.py to stop using its own logic to generate the same db conn string.

geokala:
flask_utils has been modified to use the same code.

Now just needs some logging.

geokala:
Logging is now in place, and the general behaviour seems fine.

tehasdf (approved):

PR #2397, opened 2020-10-19
CY-3689 Update monitoring on manager addition/removal

mateumann:
Not only Syncthing should be notified on changes to the manager nodes list,
but also a Cloudify Monitoring Service.  The latter one is going to rewrite
federated managers target list.

### DIFF
--- a/mgmtworker/mgmtworker/monitoring.py
+++ b/mgmtworker/mgmtworker/monitoring.py
@@ -32,7 +32,7 @@ def update_manager_targets(rest_client):
 def _other_managers_private_ips(manager_client):
     """Generate other managers' private_ips."""
     my_hostname = get_manager_name()
-    for manager in manager_client.get_managers().items:
+    for manager in manager_client.get_managers():
         if manager[HOSTNAME] != my_hostname:
             yield manager[PRIVATE_IP]
 
--- a/mgmtworker/mgmtworker/monitoring.py
+++ b/mgmtworker/mgmtworker/monitoring.py
@@ -1,19 +1,65 @@
-import logging
-logger = logging.getLogger('mgmtworker')
+import json
+
+from jinja2 import Template
+from os import close, rename, sep
+from os.path import join
+from tempfile import mkstemp
+
+from cloudify.utils import setup_logger
+
+logger = setup_logger('cloudify.monitoring')
+
+PRIVATE_IP = 'private_ip'
+PUBLIC_IP = 'public_ip'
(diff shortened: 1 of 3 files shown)


mgmtworker/mgmtworker/monitoring.py
```diff
@@ -0,0 +1,64 @@
...
+            yield manager[PRIVATE_IP]
+
+
+def _deploy_prometheus_targets(destination, targets, labels):
+    """Deploy a target file for prometheus.
+    :param destination: Target file name in targets dir.
+    :param targets: List of targets for prometheus.
+    :param labels: Dict of labels with values for prometheus."""
+    return _render_template(
+        Template(PROMETHEUS_TARGETS_TEMPLATE),
```
tehasdf, line 46: for a 2-line template, we could just ''.format instead i guess

mgmtworker/mgmtworker/monitoring.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+
+def _render_template(template, destination, **kwargs):
+    """Render a Jinja2 template into a file destination.
+    :param template: A Jinja Template to be rendered
+    :param destination: Destination file name.
+    :param kwargs: Arguments for the template render."""
+    content = template.render(kwargs)
+    fd, file_path = mkstemp()
```
tehasdf, line 59: use NamedTemporaryFile instead of this awkward ancient api

mateumann: good idea; sure

mgmtworker/mgmtworker/monitoring.py
mateumann: good point

mgmtworker/mgmtworker/monitoring.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+
+
+def _render_template(template, destination, **kwargs):
+    """Render a template into a file destination.
+    :param template: A text template to be rendered
+    :param destination: Destination file name.
+    :param kwargs: Arguments for the template render."""
+    content = template.format(**kwargs)
+    with tempfile.NamedTemporaryFile(delete=False) as f:
```
geokala, line 58: I'm not seeing where we clean up these files, meaning there will be gradually more of them over time.

Also, I'm not convinced it's necessary to write this to a temp location- can't you just overwrite the destination since you don't need sudo?

geokala (approved):

PR #2401, opened 2020-10-21
RD-168 Update cluster status retrieval

mateumann:


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -98,7 +98,7 @@
 
 QUERY_STRINGS = {
     CloudifyNodeType.DB:
-        '(up{job=~".*postgresql"} == 1 and pg_up{job=~".*postgresql"} == 1)',
+        '(pg_up{job=~".*postgresql"} == 1 and up{job=~".*postgresql"} == 1)',
```
mateumann, line 101: This way we don't report vague `up` metrics, but a more precise `pg_up`.

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -546,6 +546,7 @@ def _process_metric(metric, timestamp, healthy):
         'healthy': healthy,
         'metric_type': metric.get('__name__', 'unknown'),
         'metric_name': metric_name,
+        'host': metric.get('host', 'localhost')
```
mateumann: Relying on specific format of `metric_name` is probably not the best idea.

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -428,14 +449,19 @@ def append_service_result(pm, res):
...
             if 'instances' not in service_results[dm]['extra_info'][pm]:
                 service_results[dm]['extra_info'][pm]['instances'] = []
-            # ... append an instance to the list of instances
-            service_results[dm]['extra_info'][pm]['instances'].append(
-                res['extra_info'][pm].get('instances'))
+            # keep only one instance per host
+            for res_instance in res['extra_info'][pm].get('instances'):
+                if res_instance.get('host') not in [
+                        instance.get('host') for instance in
+                        service_results[dm]['extra_info'][pm]['instances']]:
```
mateumann, line 461: This is madness, but rather a necessary one. We want to be backwards compatible, thus we should keep this extremely deeply nested structure.

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -517,13 +546,16 @@ def _process_metric(metric, timestamp, healthy):
...
+        metric_name += ' ({})'.format(metric.get('instance'))
+    if metric.get('host'):
+        metric_name += ' on {}'.format(metric.get('host'))
 
     return {
         'last_check': last_check,
         'healthy': healthy,
         'metric_type': metric.get('__name__', 'unknown'),
         'metric_name': metric_name,
+        'host': metric.get('host', 'localhost'),
```
mateumann, line 558: Relying on specific `metric_name` format is probably not the best idea.

tehasdf (approved):

PR #2414, opened 2020-10-29
RD-350 Fix profile load to load new JSON profile from the container

glukhman:


tests/integration_tests/framework/utils.py
```diff
@@ -58,24 +60,9 @@ def sh_bake(command):
...
-
-
-def get_manager_tenant():
-    return cli_env.get_tenant_name()
-
-
-def update_profile_context():
-    cli_env.profile = cli_env.get_profile_context(get_manager_ip())
+def get_profile_context(container_id):
+    profile_context_cmd = 'cat /root/.cloudify/profiles/localhost/context.json'
```
mabuaisha, line 64: @glukhman what about exposing the manager info using the same api used by the old framework. I mean at the end of the day we can load the context for the active profile then it should be fine right to load the manager info ?

mabuaisha (approved):

PR #2441, opened 2020-11-12
inte-tests: daemonize the script when installing an agent

tehasdf:
Detaching from the parent process so that the agent is not
killed when the mgmtworker dies

Also, the ad-hoc detach in test_resume.py is now unnecessary

### DIFF
--- a/tests/integration_tests_plugins/dockercompute/dockercompute/operations.py
+++ b/tests/integration_tests_plugins/dockercompute/dockercompute/operations.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import subprocess
 import tempfile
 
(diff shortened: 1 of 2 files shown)


tests/integration_tests_plugins/dockercompute/dockercompute/operations.py
```diff
@@ -25,7 +26,21 @@ def start(ctx, **_):
...
         f.write(install_agent_script)
-    subprocess.check_call(['bash', f.name])
+
+    # daemonize the script, so that we're not a child of the mgmtworker,
+    # so the agent is not killed when the mgmtworker dies
+    if os.fork() == 0:
+        if not os.path.exists('/tmp/supervisor.sock'):
+            # under systemd, also escape the mgmtworker's cgroup
+            subprocess.check_call([
+                'sudo', '/usr/bin/bash', '-c',
```
tehasdf, line 36: this _can_ well run under supervisord as well, but it's the only `sudo` call in the whole thing, so I'm adding the `if` so that it doesn't require sudo at all under supervisord.
(we are allowing sudo in the allow_agents fixture)

mabuaisha (approved):

PR #2446, opened 2020-11-15
Integ tests fix test_get_instance_ip and test_executions

glukhman:


glukhman:
In _get_instance_ip.yaml_, having the source and target relationships establish call the same function, triggered an (IIUC unplanned) update of node instances, which might then result in a version conflict

glukhman:
Following RD-208 which replaced tasks polling with callbacks, running  _test_legacy_cancel_execution_ which makes mack_workflows' sleep_with_cancel_support return api.EXECUTION_CANCELLED_RESULT, will raise the newly-added RuntimeError: Result already set

geokala:

PR #2448, opened 2020-11-16
RD-475 Add labels when creating deployments

YoniItzhak:
Updating the current rest endpoint `/deployments/DEPLOYMENT_ID` to include labels data.

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -2448,9 +2448,9 @@ class ResourceManager(object):
         """
         current_time = utils.get_formatted_timestamp()
         for label in labels_list:
-            key, value = list(label.items())[0]
+            key, value = label.popitem()
             self.sm.put(
-                models.DeploymentsLabels(
+                models.DeploymentLabel(
                     key=key.lower(),
                     value=value.lower(),
                     created_at=current_time,
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -2444,11 +2444,10 @@ class ResourceManager(object):
 
         :param deployment: A deployment object
         :param labels_list: A list of labels of the form:
-                            [{key1: value1}, {key2: value2}]
+                            [(key1, value1), (key2, value2)]
         """
         current_time = utils.get_formatted_timestamp()
-        for label in labels_list:
-            key, value = label.popitem()
+        for key, value in labels_list:
             self.sm.put(
                 models.DeploymentLabel(
                     key=key.lower(),
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -1404,7 +1404,8 @@ class ResourceManager(object):
                           bypass_maintenance=None,
                           skip_plugins_validation=False,
                           site_name=None,
-                          runtime_only_evaluation=False):
+                          runtime_only_evaluation=False,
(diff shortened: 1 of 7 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -2434,6 +2438,28 @@ def _create_deployment_initial_dependencies(self,
...
+        """
+        Populate the deployments_labels table.
+
+        :param deployment: A deployment object
+        :param labels_list: A list of labels of the form:
+                            [{key1: value1}, {key2: value2}]
+        """
+        current_time = utils.get_formatted_timestamp()
+        for label in labels_list:
+            key, value = list(label.items())[0]
```
tehasdf, line 2451:
```suggestion
            key, value = label.popitem()
```

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -93,11 +93,30 @@ def put(self, deployment_id, **kwargs):
...
+def _get_labels(request_dict):
+    if 'labels' not in request_dict:
+        return None
+
+    labels_list = request_dict['labels']
+    for label in labels_list:
+        for key, value in label.items():
+            if ((not isinstance(key, text_type)) or
+                    (not isinstance(value, text_type)) or
+                    len(list(label.items())) > 1):
```
tehasdf, line 111:
```suggestion
                    len(label) > 1):
```

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -93,11 +93,30 @@ def put(self, deployment_id, **kwargs):
...
+                raise BadParametersError(
+                    'Labels are defined as a list of dictionaries of '
+                    'the form [{<key1>: <value1>}, {<key2>: <value2>}, ...]')
```
tehasdf, line 114:
```suggestion
                raise BadParametersError(
                    'Labels must be a list of 1-entry dicts: '
                    '[{<key1>: <value1>}, {<key2>: <value2>}, ...]')
```
shorter I guess? anyway, `are defined as` just shouts "shorten me!!!" to me :P

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -93,11 +93,30 @@ def put(self, deployment_id, **kwargs):
...
-                'runtime_only_evaluation', False)
+                'runtime_only_evaluation', False),
+            labels=_get_labels(request_dict)
         )
         return deployment, 201
 
 
+def _get_labels(request_dict):
+    if 'labels' not in request_dict:
+        return None
```
tehasdf, line 104: if you `return []` here, then the signature of this function is simpler - just "returns iterable of labels", vs "returns iterable of labels or None", and then you'll be able to drop the `if` inside of resource-manager

YoniItzhak: Nice, thanks!

YoniItzhak: I went over it now and it might be a bit tricky. The default value of labels is `None` in the `resource_manager.create_deployment()` function, so removing the `if` statement might cause issues if there's a code that doesn't pass labels to the function.

tehasdf (dismissed):

tehasdf (approved):

PR #2450, opened 2020-11-16
Plugin update: cope with plugin being deleted concurrently

tehasdf:
If this update runs at the same time as deleting the plugin
(which will happen, since we delete a plugin, and at the same time
agents delete its directory and attempt to update the state to
"uninstalled"), then we will see a race between updating the state
and deleting the rows.

If update happens first, great, the row is deleted later.

If delete happens first, the .get() will throw a 404 and all's good.

If however the delete happens between the .get and the update,
then we get an error; in this case, let's confirm that the
plugin is indeed deleted, and throw a 404 as expected.

Getting a 'stale' error in other cases than the plugin being
already deleted, should never happen, and will still be a 500.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/plugins.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/plugins.py
@@ -14,6 +14,7 @@
 #  * limitations under the License.
 
 from flask_restful_swagger import swagger
+from sqlalchemy.orm.exc import StaleDataError
 from werkzeug.exceptions import BadRequest
 
 from cloudify._compat import text_type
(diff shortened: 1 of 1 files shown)


rest-service/manager_rest/rest/resources_v3_1/plugins.py
```diff
@@ -236,9 +237,15 @@ def put(self, plugin_id, **kwargs):
             manager = sm.get(
                 models.Manager, None, filters={'hostname': manager_name})
 
-        get_resource_manager().set_plugin_state(
-            plugin=plugin, manager=manager, agent=agent,
-            state=request_dict['state'], error=request_dict.get('error'))
-
-        if plugin:
-            return sm.update(plugin)
```
tehasdf, line 244: the `if` is useless because the .get() will throw a 404 and not return None, if the plugin isn't there

this `update` was useless as well, because we're not actually changing anything about the plugin itself, we're just changing the _PluginState rows (inside of the resource manager), which is a separate table

glukhman (approved):

geokala (approved):

PR #2461, opened 2020-11-19
Adding distinct to storage manager list

YoniItzhak:


rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -562,6 +566,7 @@ def list(self,
         :param get_all_results: Get all the results without the limitation of
                                 size or pagination. Use it carefully to
                                 prevent consumption of too much memory
+        :param distinct: If True, distinct ressults will return
```
geokala, line 569: I will be taking that s back, for better results.

YoniItzhak: Thanks! Changing

geokala (approved):

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -399,6 +399,9 @@ def _paginate(query, pagination, get_all_results=False):
             size = config.instance.default_page_size
             offset = 0
 
+        if distinct:
+            query = query.distinct()
```
tehasdf, line 403: that won't work because we're querying for models here, so this will generate just a `SELECT DISTINCT labels.key, labels.value, labels.created_by FROM labels` which means distinct on all fields, ie. *not* just distinct on keys, ie. it will then return all entries ever

instead, we need a distinct on keys, eg. "SELECT DISTINCT ON (labels.key) labels.key, labels.value, labels.created_by FROM labels" (well, we don't really need the other fields, but we'll get them anyway, because storage-manager works in terms of models)

to do that, we can do `query.distinct(models.DeploymentLabel.key)`, so the distinct param should be a column to distinct on, and not just a flag

note that "DISTINCT ON field" also requires an ORDER BY on that same field

tehasdf, line 403: that won't work because we're querying for models here, so this will generate just a `SELECT DISTINCT labels.key, labels.value, labels.created_by FROM labels` which means distinct on all fields, ie. *not* just distinct on keys, ie. it will then return all entries ever

instead, we need a distinct on keys, eg. "SELECT DISTINCT ON (labels.key) labels.key, labels.value, labels.created_by FROM labels" (well, we don't really need the other fields, but we'll get them anyway, because storage-manager works in terms of models)

to do that, we can do `query.distinct(models.DeploymentLabel.key)`, so the distinct param should be a column to distinct on, and not just a flag

note that "DISTINCT ON field" also requires an ORDER BY on that same field

tehasdf (changes requested):

rest-service/manager_rest/storage/storage_manager.py
tehasdf: alternatively, I do understand that this is to be used with `include=[]`, only when include contains only 1 column, then I think it would work. But then that requirement must be put in code somehow

YoniItzhak: Great comment! I'll implement it such that `distinct` would be a list of columns and this way it would cover the use-case of only one column in the `include`.

tehasdf (approved):

PR #2463, opened 2020-11-22
Handling labels tenant

YoniItzhak:
When listing the existing deployments' labels' keys, we have a problem identifying in which tenant these deployments are.
This can be solved by declaring the labeled object's model (using the labeled_model attribute) in the different labels models and filtering the tenant using it.

### DIFF
--- a/rest-service/manager_rest/storage/storage_manager.py
+++ b/rest-service/manager_rest/storage/storage_manager.py
@@ -204,7 +204,7 @@ class SQLStorageManager(object):
         ancestor who has such a relationship)
         """
         # Users/Groups etc. don't have tenants
-        if not model_class.is_resource:
+        if not (model_class.is_resource or model_class.is_label):
             return query
 
         # not used from a request handler - no relevant user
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -230,9 +230,12 @@ def _add_tenant_filter(self, query, model_class, all_tenants):
...
 
         # Match any of the applicable tenant ids or if it's a global resource
+        model = (model_class.labeled_model if model_class.is_label
+                 else model_class)
+
         tenant_filter = sql_or(
-            model_class.visibility == VisibilityState.GLOBAL,
-            model_class._tenant_id.in_(tenant_ids)
+            model.visibility == VisibilityState.GLOBAL,
+            model._tenant_id.in_(tenant_ids)
```
tehasdf, line 238: nice
but the alternative is:
in the label class, do define visibility and tenant id, like so:
```python
@declared_attr
def visibility(cls):
    return cls.labeled_model.visibility
```
then we don't need the if here. It should emit the same/equivalent sql, just different ways of writing it on the python side

which one do you think is better?

YoniItzhak: I think this idea is a lot nicer, will implement it.

tehasdf (approved):

PR #2467, opened 2020-11-23
inte-tests: reset_storage.py: cope with the new permissions engine

tehasdf:
It's now db-based and not file-based and so this script must
work like that as well.
Persist permissions between tests because that's what we've always
done.


tests/integration_tests/resources/scripts/reset_storage.py
```diff
@@ -79,8 +79,7 @@ def _add_defaults(app, amqp_manager, script_config):
     default_tenant = create_default_user_tenant_and_roles(
         admin_username=script_config['username'],
         admin_password=script_config['password'],
-        amqp_manager=amqp_manager,
-        authorization_file_path=script_config['config']['authorization']
+        amqp_manager=amqp_manager
```
geokala, line 82: You know what I'm about to say here about commas, right? <smiles>

geokala (approved):

PR #2474, opened 2020-11-25
Jenkinsfile: activate the virtualenv in each script

tehasdf:
Remember that activating a venv only sets envvars, and so it only
applies for the current shell session. It needs to be activated
again in every session.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -9,7 +9,6 @@ def vEnv(){
   echo 'creating and sourcing virtualenv'
   sh '''#!/bin/bash
     virtualenv .venv
-    source .venv/bin/activate
     '''
 }
 @Library('pipeline-shared-library') _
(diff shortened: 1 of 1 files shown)


mabuaisha (dismissed):

mabuaisha (approved):

jenkins/Jenkinsfile
```diff
@@ -7,9 +7,8 @@ def configuration = [vaultUrl: "${env.VAULT_URL}",  vaultCredentialId: 'vault-ap
 
 def vEnv(){
   echo 'creating and sourcing virtualenv'
```
geokala, line 9: Lies.

geokala: Un-lie-ified.

geokala (approved):

geokala (approved):

PR #2490, opened 2020-12-02
RD-666 Load permissions from the file during migrations

tehasdf:
When migrating the db on a machine that already has the auth.conf
file, load permissions from that file. And insert them.

### DIFF
--- a/resources/rest-service/cloudify/migrations/versions/5ce2b0cbb6f3_5_1_to_5_1_1.py
+++ b/resources/rest-service/cloudify/migrations/versions/5ce2b0cbb6f3_5_1_to_5_1_1.py
@@ -170,7 +170,7 @@ def _load_permissions(permissions_table):
                 .from_select(
                     ['name', 'role_id'],
                     select([
-                        op.inline_literal(permission),
+                        op.inline_literal(permission).label('name'),
                         roles_table.c.id
                     ])
                     .where(roles_table.c.name == op.inline_literal(role))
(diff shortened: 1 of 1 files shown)


resources/rest-service/cloudify/migrations/versions/5ce2b0cbb6f3_5_1_to_5_1_1.py
```diff
@@ -148,3 +151,29 @@ def _create_maintenance_mode_table():
...
+            data = yaml.safe_load(f)
+            permissions = data['permissions']
+    except (IOError, KeyError):
+        return
+    roles_table = table('roles', column('id'), column('name'))
+
+    for permission, roles in permissions.items():
+        for role in roles:
+            op.execute(
+                permissions_table.insert()
```
tehasdf, line 169: this renders to something similar to
```sql
INSERT INTO permissions (name, role_id) SELECT 'license_list' AS anon_1, roles.id 
FROM roles 
WHERE roles.name = 'default' 
 LIMIT 1;
```

tehasdf: and now with the label, no more anon_1:
```sql
INSERT INTO permissions (name, role_id) SELECT 'license_list' AS name, roles.id 
FROM roles 
WHERE roles.name = 'default' 
 LIMIT 1;
```

mateumann (approved):

PR #2491, opened 2020-12-02
RD-507 Fix usage collector integration tests

YoniItzhak:
This PR fixes the usage_collector integration tests by:
1. Disabling the `cron` jobs on the manager container. This way, the usage_collector cron jobs won't be activated during the tests. And besides, it doesn't make sense for a cron job to work during tests, it's better to activate the scripts manually. 

2. Changing the `customer_id`  in the container license to `MockCustomer`. Until now, if `test_license.py` ran before the usage_collector tests, the customer_id would have changed from `mock_customer` to `MockCustomer` as defined in the test's licenses. This is prevented by changing the `customer_id` in the container license to `MockCustomer`. Moreover, to keep the correct license on the container, we "reset" the license at the end of each license test. 

3. Cleaning the usage_collector log at the end of each test. The usage_collector tests use the log file for the assertion, therefore we need to clean it at the end of each test. We do so by moving the log file at the end of each test to a file with the test's name.

### DIFF
--- a/tests/integration_tests/tests/usage_collector_base.py
+++ b/tests/integration_tests/tests/usage_collector_base.py
@@ -36,9 +36,6 @@ class TestUsageCollectorBase(BaseTestCase):
         self.undeploy_application(deployment.id)
 
     def run_collector_scripts_and_assert(self, messages):
-        docker.execute(self.env.container_id, 'mkdir -p {0}'.format(LOG_PATH))
-        docker.execute(self.env.container_id, 'echo > {0}'.format(
-            join(LOG_PATH, LOG_FILE)))
         for script in COLLECTOR_SCRIPTS:
             docker.execute(self.env.container_id, '{0} {1}.py'.format(
                 MANAGER_PYTHON,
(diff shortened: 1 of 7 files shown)


tests/integration_tests/tests/usage_collector_base.py
```diff
@@ -49,9 +47,21 @@ def run_collector_scripts_and_assert(self, messages):
...
             self.env.container_id,
-            "UPDATE usage_collector SET hourly_timestamp=0, daily_timestamp=0 "
-            "WHERE id=0")
+            "UPDATE usage_collector SET hourly_timestamp=NULL, "
+            "daily_timestamp=NULL")
+
+    def clean_usage_collector_log(self):
+        # We need to clean the usage_collector log before each test, because
+        # each test uses it for asserting different values.
+        old_usage_log = join(LOG_PATH, self._testMethodName)
```
tehasdf, line 60: so you went with this after all :P

tests/integration_tests/tests/usage_collector_base.py
```diff
@@ -49,9 +47,21 @@ def run_collector_scripts_and_assert(self, messages):
...
+        test_usage_log = join(LOG_PATH, LOG_FILE)
+
+        docker.execute(self.env.container_id,
+                       'mv {0} {1}'.format(test_usage_log, old_usage_log))
```
tehasdf, line 64:
```suggestion
        self.execute_on_manager(['mv', test_usage_log, old_usage_log])
```
ditto for every usage of `docker.execute`. We made the interface on `self` for the tests to use so that they don't need to go to a lower level and know about docker at all (or the container id)

And no need to format it into a string, it will just be split again :P

tests/integration_tests/tests/usage_collector_base.py
YoniItzhak: Yep, found out that it works with pytest :)

tests/integration_tests/tests/usage_collector_base.py
YoniItzhak: Nice, I'll change it, thanks!

tehasdf (approved):

PR #2494, opened 2020-12-06
RD-691 Create filters table

YoniItzhak:


resources/rest-service/cloudify/migrations/versions/9d261e90b1f3_5_1_1_to_5_1_2.py
```diff
@@ -0,0 +1,134 @@
...
+    _blueprint_async_upload_changes()
+    _create_filters_table()
+
+
+def downgrade():
+    _revert_blueprint_async_upload_changes()
+    _drop_filters_table()
+
+
+def _blueprint_async_upload_changes():
```
tehasdf, line 42: not in this pr?

YoniItzhak: Right, @glukhman created this migration file and I used it. Will fix it to include only my functions, and when the time comes, will rebase.

tehasdf (approved):

PR #2503, opened 2020-12-08
Add a load_permissions script

tehasdf:
For loading your existing authorization.conf into the permissions
table on-demand.
Use by `/opt/manager/env/bin/python -m manager_rest.load_permissions`

### DIFF
--- a/rest-service/manager_rest/load_permissions.py
+++ b/rest-service/manager_rest/load_permissions.py
@@ -1,0 +1,105 @@
+import argparse
+import logging
+import os
+
+import yaml
+
+from manager_rest import config
+from manager_rest.flask_utils import setup_flask_app
+from manager_rest.storage.models_base import db
+
+
+def _update_roles(new_roles, keep):
+    existing_roles = {r.name: r for r in
+                      db.session.execute('select id, type, name from roles')}
+    logging.info('Found %d existing roles', len(existing_roles))
+    logging.debug('Found roles: %s', existing_roles)
+    new_roles = {r['name']: r for r in new_roles}
+    roles_to_drop = set(existing_roles) - set(new_roles)
+    roles_to_add = set(new_roles) - set(existing_roles)
+    for role in roles_to_add:
+        db.session.execute(
+            'insert into roles(name, type, description) '
+            'values (:name, :type, :description)',
+            {
+                'name': role['name'],
+                'description': role.get('description'),
+                'type': role.get('type', 'system_role')
+            }
+        )
+    logging.debug('Added roles: %s', roles_to_add)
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/load_permissions.py
```diff
@@ -0,0 +1,105 @@
...
+            'values (:name, :type, :description)',
+            {
+                'name': role['name'],
+                'description': role.get('description'),
+                'type': role.get('type', 'system_role')
+            }
+        )
+    logging.debug('Added roles: %s', roles_to_add)
+    logging.info('Inserted %d new roles', len(roles_to_add))
+    if not keep:
```
geokala, line 32: I'd probably swap this around- if keep: logging.debug('not dropping'; else
Not a major thing, I just tend to prefer to make if statements positive where possible.

rest-service/manager_rest/load_permissions.py
```diff
@@ -0,0 +1,105 @@
...
+        for r in roles_to_drop:
+            db.session.execute('delete from roles where id=:id', {'id': r.id})
```
geokala, line 34: Can this cascade in a way that causes cascastrophe? (that's such a horrible portmanteau it must remain!)

rest-service/manager_rest/load_permissions.py
```diff
@@ -0,0 +1,105 @@
...
+def _update_roles(new_roles, keep):
+    existing_roles = {r.name: r for r in
+                      db.session.execute('select id, type, name from roles')}
+    logging.info('Found %d existing roles', len(existing_roles))
+    logging.debug('Found roles: %s', existing_roles)
+    new_roles = {r['name']: r for r in new_roles}
+    roles_to_drop = set(existing_roles) - set(new_roles)
+    roles_to_add = set(new_roles) - set(existing_roles)
+    for role in roles_to_add:
+        db.session.execute(
```
geokala, line 21: Should we maybe turn off autocommit for this so that we get all or nothing, rather than partially inserted data?

geokala: Oh, or do we do that? I see you have a commit elsewhere, but I thought the db object would automatically make transactions?

tehasdf: oh, in flask-sqla, autocommit is a thing tied to requests and to the request context; it has no power here!

rest-service/manager_rest/load_permissions.py
tehasdf: no, because roles cascade to nothing, but it will break on a FK if the role happens to be used

geokala (dismissed):

mateumann (approved):

PR #2505, opened 2020-12-10
RD-743 Correctly detect use of clustered DB

geokala:


tehasdf (dismissed):

geokala:
Yeah, it helps if I refer to the attribute on the object that has it, rather than just some arbitrary non-existent variable. Who could've guessed?

tehasdf (approved):

PR #2509, opened 2020-12-11
inte-tests: make events_printer reconnect

tehasdf:
when disconnected, wait 3 seconds, and reconnect.

The thread stops when the program finishes because it's a daemon
thread.


tehasdf:
also rewriting the class to be just a function. Subclassing Thread is kinda nasty.

geokala (approved):

PR #2512, opened 2020-12-15
RD-812 Snapshots: use alembic directly

tehasdf:
Instead of going through an additional script, use alembic directly.

I don't know why was the script there in the first place. I suppose
we were tunnel-visioned on flask-migrate. And even then, flask-migrate
also has its own CLI, so... the script I guess allowed to load configuration
first, when using flask-migrate?

Anyway, alembic's env.py is going to load configuration the same way
restservice itself does, and so it will be able to handle db connections
the same way.


tehasdf:
fwiw yes, alembic also handles `downgrade -1`
```
[root@t-2 migrations]# /opt/manager/env/bin/alembic current
5ce2b0cbb6f3 (head)
[root@t-2 migrations]# /opt/manager/env/bin/alembic downgrade -1
[root@t-2 migrations]# /opt/manager/env/bin/alembic current
387fcd049efb
```

geokala (dismissed):

geokala (approved):

geokala (approved):

PR #2519, opened 2020-12-22
RD-827 Fix delete deployment with component fails

YoniItzhak:
At the end of any execution that ended in status `canceled`, `failed`, or `terminated`, the manager goes through the inter_deployment_dependencies table and evaluates each target_deployment_func to a deployment ID, and the corresponding deployment is being retrieved from the DB. The error occurs when this deployment doesn’t exist in the DB, either because it was deleted or because it wasn’t created yet. 

To solve this issue, this PR:
1. Fails silently in case the relevant deployment doesn't exist. 
2. Updates the inter deployment dependencies only when the execution ended in state `terminated`. 

Moreover, this PR adds a logging message to `storage_manager.get()` in case `fail_silently=True`. This way we will know if an element was supposed to be retrieved from the DB.

### DIFF
--- a/rest-service/manager_rest/rest/rest_utils.py
+++ b/rest-service/manager_rest/rest/rest_utils.py
@@ -548,8 +548,7 @@ class RecursiveDeploymentDependencies(object):
 
 
 def update_inter_deployment_dependencies(sm):
-    dependencies_list = sm.list(models.InterDeploymentDependencies,
-                                filters={'target_deployment': None})
+    dependencies_list = sm.list(models.InterDeploymentDependencies)
     for dependency in dependencies_list:
         if (dependency.target_deployment_func and
                 not dependency.external_target):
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -546,36 +546,42 @@ def retrieve_and_display_dependencies(self, target_id,
...
 def _update_dependency_target_deployment(sm, dependency):
-    target_deployment_id = _evaluate_target_func(dependency)
-    if target_deployment_id:
-        target_deployment_instance = sm.get(models.Deployment,
-                                            target_deployment_id)
-        dependency.target_deployment = target_deployment_instance
+    eval_target_deployment = _get_deployment_from_target_func(
+        sm, dependency.target_deployment_func, dependency.source_deployment_id)
+    if (eval_target_deployment and
+            eval_target_deployment != dependency.target_deployment):
```
YoniItzhak, line 560: @tehasdf, is this the right way to compare between elements, or should I use their IDs?

tehasdf: yeah this is fine, and with sqlalchemy i think you can even use `xx is yy`, not just equality

YoniItzhak: Thanks! This is fine even if the element is `None`, right?

tehasdf: yup

tehasdf (approved):

PR #2521, opened 2020-12-24
Restore managers and certificated tables saparately

glukhman:


workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -259,13 +260,37 @@ def dump_config_tables(self, tempdir):
...
     def restore_config_tables(self, config_path):
         new_dump_file = self._prepend_dump(config_path, [
             'delete from {0};'.format(table)
             for table in self._CONFIG_TABLES
         ])
         self._restore_dump(new_dump_file, self._db_name)
 
+    def restore_manager_tables(self, manager_tables_path):
+        new_dump_file = self._prepend_dump(manager_tables_path, [
+            'delete from {0};'.format(table)
```
geokala, line 289: Don't delete, it'll cascade- and we're only doing this when the tables will not be there anyway

geokala (approved):

PR #2547, opened 2021-01-06
RD-693 Add filters functionality

YoniItzhak:
This PR implements the main logic of filters. Meaning, the conditions added to the query based on the provider filter rules.
I would like to get a review on the way I implemented it, there might be a more efficient way.

### DIFF
--- a/rest-service/manager_rest/storage/filters.py
+++ b/rest-service/manager_rest/storage/filters.py
@@ -7,53 +7,42 @@ from manager_rest.storage.models_base import db
 def add_labels_filters_to_query(query, labels_model, labels_filters):
     query = query.join(labels_model)
     for labels_filter in labels_filters:
-        try:
-            if '!=' in labels_filter:
-                label_key, raw_label_value = labels_filter.split('!=')
-                label_value = _get_label_value(raw_label_value)
-                if isinstance(label_value, list):
-                    query = query.filter(key_not_equal_list_values(
-                        labels_model, label_key, label_value))
-                else:
-                    query = query.filter(key_not_equal_value(
-                        labels_model, label_key, label_value))
+        if '!=' in labels_filter:
+            label_key, raw_label_value = labels_filter.split('!=')
+            label_value = _get_label_value(raw_label_value)
+            if isinstance(label_value, list):
+                query = query.filter(key_not_equal_list_values(
+                    labels_model, label_key, label_value))
+            else:
+                query = query.filter(key_not_equal_value(
+                    labels_model, label_key, label_value))
 
-            elif '=' in labels_filter:
-                label_key, raw_label_value = labels_filter.split('=')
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/storage/filters.py
```diff
@@ -0,0 +1,130 @@
...
+        '<key>=[<value1>,<value2>,...], <key>!=<value>, '
+        '<key>!=[<value1>,<value2>,...], <key> is null, '
+        '<key> is not null'.format(labels_filter_value)
+    )
+
+
+def _get_label_value(raw_label_value):
+    # raw_label_value should be of the form [<value1>,<value2>,...]
+    # in order to count as a list
+    match_list = re.match(r'^\[(\S+)\]$', raw_label_value)
```
YoniItzhak, line 72: This means that this list `[b, c]` is invalid, due to the space.

rest-service/manager_rest/storage/filters.py
```diff
@@ -0,0 +1,130 @@
...
+                elif match_not_null:
+                    query = query.filter(
+                        key_exist(labels_model, match_not_null.group(1)))
+                else:
+                    _raise_bad_labels_filter(labels_filter)
+
+            else:
+                _raise_bad_labels_filter(labels_filter)
+
+        except SyntaxError:
```
tehasdf, line 46: nothing in here raises a SyntaxError i dont think, the only thing that even would, was the ast

YoniItzhak: Correct, this was a leftover.

rest-service/manager_rest/storage/filters.py
geokala: Shouldn't this be r'^\[([^\]]+)\]$'?

YoniItzhak: Right, this makes more sense, changing.

rest-service/manager_rest/test/endpoints/test_filters.py
```diff
@@ -0,0 +1,36 @@
...
+from manager_rest.storage.filters import add_labels_filters_to_query
+from manager_rest.storage.resource_models import Deployment, DeploymentLabel
+
+
+class FiltersFunctionalityTest(base_test.BaseServerTestCase):
+
+    LABELS = [{'a': 'b'}, {'a': 'z'}, {'c': 'd'}]
+    LABELS_2 = [{'a': 'b'}, {'c': 'z'}, {'e': 'f'}]
+
+    def test_filters_functionality(self):
```
geokala, line 12: Is it worth adding a test to ensure that we complain if passed an invalid label in a filter expression?

YoniItzhak: Of course, my bad, adding.

tehasdf (approved):

PR #2572, opened 2021-01-17
RD-1053 Add execution schedules DB model

glukhman:


rest-service/manager_rest/storage/resource_models.py
```diff
@@ -531,6 +533,81 @@ def resource_fields(cls):
         return fields
 
 
+class ExecutionSchedule(CreatedAtMixin, SQLResourceBase):
+    __tablename__ = 'execution_schedules'
+    next_occurrence = db.Column(UTCDateTime, nullable=True, index=True)
+    since = db.Column(UTCDateTime, nullable=True)
+    until = db.Column(UTCDateTime, nullable=True)
+    rule = db.Column(JSONString(), nullable=False)
+    slip = db.Column(db.Integer, nullable=False)
```
YoniItzhak, line 542: `slip` is described as timestamp in the design doc, so should it be an integer?

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -531,6 +533,81 @@ def resource_fields(cls):
...
+
+    def get_rrule(self, rule):
+        if 'rrule' in rule.keys():
+            return rrule.rrulestr(rule['rrule'])
+        if 'frequency' not in rule.keys():
+            if rule.get('count') == 1:
+                frequency = rrule.DAILY
+                interval = None
+            else:
+                return
```
YoniItzhak, line 581: What happens if count != 1? Why do we just return?

YoniItzhak (commented):
Nice work!

YoniItzhak (approved):

PR #2581, opened 2021-01-20
RD-1053-Fix-exec-sched-migfile

glukhman:


resources/rest-service/cloudify/migrations/versions/9d261e90b1f3_5_1_1_to_5_2.py
```diff
@@ -238,7 +238,7 @@ def create_execution_schedules_table():
         sa.Column('id', sa.Text(), nullable=True),
         sa.Column('visibility', VISIBILITY_ENUM, nullable=True),
         sa.Column('created_at', UTCDateTime(), nullable=False),
-        sa.Column('next_occurrence', UTCDateTime(), nullable=False),
+        sa.Column('next_occurrence', postgresql.TIMESTAMP(), nullable=False),
```
YoniItzhak, line 241: I think this can stay `UTCDateTime()`. I don't see anywhere that we use `postgresql.TIMESTAMP()`.

YoniItzhak (approved):

PR #2588, opened 2021-01-21
RD-1121 Deployment list: filter by group id

tehasdf:
It must be unfortunately separate because `@create_filters` has
no way to figure out the sql necessary for this.

This is the way to filter by a many-to-many. With the any.


tehasdf:
also several trivial changes, see each commit message

geokala (approved):

PR #2590, opened 2021-01-21
RD-1054 Add exec-schedule REST endpoints

glukhman:


rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+from manager_rest.storage import get_storage_manager, models
+
+
+class ExecutionSchedules(SecuredResource):
+    @swagger.operation(
+        responseClass='List[{0}]'.format(models.ExecutionSchedule.__name__),
+        nickname='list',
+        notes='Returns a list of existing execution schedules.'
+    )
+    @authorize('snapshot_list', allow_all_tenants=True)
```
tehasdf, line 29: not `snapshot_list`

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+                'force',
+                request_dict.get('force', False)),
+            'dry_run': verify_and_convert_bool(
+                'dry_run',
+                request_dict.get('dry_run', False)),
+            'queue': verify_and_convert_bool(
+                'queue',
+                request_dict.get('queue', False)),
+            'wait_after_fail': request_dict.get('wait_after_fail', 600)
+        }
```
tehasdf, line 77: instead of having the odd `# ---` comments, self-contained logic such as "figuring out execution arguments" would fit well into just a separate method

then this method would be shorter than 100000 lines perhaps

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+        parameters = request_dict.get('parameters', None)
+        if parameters is not None and parameters.__class__ is not dict:
```
tehasdf, line 79:
```suggestion
        if parameters is not None and not isinstance(parameters, dict):
```

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+            raise manager_exceptions.BadParametersError(
+                "request body's 'parameters' field must be a dict but"
+                " is of type {0}".format(parameters.__class__.__name__))
```
tehasdf, line 82:
```suggestion
                "parameters: expected dict, but got: {0}".format(parameters))
```
look how much shorter and without the `__` attrs! :P

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # --- Handle deployment
+        rm = get_resource_manager()
+        deployment_id = request_dict['deployment_id']
+        deployment = rm.sm.get(models.Deployment, deployment_id)
+        rm._validate_permitted_to_execute_global_workflow(deployment)
```
tehasdf, line 88: why the global check?

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+                    "than 1")
+            rule = {
+                'frequency': frequency,
+                'count': count,
+                'weekdays': weekdays
+            }
+
+        slip = request_dict['slip']
+        stop_on_fail = verify_and_convert_bool(
+            'stop_on_fail',  request_dict.get('stop_on_fail', False))
```
tehasdf, line 131: figuring out the rules can fit well in a separate method as well, and it would be easy to test just that method too

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+
+        now = get_formatted_timestamp()
```
tehasdf, line 133: no, just `datetime.utcnow()`. Let's stop passing strings into the timestamp columns, because eventually we will have to transition to proper timestamp columns and not text columns. Using strings in place of datetimes is super ugly

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+        return rm.sm.put(schedule), 201
+
+    @rest_decorators.marshal_with(models.ExecutionSchedule)
+    @authorize('execution_start')
+    def patch(self, schedule_id, **kwargs):
+        """Updates scheduling parameters of an existing execution schedule"""
+
+        validate_inputs({'schedule_id': schedule_id})
+        request_dict = get_json_and_verify_params()
```
tehasdf, line 155: not much point for either of those lines? the verify has no arguments, and the schedule_id, we look that up in the db anyway

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+            rule = schedule.rule.copy()
+            if 'rrule' in rule:
+                del rule['rrule']  # RRULE will be calculated anew later
+            if count:
+                rule['count'] = convert_to_int(count)
+            if frequency:
+                rule['frequency'] = self._verify_frequency(frequency)
+            if weekdays:
+                rule['weekdays'] = self._verify_weekdays(weekdays)
+            schedule.rule = rule
```
tehasdf, line 200: hopefully figuring out the rule can use the same method as in the PUT flow

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+            slip=slip,
+            workflow_id=workflow_id,
+            parameters=parameters,
+            execution_arguments=execution_arguments,
+            stop_on_fail=stop_on_fail,
+        )
+        return rm.sm.put(schedule), 201
+
+    @rest_decorators.marshal_with(models.ExecutionSchedule)
+    @authorize('execution_start')
```
tehasdf, line 150: probably a schedule-specific permission name, not `execution_start`?
if you think you'll be able to avoid creating new permissions, sorry, no dice, because the storage-manager insists on having a separate permission for every model, in this: https://github.com/cloudify-cosmo/cloudify-manager/blob/2a446b06256d76553c1ebb3ecaac227a61696e75/rest-service/manager_rest/storage/storage_manager.py#L269

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+        if rrule_string:
+            if frequency or weekdays or count:
+                raise manager_exceptions.BadParametersError(
+                    "`rrule` cannot be provided together with `frequency`, "
+                    "`weekdays` or `count`.")
+            try:
+                rrule.rrulestr(rrule_string)
+            except ValueError as e:
+                raise manager_exceptions.BadParametersError(
+                    "invalid RRULE string provided: {}".format(str(e)))
```
tehasdf, line 188:
```suggestion
                    "invalid RRULE string provided: {}".format(e))
```
str.format already calls str() on the arguments

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
tehasdf: (same for other `@authorize` calls)

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+        for fmt in ('%Y-%m-%d', '%Y-%m-%dT%H:%M:%S',
+                    '%Y-%m-%d %H:%M:%S', '%Y-%m-%d %H:%M:%S.%f',
+                    '%Y-%m-%dT%H:%M:%S.%f', '%Y-%m-%dT%H:%M:%S.%fZ'):
+            try:
+                datetime.strptime(date_str, fmt)
+                return date_str
+            except ValueError:
+                pass
+        raise manager_exceptions.BadParametersError(
+            "{} is not a legal UNIX time".format(date_str))
```
tehasdf, line 245: UNIX? :) none of these formats have to do with UNIX

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+
+    @staticmethod
+    def _verify_frequency(frequency_str):
+        if not frequency_str:
+            return
+        parsed = re.findall(r"(\d+) (seconds?|minutes?|hours?|days?|weeks?"
+                            "|months?|years?) ?(ago)?", frequency_str)
```
tehasdf, line 252: from this "ago", I recognize that this is the format we also use elsewhere, but then we should be able to use an already-existing method? or factor out this, and the other usages?
(something like execution delete?)

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+                .format(frequency_str))
+        return frequency_str
+
+    @staticmethod
+    def _verify_weekdays(weekdays):
+        if not weekdays:
+            return
+        if not isinstance(weekdays, list):
+            raise manager_exceptions.BadParametersError(
+                "weekdays should be a list, got {}".format(type(weekdays)))
```
tehasdf, line 266: better than before, but I'd just show `weekdays` here directly anyway. The user will then find the offending value easier, than just by the type.

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
```diff
@@ -0,0 +1,275 @@
...
+        weekdays_caps = set(d.upper() for d in weekdays)
+        valid_weekdays = [str(d) for d in rrule.weekdays]
```
tehasdf, line 268: no reason for this to be a list, a set would be better
then you could also `invalid = weekdays_caps - valid_weekdays` and skip the forloop

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -626,9 +625,9 @@ def compute_next_occurrence(self):
     # {'rrule': 'RRULE:FREQ=DAILY;INTERVAL=3'} -- pass RRULE directly
 
     def get_rrule(self, rule):
-        if 'rrule' in rule.keys():
+        if rule.get('rrule'):
             return rrule.rrulestr(rule['rrule'])
-        if 'frequency' not in rule.keys():
```
tehasdf, line 631: wow was it me who reviewed the one with the `.keys()`?!

tehasdf (commented):
as this has a lot of non-obvious logic related to figuring out the rules, these parts absolutely need tests

rest-service/manager_rest/rest/resources_v3_1/execution_schedules.py
glukhman: We pretty much use `created_at=get_formatted_timestamp()` across the board currently. Wouldn't changing it here but not in  all other create/update methods lead to inconsistency if we want to compare creation times?

tehasdf: it wouldn't lead to incosistencies because it's stored as a datetime type anyway in the db
going through a string just formats and parses for no reason

tehasdf (approved):

PR #2596, opened 2021-01-25
RD-1179 Fix detach agent installation failed tests

mabuaisha:
This PR added in order to fix integration tests that trying to install agent using plugin type with detach service
`cfyuser's homedir` was moved over to `/etc` because writing to `/opt` would require root access which we wanted to avoid.


mateumann:
cfyuser's homedir was moved over to `/etc` because writing to `/opt` would require root access which we wanted to avoid.

mateumann (approved):

tehasdf:
this is analogous to cloudify-cosmo/cloudify-manager-install#1056

tehasdf (approved):

PR #2601, opened 2021-01-25
RD-1162 Update deployment groups

tehasdf:
The PATCH with 'add'/'remove' approach.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/deployments.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/deployments.py
@@ -526,7 +526,7 @@ class DeploymentGroupsId(SecuredResource):
         self._set_group_deployments(sm, group, request_dict)
         return group
 
-    @authorize('deployment_group_create')
+    @authorize('deployment_group_update')
     @rest_decorators.marshal_with(models.DeploymentGroup)
     def patch(self, group_id):
         request_dict = rest_utils.get_json_and_verify_params({
--- a/rest-service/manager_rest/rest/resources_v3_1/deployments.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/deployments.py
@@ -526,6 +526,33 @@ class DeploymentGroupsId(SecuredResource):
         self._set_group_deployments(sm, group, request_dict)
         return group
 
+    @authorize('deployment_group_create')
+    @rest_decorators.marshal_with(models.DeploymentGroup)
+    def patch(self, group_id):
+        request_dict = rest_utils.get_json_and_verify_params({
+            'description': {'optional': True},
+            'blueprint_id': {'optional': True},
+            'default_inputs': {'optional': True},
+            'visibility': {'optional': True},
+            'add': {'optional': True},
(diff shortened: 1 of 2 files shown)


tehasdf:
todo unittests

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -540,18 +567,23 @@ def _set_group_attributes(self, sm, group, request_dict):
...
                 models.Blueprint, request_dict['blueprint_id'])
 
-    def _set_group_deployments(self, sm, group, request_dict):
+    def _set_group_deployments(self, sm, group, request_dict, clear=True):
         deployment_ids = request_dict.get('deployment_ids')
         if deployment_ids is not None:
             deployments = [sm.get(models.Deployment, dep_id)
                            for dep_id in deployment_ids]
-            group.deployments.clear()
+            if clear:
```
geokala, line 575: This means that if clear is true but no new deployment IDs were passed in, the group won't be cleared. Should that be the case?

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -540,18 +567,23 @@ def _set_group_attributes(self, sm, group, request_dict):
...
+                group.deployments.clear()
             for dep in deployments:
                 group.deployments.append(dep)
 
         deployment_count = len(group.deployments)
         rm = get_resource_manager()
         input_overrides = request_dict.get('inputs') or []
+        if input_overrides and not group.default_blueprint:
```
geokala, line 583: Why does the default blueprint only matter when there are input overrides?

tehasdf: `input_overrides` is the only way we actually create new deployments in this, so only then we need a blueprint

tehasdf: (see below for the deployment-creation loop)

rest-service/manager_rest/rest/resources_v3_1/deployments.py
tehasdf: indeed. The point is the converse: if deployment ids were passed in, we need both a function to replace (PUT) and append (PATCH).
If deployment ids were not passed in, nothing to do, because at most we're only appending.

Note that indeed this means PUTing an object which does not contain deployment_ids, will not clear the group. That is deliberate, there's a test that checks this (`test_update_description`), and this is because well, it's just seems more useful to not clear the group (a very destructive operation) unless _very explicitly_ asked for it

tehasdf: now that I look at it, that test doesn't actually have the assert that it suggests it would have (in the docstring). So I'm adding that assert.

geokala: OK, sounds reasonable.

geokala (approved):

PR #2608, opened 2021-01-27
RD-1055 Add exceution scheduler service to manager

glukhman:


execution-scheduler/execution_scheduler/main.py
```diff
@@ -0,0 +1,116 @@
...
+    execution_arguments['queue'] = True
+
+    execution = rm.execute_workflow(
+        deployment_id=schedule.deployment.id,
+        workflow_id=schedule.workflow_id,
+        parameters=schedule.parameters,
+        execution_creator=schedule.creator,
+        **schedule.execution_arguments
+    )
+    set_current_tenant(None)
```
tehasdf, line 91: it would be nicer to have this in a `finally`, in case something throws, so that we dont leave a tenant set; probably use a contextmanager

execution-scheduler/execution_scheduler/main.py
```diff
@@ -0,0 +1,116 @@
...
+    return execution
+
+
+def main():
+    while True:
+        check_schedules()
+        time.sleep(60)
```
tehasdf, line 98: not for right now, but i note that this means the loop won't run every 60 seconds, but 60+time_it_took_for_check_schedules

execution-scheduler/execution_scheduler/main.py
```diff
@@ -0,0 +1,116 @@
...
+        os.environ['MANAGER_REST_CONFIG_PATH'] = CONFIG_PATH
+        os.environ['MANAGER_REST_SECURITY_CONFIG_PATH'] = REST_SECURITY_PATH
+    config.instance.load_configuration()
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--logfile', default=DEFAULT_LOG_PATH,
+                        help='Path to the log file')
+    parser.add_argument('--log-level', dest='loglevel', default='INFO',
+                        help='Logging level')
+    args = parser.parse_args()
+    logging.basicConfig(level=args.get('loglevel').upper(),
```
tehasdf, line 112: .parse_args() returns a Namespace, which doesn't have a .get, so just do `args.loglevel.upper()` (ditto logfile)

execution-scheduler/setup.py
```diff
@@ -0,0 +1,17 @@
...
+    author='Cloudify',
+    author_email='cosmo-admin@cloudify.co',
+    packages=['execution_scheduler'],
+    entry_points={
+        'console_scripts': [
+            'cloudify-execution-scheduler = execution_scheduler.main:main',
+        ]
+    },
+    install_requires=[
+        'cloudify-common==5.2.0.dev1',
```
tehasdf, line 15: the scheduler doesnt use common directly, but it uses manager_rest, 
```suggestion
        'cloudify-rest-service==5.2.0.dev1',
```

execution-scheduler/setup.py
```diff
@@ -0,0 +1,17 @@
+from setuptools import setup
+
+setup(
+    name='cloudify-execution-scheduler',
+    version='5.2.0.dev1',
+    author='Cloudify',
+    author_email='cosmo-admin@cloudify.co',
+    packages=['execution_scheduler'],
```
tehasdf, line 8: use find_packages instead (`from setuptools import find_packages`) so that we dont have to change this, when we add more subpackages

execution-scheduler/setup.py
```diff
@@ -0,0 +1,17 @@
...
+    entry_points={
+        'console_scripts': [
+            'cloudify-execution-scheduler = execution_scheduler.main:main',
```
tehasdf, line 11: this will just run the `main` function, ie. NOT the `if __name__` block, and that block has some important setup

execution-scheduler/execution_scheduler/main.py
```diff
@@ -0,0 +1,116 @@
...
+from manager_rest.flask_utils import setup_flask_app
+from manager_rest.resource_manager import get_resource_manager
+from manager_rest.storage.storage_utils import (try_acquire_lock_on_table,
+                                                unlock_table)
+
+
+logger = logging.getLogger(__name__)
+BROKER_PORT_SSL = 5671
+BROKER_PORT_NO_SSL = 5672
+SCHEDULER_LOCK_BASE = 10000
```
tehasdf, line 23: add a comment about why are we even adding that

execution-scheduler/execution_scheduler/main.py
glukhman: right, so perhaps
```suggestion
def main():
    while True:
        a=datetime.now()
        check_schedules()
         b=datetime.now()
        time.sleep(60-b+a)
        ```

tehasdf (approved):

tehasdf (approved):

PR #2632, opened 2021-02-05
tests: more pika 1.1 compat

tehasdf:
Some calls, and the CA cert

### DIFF
--- a/tests/integration_tests/framework/utils.py
+++ b/tests/integration_tests/framework/utils.py
@@ -34,7 +34,7 @@ import ssl
 
 from . import constants
 
-from cloudify.utils import get_broker_ssl_cert_path, setup_logger
+from cloudify.utils import setup_logger
 from cloudify_rest_client import CloudifyClient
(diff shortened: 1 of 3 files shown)


tests/integration_tests/framework/utils.py
```diff
@@ -107,16 +107,20 @@ def create_rest_client(host, **kwargs):
...
     credentials = pika.credentials.PlainCredentials(
         username='cloudify',
         password='c10udify')
+    ssl_ctx = ssl.create_default_context(cadata=ca_data)
     return pika.BlockingConnection(
         pika.ConnectionParameters(host=host,
                                   port=5671,
-                                  ssl_options=pika.SSLOptions(
-                                      ssl.create_default_context(
-                                          cafile=get_broker_ssl_cert_path())),
```
tehasdf, line 119: we of course can't use that function, because that path only exists inside the container, and we're calling it from outside

geokala (approved):

PR #2636, opened 2021-02-07
Fix bugs in exec. scheduler service

glukhman:


mabuaisha (dismissed):

mabuaisha (approved):

glukhman:
Here, 
`slip = timedelta(minutes=schedule.slip, seconds=60)`
changed 59 to 60, because if we're unlucky enough for the service to start at X:Y:59, executions will always be skipped

PR #2651, opened 2021-02-09
Fix flaky blueprint validation integ. tests

glukhman:


glukhman:
Added a 3 sec. wait to give event logs a chance to actually write on slower machines before checking them.
That's the same as we do here: https://github.com/cloudify-cosmo/cloudify-manager/blob/master/tests/integration_tests/tests/agentless_tests/test_executions.py#L862-L867

mabuaisha (approved):

PR #2652, opened 2021-02-09
Fix flaky integ. test_status_rabbit_inactive

glukhman:


glukhman:
Added retrying instead of fixed wait: 5 sec. may not be enough for rabbit+pika to restart, but are more than needed for the other services

mabuaisha (approved):

PR #2655, opened 2021-02-09
Wait just a little bit longer for service status

mateumann:
We have experienced some flakyness in the test.  In case a machine it
was running on was under a load, 5 seconds sleep was not enough.


mateumann:
already done in https://github.com/cloudify-cosmo/cloudify-manager/pull/2653

PR #2656, opened 2021-02-09
Make exec.scheduler service handle failed and unfinished executions (…

glukhman:
…#2649)


glukhman:
- Unfinished executions: we want to avoid a queued executions pileup in a scenario where e.g. we run an exec. every 1 minute but it takes 2 min. to finish, so we're skipping runs if the previous execution is still running
- Failed executions: if the user specified the schedule to stop on fail, we disable the schedule and inform the user via the log. The user can re-enable the schedule when whatever caused the failure is fixed

tehasdf (approved):

PR #2665, opened 2021-02-11
RD-1375 plugin_update_imports script fix

mateumann:


mateumann:
In case of some ill-formatted blueprint files, import lines (YAML tokens) were not properly recognized.  As a result the blueprint with updated imports was close to a disaster.  This patch prevents the script from reading the rest of the blueprint as an import line.

tehasdf (approved):

PR #2679, opened 2021-02-16
RD-1414 upload_blueprint: update state when download fails

tehasdf:
The client.blueprints.download call can fail, in which case
the blueprint state must be updated as well. Otherwise the blueprint
is stuck in the 'uploading' state forever.


tehasdf:
that function screams for some refactor, but not today

geokala (approved):

PR #2689, opened 2021-02-17
RD-1450 Endpoint for creating Nodes

tehasdf:
Add an endpoint for batch-creating Nodes.

This is going to be useful for the "async deployment creation" track, where
we move creating nodes & node-instances out to a workflow. Then the workflow
of course needs a way to create those nodes.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/nodes.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/nodes.py
@@ -4,12 +4,10 @@ from manager_rest import manager_exceptions
 from manager_rest.rest import rest_utils
 from manager_rest.security.authorization import authorize
 from manager_rest.storage import get_storage_manager, models
-from manager_rest.rest import rest_decorators
 
 
 class Nodes(v3_Nodes):
     @authorize('node_create')
-    @rest_decorators.marshal_with(models.Node)
     def post(self):
         request_dict = rest_utils.get_json_and_verify_params({
             'nodes': {'type': list},
@@ -25,7 +23,7 @@ class Nodes(v3_Nodes):
                 node.set_deployment(deployment)
                 sm.put(node)
                 created.append(node)
-        return created, 200
+        return None, 201
 
     def _deployment_id_from_nodes(self, raw_nodes):
         deployment_ids = set()
--- a/rest-service/manager_rest/rest/resources_v3_1/nodes.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/nodes.py
@@ -14,6 +14,8 @@ class Nodes(v3_Nodes):
         })
         sm = get_storage_manager()
         raw_nodes = request_dict['nodes']
+        if not raw_nodes:
+            return None, 204
         created = []
         with sm.transaction():
             deployment_id = self._deployment_id_from_nodes(raw_nodes)
--- a/rest-service/manager_rest/rest/resources_v3_1/nodes.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/nodes.py
@@ -1,5 +1,6 @@
 from ..resources_v3 import Nodes as v3_Nodes
 
+from manager_rest import manager_exceptions
 from manager_rest.rest import rest_utils
 from manager_rest.security.authorization import authorize
 from manager_rest.storage import get_storage_manager, models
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/resources_v3_1/nodes.py
```diff
@@ -0,0 +1,75 @@
...
+            max_number_of_instances=scalable.get('max_instances') or 1,
+            host_id=raw_node.get('host_id'),
+            properties=raw_node.get('properties') or {},
+            operations=raw_node.get('operations') or {},
+            plugins=raw_node.get('plugins') or {},
+            plugins_to_install=raw_node.get('plugins_to_install'),
+            relationships=self._prepare_node_relationships(raw_node)
+        )
+
+    def _prepare_node_relationships(self, raw_node):
```
tehasdf, line 61: this is just copied from resource_manager, but I'll leave it in there as well, until RD-1360 is done.

tehasdf:
needs cloudify-cosmo/cloudify-common#667

rest-service/manager_rest/rest/resources_v3_1/nodes.py
```diff
@@ -0,0 +1,75 @@
...
+            deployment_id = self._deployment_id_from_nodes(raw_nodes)
+            deployment = sm.get(models.Deployment, deployment_id)
+            for raw_node in raw_nodes:
+                node = self._node_from_raw_node(raw_node)
+                node.set_deployment(deployment)
+                sm.put(node)
+                created.append(node)
+        return created, 200
+
+    def _deployment_id_from_nodes(self, raw_nodes):
```
geokala, line 28: Isn't this roughly equivalent to just doing raw_nodes[0]['deployment_id'], but with an effective extra implicit assertion that all raw nodes contain a deployment ID?

tehasdf: it is, you think there's a better way to do this? i think i want to have the check in some form

geokala: I think making it more explicit might be better, e.g.
deployment_id = raw_nodes[0]['deployment_id']
for node in raw_nodes:
-indent-eaten-by-github-    if node['deployment_id'] != deployment_id: raise BadThingHappened('I feel sad')
return deployment_id

rest-service/manager_rest/test/endpoints/test_nodes.py
```diff
@@ -427,14 +427,24 @@ def test_node_and_node_instance_properties(self):
 
 
 class NodesCreateTest(base_test.BaseServerTestCase):
-    def test_create_single(self):
```
geokala: Do we not want to test creating a single node as well?

tehasdf: there isn't a way to create a single node, this only implements the "create batch" endpoint, which always takes a list. A test for 1-item list isn't meaningfully different I don't think?

rest-service/manager_rest/rest/resources_v3_1/nodes.py
geokala: OK, I see what you did with this later- that makes it make more sense, so ignore this.

rest-service/manager_rest/test/endpoints/test_nodes.py
```diff
@@ -448,3 +448,21 @@ def test_create_nodes(self):
         node1 = node1[0]
         assert len(node2) == 1
         assert node1.deployment_id == 'dep1'
+
+    def test_create_different_deployments(self):
+        self.put_blueprint()
+        self.client.deployments.create('blueprint', 'dep1')
+        self.client.deployments.create('blueprint', 'dep2')
+        with self.assertRaisesRegexp(CloudifyClientError, 'deployment'):
```
geokala, line 456: Maybe assert that it contains 'same deployment'?

rest-service/manager_rest/rest/resources_v3_1/nodes.py
tehasdf: that makes for a worse error message, but OK, it's faster

geokala (approved):

geokala (approved):

PR #2690, opened 2021-02-17
Add cfyuser user and group to cloudify-rabbitmq

geokala:


tehasdf (approved):

geokala:
Not needed, the installer spec file does this already.

PR #2708, opened 2021-02-23
Inputs type casting functionality

mateumann:
In some rare (?) cases there might exist deployments with invalid type
of inputs (e.g. '20' stored where input type should be integer).  This
patch adds functionality to attempt to automatically cast types of input
values to expected (a.k.a. inputs auto-correction).


mateumann:
wrong cherry-pick-ing.  will retry

PR #2713, opened 2021-02-24
RD-1397 Allow schedules handle complex weekdays

glukhman:
This allows the user to specify complex weekday rules (via CLI or a blueprint), such as ''3mo, l-fr", which will evaluate to "run every 3rd Monday and last Friday of a month"

### DIFF
--- a/rest-service/manager_rest/test/endpoints/test_execution_schedules.py
+++ b/rest-service/manager_rest/test/endpoints/test_execution_schedules.py
@@ -134,7 +134,7 @@ class ExecutionSchedulesTestCase(BaseServerTestCase):
     def test_schedule_invalid_weekdays(self):
         self.assertRaisesRegex(
             CloudifyClientError,
-            '400: weekdays list contains invalid weekdays',
+            '400:.* invalid weekday',
             self.client.execution_schedules.create,
             'bad-weekdays', self.deployment_id, 'install',
             since=self.an_hour_from_now, frequency='4 hours',
@@ -147,7 +147,7 @@ class ExecutionSchedulesTestCase(BaseServerTestCase):
         )
         self.assertRaisesRegex(
             CloudifyClientError,
-            '400: weekdays list contains invalid weekdays',
+            '400:.* invalid weekday',
             self.client.execution_schedules.update,
             'good-weekdays', weekdays='oneday, someday'
         )
--- a/rest-service/manager_rest/utils.py
+++ b/rest-service/manager_rest/utils.py
@@ -391,7 +391,7 @@ def get_rrule(rule, since, until):
     if not rule.get('frequency'):
         if rule.get('count') == 1:
             frequency = rrule.DAILY
-            interval = None
+            interval = 0
         else:
             return
     else:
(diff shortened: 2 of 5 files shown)


rest-service/manager_rest/test/endpoints/test_execution_schedules.py
```diff
@@ -147,7 +147,7 @@ def test_schedule_invalid_weekdays(self):
         )
         self.assertRaisesRegex(
             CloudifyClientError,
-            '400: weekdays list contains invalid weekdays',
+            '400: weekdays list contains an invalid weekday',
```
tehasdf, line 150: plz stop doing this, we have a TON of very tightly coupled checking the error message, and that makes it hell to change them ever, I don't see the point of asserting that the message is exactly this, I say just `"invalid weekday"` is a good regexp for this: it asserts that the error message points to the problem

rest-service/manager_rest/utils.py
```diff
@@ -411,17 +411,47 @@ def get_rrule(rule, since, until):
...
+        until_cap = rule_set[count-1]
+    except IndexError:
+        return rule_set
+    capped_rule_set = rrule.rruleset()
+    for weekday in weekdays:
+        capped_rule_set.rrule(
+            rrule.rrule(freq=frequency, interval=interval, dtstart=since,
+                        until=until_cap, byweekday=weekday[0],
+                        bysetpos=weekday[1], cache=True))
+    return capped_rule_set
```
tehasdf, line 438: I'd split it to multiple functions at least:
1. something that encapsulates the `for weekday in weekdays` loop and returns a rruleset
2. something that calls the function from 1 twice, based on the until-cap
3. get_rrule calls the function from 2

rest-service/manager_rest/utils.py
```diff
@@ -411,17 +411,47 @@ def get_rrule(rule, since, until):
...
+
+
+def _get_weekdays(raw_weekdays):
```
tehasdf, line 441: as usual, this needs at least a docstring, because it's impossible to understand without reading the whole function, what is the argument and return values

rest-service/manager_rest/utils.py
```diff
@@ -411,17 +411,47 @@ def get_rrule(rule, since, until):
...
+    weekdays_map = {str(wd): wd for wd in rrule.weekdays}
+    weekdays = []
+    for prefixed_weekday in raw_weekdays:
+        weekday, position = prefixed_weekday[-2:], prefixed_weekday[:-2]
```
tehasdf, line 445: cute

tehasdf: but you already parse this input in the validation as well, I'd consider making a single function to parse it, and use it in both here, and the validation
```python
WEEKDAY_REGEXP = re.compile('(?<position>[1-4]|L-)(?<weekday>mo|tu|we|...)')  # i guess '|'.join or something, or can be just hardcoded, it's not like weekday names are ever gonna change
def _parse_complex_weekday(weekday):
   """Parse a string representing a weekday into a pair of (weekday, position)
   """
   match = WEEKDAY_REGEXP.match(weekday)
   if not match:
      return None
   return match.group('weekday'), match.group('position')  # I guess also int() or whatever
```

rest-service/manager_rest/utils.py
tehasdf: i imagine the 2 func would take `**rrule_kwargs` for passing in all the freq, since etc

tehasdf (dismissed):

tehasdf (approved):

PR #2716, opened 2021-02-25
RD-1538 Allow more attributes in PATCH /deps/id

tehasdf:
Set more attributes in the PATCH, so that all the values that come
from the plan, can be set in the create-dep-env workflow.

For now, replacing them is going to be disallowed.


tehasdf:
needs cloudify-cosmo/cloudify-common#684

mateumann (approved):

PR #2718, opened 2021-02-25
RD-1539 Allow passing through the handler to workflow_executor

tehasdf:
Just allow taking a handler, so that users are able to create
an amqp connection up front, and avoid making an amqp connection
per execution.

The refactor I've dreamt of isn't going to be in this :/

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -804,7 +804,8 @@ class ResourceManager(object):
                          wait_after_fail=600,
                          execution_creator=None,
                          scheduled_time=None,
-                         allow_overlapping_running_wf=False):
+                         allow_overlapping_running_wf=False,
+                         send_handler=None):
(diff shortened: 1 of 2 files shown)


tehasdf:
the bigger refactor is going to be RD-1551 - but I don't have the time for it right now

mateumann (dismissed):

rest-service/manager_rest/resource_manager.py
```diff
@@ -806,7 +810,8 @@ def execute_workflow(self,
                          wait_after_fail=600,
                          execution_creator=None,
                          scheduled_time=None,
-                         allow_overlapping_running_wf=False):
+                         allow_overlapping_running_wf=False,
+                         send_handler: 'SendHandler' = None):
```
tehasdf, line 814: yes this is the py3.6 way

mateumann (approved):

PR #2722, opened 2021-02-26
node/instances create_many: pass deployment_id separately

tehasdf:
just nicer this way


tehasdf:
needs cloudify-cosmo/cloudifycommon#687

mateumann (approved):

PR #2726, opened 2021-03-01
RD-1506 + RD-1568 Add specialized execution FKs

tehasdf:
- FK in the deployments table that points to the create-dep-env execution
- FK in the blueprints table that points to the upload-blueprint execution

Both are nullable so that snapshot-restore doesn't need any special treatment.
They'll just be null. (and old versions don't have an upload-blueprint
execution anyway)

`ondelete=SET NULL` means that when the execution is deleted, the FK is set to null.
`cascade='all, delete'` means that when the deployment is deleted (but only from the orm),
the execution is deleted as well.

### DIFF
--- a/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_5_3.py
+++ b/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_5_3.py
@@ -22,11 +22,79 @@ depends_on = None
 def upgrade():
     _create_blueprints_labels_table()
     _modify_deployments_labels_table()
+    _add_specialized_execution_fk()
 
 
 def downgrade():
+    _drop_specialized_execution_fk()
     _revert_changes_to_deployments_labels_table()
     _drop_blueprints_labels_table()
+
+
+def _add_specialized_execution_fk():
+    """Add FKs that point to special executions:
+    - the upload_blueprint execution for a blueprint
+    - the create-dep-env execution for a deployment
+    """
+    op.add_column(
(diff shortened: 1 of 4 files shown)


tehasdf:
example of the cascades 
```
cloudify_db=# select id, _upload_execution_fk from blueprints;
 id | _upload_execution_fk 
----+----------------------
 b1 |                    4
 b2 |                    5
(2 rows)

cloudify_db=# delete from executions where _storage_id = 5;   
DELETE 1
cloudify_db=# select id, _upload_execution_fk from blueprints;
 id | _upload_execution_fk 
----+----------------------
 b1 |                    4
 b2 |                     
(2 rows)

cloudify_db=# \q
[root@4e8d42ae7e5b bp]# cfy blu del b1
Deleting blueprint b1...
Blueprint deleted
[root@4e8d42ae7e5b bp]# cfy exe li
Listing all executions...

Executions:
+----+-------------+--------+------------+---------------+------------+------------+---------------+------------+-------------+------------+
| id | workflow_id | status | is_dry_run | deployment_id | created_at | started_at | scheduled_for | visibility | tenant_name | created_by |
+----+-------------+--------+------------+---------------+------------+------------+---------------+------------+-------------+------------+
+----+-------------+--------+------------+---------------+------------+------------+---------------+------------+-------------+------------+

Showing 0 of 0 executions
[root@4e8d42ae7e5b bp]# 
```

geokala (approved):

geokala (approved):

PR #2727, opened 2021-03-02
RD-1519 Modify filters value column

YoniItzhak:
This PR modifies the filters value column to be of the following form:
```
[
           {
               key: <key>,
               values: [<list of values>],
               operator: <LabelsOperator> or <AttrsOperator>,
               type: <label or attribute>
            }
]
```

For this change to take place, some changes were made to the code: 

- Adding the `filtered_resource` column to the `filters` table. Since we introduce the ability to filter by a model's attributes, we need to determine which model is being filtered. This column was added to support this. 
- Changing the `equal` and `not_equal` operators to `any_of` and `not_any_of` respectively, based on some comments that were raised during the design of the feature. 
- Splitting the `/filters` endpoint to `/filters/blueprints` and `/filters/deployments`. This was done for the same reason specified in the first bullet. 
- Modifying the filters unit-tests based on the changes. 

Important note: According to Ofer, the filters feature will not be mentioned in the docs and changelog of v5.2, so there is no need to implement the logic for backwards compatibility. 

Complementary PRs:
https://github.com/cloudify-cosmo/cloudify-common/pull/692
https://github.com/cloudify-cosmo/cloudify-premium/pull/739

### DIFF
--- a/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
+++ b/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
@@ -23,9 +23,11 @@ def upgrade():
     _create_blueprints_labels_table()
     _modify_deployments_labels_table()
     _add_specialized_execution_fk()
+    _modify_filters_table()
 
 
 def downgrade():
+    _modify_filters_table()
     _drop_specialized_execution_fk()
     _revert_changes_to_deployments_labels_table()
     _drop_blueprints_labels_table()
@@ -174,6 +176,15 @@ def _modify_deployments_labels_table():
     op.drop_column('deployments_labels', '_deployment_fk')
 
 
+def _modify_filters_table():
+    op.add_column('filters',
+                  sa.Column('filtered_resource', sa.Text(), nullable=False))
+    op.create_index(op.f('filters_filtered_resource_idx'),
+                    'filters',
+                    ['filtered_resource'],
+                    unique=False)
+
+
 def _revert_changes_to_deployments_labels_table():
     op.add_column('deployments_labels',
                   sa.Column('_deployment_fk',
@@ -216,3 +227,8 @@ def _drop_blueprints_labels_table():
     op.drop_index(op.f('blueprints_labels__creator_id_idx'),
                   table_name='blueprints_labels')
     op.drop_table('blueprints_labels')
+
+
+def _revert_changes_to_filters_table():
+    op.drop_index(op.f('filters_filtered_resource_idx'), table_name='filters')
+    op.drop_column('filters', 'filtered_resource')
--- a/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
+++ b/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
@@ -27,7 +27,7 @@ def upgrade():
 
 
 def downgrade():
-    _modify_filters_table()
+    _revert_changes_to_deployments_labels_table()
     _drop_specialized_execution_fk()
     _revert_changes_to_deployments_labels_table()
     _drop_blueprints_labels_table()
--- a/rest-service/manager_rest/constants.py
+++ b/rest-service/manager_rest/constants.py
@@ -111,11 +111,8 @@ class FilterRuleType(str, Enum):
     ATTRIBUTE = 'attribute'
 
 
-LABELS_OPERATORS = [labels_operator.value for labels_operator in
-                    LabelsOperator.__members__.values()]
+LABELS_OPERATORS = [operator.value for operator in LabelsOperator]
 
-ATTRS_OPERATORS = [attrs_operator.value for attrs_operator in
-                   AttrsOperator.__members__.values()]
+ATTRS_OPERATORS = [attrs_operator.value for attrs_operator in AttrsOperator]
 
-FILTER_RULE_TYPES = [filter_rule_type.value for filter_rule_type in
-                     FilterRuleType.__members__.values()]
+FILTER_RULE_TYPES = [rule_type.value for rule_type in FilterRuleType]
--- a/rest-service/manager_rest/constants.py
+++ b/rest-service/manager_rest/constants.py
@@ -13,6 +13,10 @@
 #  * See the License for the specific language governing permissions and
 #  * limitations under the License.
 
+
+from enum import Enum
+
+
 CONVENTION_APPLICATION_BLUEPRINT_FILE = 'blueprint.yaml'
 
 SUPPORTED_ARCHIVE_TYPES = ['zip', 'tar', 'tar.gz', 'tar.bz2']
@@ -74,11 +78,6 @@ MODELS_TO_PERMISSIONS = {
 FORBIDDEN_METHODS = ['POST', 'PATCH', 'PUT']
 SANITY_MODE_FILE_PATH = '/opt/manager/sanity_mode'
 
-EQUAL = 'equal'
-NOT_EQUAL = 'not_equal'
-IS_NULL = 'is_null'
-IS_NOT_NULL = 'is_not_null'
-
 CFY_LABELS = {'csys-obj-name',
               'csys-obj-type',
               'csys-env-type',
@@ -89,3 +88,34 @@ CFY_LABELS = {'csys-obj-name',
               'csys-obj-parent'}
 
 CFY_LABELS_PREFIX = 'csys-'
+
+
+class LabelsOperator(str, Enum):
+    ANY_OF = 'any_of'
+    NOT_ANY_OF = 'not_any_of'
+    IS_NULL = 'is_null'
+    IS_NOT_NULL = 'is_not_null'
+
+
+class AttrsOperator(str, Enum):
+    ANY_OF = 'any_of'
+    NOT_ANY_OF = 'not_any_of'
+    CONTAIN = 'contain'
+    NOT_CONTAIN = 'not_contain'
+    START_WITH = 'start_with'
+    END_WITH = 'end_with'
+
+
+class FilterRuleType(str, Enum):
+    LABEL = 'label'
+    ATTRIBUTE = 'attribute'
+
+
+LABELS_OPERATORS = [labels_operator.value for labels_operator in
+                    LabelsOperator.__members__.values()]
+
+ATTRS_OPERATORS = [attrs_operator.value for attrs_operator in
+                   AttrsOperator.__members__.values()]
+
+FILTER_RULE_TYPES = [filter_rule_type.value for filter_rule_type in
+                     FilterRuleType.__members__.values()]
--- a/rest-service/manager_rest/rest/filters_utils.py
+++ b/rest-service/manager_rest/rest/filters_utils.py
@@ -1,12 +1,12 @@
-from typing import List
+from typing import List, Union, NewType
 
 from flask import request
 
 from cloudify._compat import text_type
 
 from manager_rest.rest.rest_utils import validate_inputs
-from manager_rest.manager_exceptions import BadFilterRule, BadParametersError
 from manager_rest.storage import get_storage_manager, models
+from manager_rest.manager_exceptions import BadFilterRule, BadParametersError
 from manager_rest.constants import (ATTRS_OPERATORS,
                                     FilterRuleType,
                                     LabelsOperator,
@@ -14,7 +14,20 @@ from manager_rest.constants import (ATTRS_OPERATORS,
                                     FILTER_RULE_TYPES)
 
 
-def get_filter_rules(resource_model: models):
+class FilterRule(dict):
+    def __init__(self, key, values, operator, filter_rule_type):
+        super().__init__()
+        self['key'] = key.lower()
+        self['values'] = [value.lower() for value in values]
+        self['operator'] = operator
+        self['type'] = filter_rule_type
+
+
+FilteredModels = NewType('FilteredModels',
+                         Union[models.Deployment, models.Blueprint])
+
+
+def get_filter_rules(resource_model: FilteredModels):
     filter_rules = request.args.get('_filter_rules')
     filter_id = request.args.get('_filter_id')
 
@@ -28,7 +41,7 @@ def get_filter_rules(resource_model: models):
         )
 
     if filter_rules:
-        return parse_filter_rules(filter_rules, resource_model)
+        return create_filter_rules_list(filter_rules, resource_model)
 
     if filter_id:
         validate_inputs({'filter_id': filter_id})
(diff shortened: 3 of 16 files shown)


rest-service/manager_rest/storage/filters.py
```diff
@@ -1,42 +1,67 @@
...
+            raise BadFilterRule(filter_rule)
 
-    for label_key, label_values in labels_filters.get(EQUAL, {}).items():
-        query = query.filter(key_equal_value(
-            labels_model, label_key, label_values))
+    return query
+
+
+def add_attrs_filter_to_query(query, filter_rule):
+    return query
```
YoniItzhak, line 22: Will be modified in one of the next PRs.

rest-service/manager_rest/rest/filters_utils.py
```diff
@@ -1,129 +1,138 @@
...
-        super(BadLabelsFilter, self).__init__(
-            'The labels filter `{0}` is not in the right format. It must be '
-            'one of: <key>=<value>, <key>=[<value1>,<value2>,...], '
-            '<key>!=<value>, <key>!=[<value1>,<value2>,...], <key> is null, '
-            '<key> is not null'.format(labels_filter_value)
-        )
 
-
-def get_filter_rules():
+def get_filter_rules(resource_model: models):
```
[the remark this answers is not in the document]
tehasdf: predicting the future eh :P

resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
```diff
@@ -23,9 +23,11 @@ def upgrade():
     _create_blueprints_labels_table()
     _modify_deployments_labels_table()
     _add_specialized_execution_fk()
+    _modify_filters_table()
 
 
 def downgrade():
+    _modify_filters_table()
```
tehasdf, line 30:
```suggestion
    _revert_changes_to_filters_table()
```

rest-service/manager_rest/constants.py
```diff
@@ -89,3 +88,34 @@
               'csys-obj-parent'}
 
 CFY_LABELS_PREFIX = 'csys-'
+
+
+class LabelsOperator(str, Enum):
```
tehasdf, line 93: enum is great but why str?

rest-service/manager_rest/constants.py
```diff
@@ -89,3 +88,34 @@
...
+
+
+LABELS_OPERATORS = [labels_operator.value for labels_operator in
+                    LabelsOperator.__members__.values()]
+
+ATTRS_OPERATORS = [attrs_operator.value for attrs_operator in
+                   AttrsOperator.__members__.values()]
+
+FILTER_RULE_TYPES = [filter_rule_type.value for filter_rule_type in
+                     FilterRuleType.__members__.values()]
```
tehasdf, line 121:
```suggestion
LABELS_OPERATORS = [labels_operator.value for labels_operator in
                    LabelsOperator]

ATTRS_OPERATORS = [attrs_operator.value for attrs_operator in
                   AttrsOperator]

FILTER_RULE_TYPES = [filter_rule_type.value for filter_rule_type in
                     FilterRuleType]
```

rest-service/manager_rest/rest/filters_utils.py
tehasdf, line 17: `models` is not the right type, because that's a module. Either `Union[Blueprint, Deployment]` or just SQLResourceBase or some other base class

rest-service/manager_rest/rest/filters_utils.py
```diff
@@ -1,129 +1,138 @@
...
-           be one of: <key>=<value>, <key>=[<value1>,<value2>,...],
-           <key>!=<value>, <key>!=[<value1>,<value2>,...], <key> is null,
-           <key> is not null
-
-    :return The labels filters mapping dictionary with the following schema:
-            {NOT_EQUAL: {}, EQUAL: {}, IS_NULL: [], IS_NOT_NULL: []}
+        return filter_elem.value
+
+
+def parse_filter_rules(filter_rules: List[dict], resource_model: models):
```
tehasdf, line 39: ditto  models type

rest-service/manager_rest/rest/filters_utils.py
```diff
@@ -1,129 +1,138 @@
...
+    """Validating the filter rules list.
```
tehasdf, line 40:
```suggestion
    """Validate the filter rules list.
```

but surely this doesn't just validate, it parses?

rest-service/manager_rest/rest/filters_utils.py
```diff
@@ -1,129 +1,138 @@
...
+
+    :param filter_rules: A list of filter rules. A filter rule is a dictionary
+           of the following form:
+           {
+               key: <key>,
+               values: [<list of values>],
+               operator: <LabelsOperator> or <AttrsOperator>,
+               type: <FilterRuleType>
+            }
```
tehasdf, line 49: this is the input format, but I see that the output format is the same, so how is that parsing anything then?

rest-service/manager_rest/rest/filters_utils.py
```diff
@@ -1,129 +1,138 @@
...
+        _assert_filter_rule_structure(filter_rule)
+
+        filter_rule_key = filter_rule['key']
+        filter_rule_values = filter_rule['values']
+        filter_rule_type = filter_rule['type']
+        filter_rule_operator = filter_rule['operator']
+
+        if not isinstance(filter_rule_key, text_type):
+            raise BadFilterRule(filter_rule,
+                                'The filter rule key must be of type string')
```
tehasdf, line 64:
```suggestion
                                'The filter rule key must be a string')
```
if you want to make it 2 words shorter :P
(ditto list, 2 lines below)

rest-service/manager_rest/rest/filters_utils.py
```diff
@@ -1,129 +1,138 @@
...
+                                'The filter rule values must be of type list')
+
+        if filter_rule_type == FilterRuleType.LABEL:
+            null_operators = [LabelsOperator.IS_NULL,
+                              LabelsOperator.IS_NOT_NULL]
+            any_of_operators = [LabelsOperator.ANY_OF,
+                                LabelsOperator.NOT_ANY_OF]
+            if filter_rule_operator not in LABELS_OPERATORS:
+                raise BadFilterRule(
+                    filter_rule, f"The operator for filtering by labels must "
```
tehasdf, line 76: hmm, together with these suffixes, the message is going to be quite long

rest-service/manager_rest/constants.py
```diff
@@ -89,3 +88,34 @@
...
+    START_WITH = 'start_with'
+    END_WITH = 'end_with'
+
+
+class FilterRuleType(str, Enum):
+    LABEL = 'label'
+    ATTRIBUTE = 'attribute'
+
+
+LABELS_OPERATORS = [labels_operator.value for labels_operator in
```
glukhman, line 114: cute

rest-service/manager_rest/rest/filters_utils.py
```diff
@@ -1,129 +1,138 @@
...
+
+    return parsed_filter_rules
+
+
+def _assert_filter_rule_structure(filter_rule):
+    if not isinstance(filter_rule, dict):
+        raise BadFilterRule(filter_rule, 'The filter rule is not a dictionary')
+
+    if not (all(key in filter_rule for key in
+                ['key', 'values', 'operator', 'type'])):
```
tehasdf, line 135:
```suggestion
    if not all(key in filter_rule for key in
               ['key', 'values', 'operator', 'type']):
```

tehasdf: what about additional keys in filter_rule?
we could also do `if filter_rule.keys() != {'key', 'values', 'operator', 'type'}:`

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -395,6 +398,11 @@ def response_fields(cls):
         fields['deployment_groups'] = flask_fields.List(flask_fields.String)
         return fields
 
+    @classproperty
+    def allowed_filter_attrs(cls):
+        return ['blueprint_id', 'created_by', 'description', 'site_name',
+                'latest_execution_status']
```
tehasdf, line 404: `latest_execution_status` doesn't exist

rest-service/manager_rest/test/base_test.py
```diff
@@ -1008,15 +1009,20 @@ def put_deployment_with_labels(self, labels, resource_id=None,
         return deployment
 
     def create_filter(self, filter_name, filter_rules,
-                      visibility=VisibilityState.TENANT, client=None):
-        client = client or self.client
-        return client.filters.create(filter_name, filter_rules, visibility)
+                      visibility=VisibilityState.TENANT, filters_client=None):
+        client = filters_client or self.client
```
tehasdf, line 1013: this won't work because self.client is just the client and filters_client asks to be client.deployments_filters or client.blueprints_filters

rest-service/manager_rest/test/base_test.py
```diff
@@ -1008,15 +1009,20 @@ def put_deployment_with_labels(self, labels, resource_id=None,
...
-    def update_filter(self, new_filter_rules=None, new_visibility=None):
+    def update_filter(self, new_filter_rules=None, new_visibility=None,
+                      filters_client=None):
         filter_id = 'filter'
-        orig_filter = self.create_filter(filter_id, ['a=b'])
-        updated_filter = self.client.filters.update(
-            filter_id, new_filter_rules, new_visibility)
+        filter_rule = {'key': 'a', 'values': ['b'], 'operator': 'any_of',
+                       'type': 'label'}
+        client = filters_client or self.client
```
tehasdf, line 1021: ditto client

rest-service/manager_rest/test/endpoints/test_filters.py
```diff
@@ -2,181 +2,242 @@
...
-        err_filters = ['a null', 'a', 'a!b']
-        for err_filter in err_filters:
-            with self.assertRaisesRegex(BadLabelsFilter,
-                                        '.*not in the right format.*'):
-                create_labels_filters_mapping([err_filter])
+            [('a', ['b'], 'any_of', 'label'),
+             ('c', ['y', 'z'], 'not_any_of', 'label')], {dep1.id})
+
+    def test_filter_rule_not_dictionary_fails(self):
+        with self.assertRaisesRegex(BadFilterRule, '.*not a dictionary'):
```
tehasdf, line 51:
```suggestion
        with self.assertRaisesRegex(BadFilterRule, 'not a dictionary'):
```
assertRaisesRegex works in terms of re.search, not re.match, so `.*` at the start or at the end means nothing

rest-service/manager_rest/test/endpoints/test_filters.py
```diff
@@ -2,181 +2,242 @@
...
+            parse_filter_rules(['a'], models.Deployment)
+
+    def test_filter_rule_missing_entry_fails(self):
+        with self.assertRaisesRegex(BadFilterRule,
+                                    '.*one of the entries in the filter rule'
```
tehasdf, line 56:
```suggestion
                                    'one of the entries in the filter rule'
```

tehasdf: and tbh i'd just make it `assertRaisesRegex(BadFilterRule, 'missing')`: it's a good idea to check that the message points to the part that is wrong, but no need to check the exact wording

rest-service/manager_rest/test/endpoints/test_filters.py
```diff
@@ -2,181 +2,242 @@
...
+             '.*list must include at least one item if the operator.*'),
+            (('blueprint_id', ['b'], 'bad_operator', 'attribute'),
+             '.*The operator for filtering by attributes must be one.*'),
+            (('bad_attribute', ['dep1'], 'any_of', 'attribute'),
+             '.*Allowed attributes to filter deployments by are.*'),
+            (('a', ['b'], 'any_of', 'bad_type'),
+             '.*Filter rule type must be one of.*'),
+            (('bad_attribute', ['dep1'], 'any_of', 'bad_type'),
+             '.*Filter rule type must be one of.*')
+        ]
```
tehasdf, line 80: ditto all the `.*`

rest-service/manager_rest/rest/filters_utils.py
```diff
@@ -1,129 +1,138 @@
...
+            try:
+                validate_inputs({'filter rule key': filter_rule_key})
+                validate_inputs({'filter rule value': value},
+                                err_prefix=value_msg_prefix)
+            except BadParametersError as e:
+                err_msg = f'The filter rule {filter_rule} is invalid. '
+                raise BadParametersError(err_msg + str(e))
+            parsed_values_list.append(value.lower())
+
+        parsed_filter_rules.append({'key': filter_rule_key.lower(),
```
tehasdf, line 122: perhaps this should just be a class, rather than a dict with always the same 4 keys

I note that even in tests you have a `_get_filter_rule_from_params` that would otherwise be unnecessary

tehasdf (commented):
main point is, consider indeed organizing the filters in term of a class rather than a dict, stronger typed = nicer to work with

resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
YoniItzhak: I wonder how this one got there, thanks!

rest-service/manager_rest/constants.py
YoniItzhak: This way we can compare a string with the Enum value. This is a simple test using IPython:
```
In [3]: from enum import Enum

In [4]: class A(Enum):
   ...:     a = 'test'

In [5]: 'test' == A.a
Out[5]: False

In [6]: class A(str, Enum):
   ...:     a = 'test'

In [7]: 'test' == A.a
Out[7]: True
```

rest-service/manager_rest/constants.py
YoniItzhak: Right, thanks!

rest-service/manager_rest/rest/filters_utils.py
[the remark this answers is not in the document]
YoniItzhak: Someone needs to do it :)

YoniItzhak: RIght! Changed to `SQLResourceBase` to be more modular.

rest-service/manager_rest/rest/filters_utils.py
YoniItzhak: Well, it mainly validates, the only parsing is converting the key and values to lowercase. But I'll modify it to only validate, and the lowercase change will be moved to the filter rule class.

rest-service/manager_rest/rest/filters_utils.py
YoniItzhak: Done

rest-service/manager_rest/rest/filters_utils.py
YoniItzhak: Previous comment.

rest-service/manager_rest/rest/filters_utils.py
YoniItzhak: Right, but I think it's good to have a long error message if it provides the reason for the error and the way to fix it.

rest-service/manager_rest/rest/filters_utils.py
YoniItzhak: Yep, you're absolutely right. I'll change it to be a class.

rest-service/manager_rest/rest/filters_utils.py
YoniItzhak: Correct.

rest-service/manager_rest/storage/resource_models.py
YoniItzhak: I thought it would be worth adding it now. But I'll wait for @mabuaisha changes.

rest-service/manager_rest/test/base_test.py
YoniItzhak: You're absolutely right. Changed.

rest-service/manager_rest/test/endpoints/test_filters.py
YoniItzhak: Thanks, changed.

rest-service/manager_rest/test/endpoints/test_filters.py
YoniItzhak: Thanks, changed.

PR #2742, opened 2021-03-05
RD-1360 Move creating node-instances to the workflow

tehasdf:
Create deployment nodes and node-instances in the
`create_deployment_environment` workflow

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -205,9 +205,11 @@ pipeline {
                   vEnv()
                   sh script: '''
                     . .venv/bin/activate
-                    pip install -Ur rest-service/dev-requirements.txt
-                    pip install -Ur rest-service/test-requirements.txt
-                    pip install -e rest-service/
+                    cd rest-service
+                    pip install -Ur dev-requirements.txt
+                    pip install -Ur test-requirements.txt
+                    pip install -e .
+                    cd ..
                     pip install -e ~/cloudify-premium
                   ''', label: "installing requirements for pytest"
                   echo 'Replace the cloudify-manager URL in dev-requirements.txt with the local path to this checkout'
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -59,9 +59,10 @@ pipeline {
                 vEnv()
                 sh script: '''
                     . .venv/bin/activate
-                  pip install -Ur rest-service/dev-requirements.txt
-                  pip install -Ur rest-service/test-requirements.txt
-                  pip install -e rest-service/
+                  cd rest-service
+                  pip install -Ur dev-requirements.txt
+                  pip install -Ur test-requirements.txt
+                  pip install -e .
                 ''', label: "installing requirements for pytest"
                 echo 'Running pytest <<rest-service/manager_rest/test/endpoints>>'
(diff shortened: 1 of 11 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -1457,6 +1464,18 @@ def assert_no_snapshot_creation_running_or_queued(self):
                     ' while a `create_snapshot` workflow is running or queued'
                     ' (snapshot id: {0})'.format(e.id))
 
+    def _override_failed_deployment(self, deployment_id):
```
tehasdf, line 1467: s/override/cleanup/

jenkins/Jenkinsfile
```diff
@@ -59,9 +59,10 @@ pipeline {
                 vEnv()
                 sh script: '''
                     . .venv/bin/activate
-                  pip install -Ur rest-service/dev-requirements.txt
-                  pip install -Ur rest-service/test-requirements.txt
-                  pip install -e rest-service/
+                  cd rest-service
```
geokala, line 62: Does this change have much benefit?

rest-service/manager_rest/resource_manager.py
```diff
@@ -1499,6 +1505,23 @@ def assert_no_snapshot_creation_running_or_queued(self):
                     ' while a `create_snapshot` workflow is running or queued'
                     ' (snapshot id: {0})'.format(e.id))
 
+    def _cleanup_failed_deployment(self, deployment_id):
```
geokala, line 1508: Doesn't this mean that we can't let the user-facing side of this be fully async because the user won't be able to check the failed create_dep_env asynchronously? (obviously it could be fixed later if so)

tehasdf: note that this gets called, and so the deployment is removed, on the next attempt to create

that is to say, if you have a failed deployment, you can `cfy dep create xxx` without getting the "deployment by that name already exists"

jenkins/Jenkinsfile
tehasdf: i want to call the actual workflow in the restservice tests, so i have to install it

the alternative is copying the implementation, which seems worse. This way, the workflow is also tested.

geokala: You're still pip installing the same code though, or am I misreading?

rest-service/manager_rest/resource_manager.py
geokala: Ah, I misread- sounds good then!

jenkins/Jenkinsfile
geokala: Actually, this is neat enough anyway, not sure why I'm commenting- I'll mark this resolved.

geokala (dismissed):

jenkins/Jenkinsfile
tehasdf: the contents of reqs.txt are going to be relative to cwd, and i put `../workflows/` inside of reqs.txt, and not `rest-service/workflows/`

i've done it this way, because we already do it this way in amqp-postgres

geokala (dismissed):

geokala (approved):

PR #2748, opened 2021-03-08
RD-1520 Create post searches endpoint

YoniItzhak:
This PR adds the `POST /searches/<resource>` endpoint. The purpose of this endpoint is to use filter rules when listing deployments or blueprints. 

Complementary PRs:
https://github.com/cloudify-cosmo/cloudify-common/pull/701
https://github.com/cloudify-cosmo/cloudify-manager-install/pull/1087

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/searches.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/searches.py
@@ -1,133 +1,94 @@
-from collections import OrderedDict
+from flask import request
 
-from cloudify._compat import text_type
-
-from manager_rest import manager_exceptions
 from manager_rest.security import SecuredResource
 from manager_rest.security.authorization import authorize
 from manager_rest.storage import models, get_storage_manager
+from manager_rest.manager_exceptions import BadParametersError
 
 from ..responses_v2 import ListResponse
 from .. import rest_decorators, rest_utils
-from ..filters_utils import create_filter_rules_list
+from ..filters_utils import (create_filter_rules_list,
+                             get_filter_rule_tuple,
+                             get_filter_rules_from_filter_id)
 
 
 class ResourceSearches(SecuredResource):
-    def post(self, resource_model, _include):
-        """List resource items
-
-        The payload for this request is of the following form:
-        {
-          include = [<list of columns to include in the response>],
-          filters = {<column_name>: <}
-          A dictionary where keys are column names to
-                        filter by, and values are values applicable for those
-                        columns (or lists of such values)
-          filter_rules = [<list of filter rules>],
-          filter_id = <The ID of an existing filter>,
-          order = [{attribute: <resource attribute>, sort: <'asc' or 'desc'>}],
-          pagination = {'size': <size>, 'offset': <offset>},
-          all_tenants = <True or False>,
-          get_all_results = <True or False>
-        }
-        """
-        request_schema = {
-            'include': {'optional': True, 'type': list},
-            'filter_rules': {'optional': True, 'type': list},
-            'filter_id': {'optional': True, 'type': text_type},
-            'order': {'optional': True, 'type': list},
-            'pagination': {'optional': True, 'type': dict},
-            'all_tenants': {'optional': True, 'type': bool},
-            'get_all_results': {'optional': True, 'type': bool}
-        }
+    def post(self, resource_model, _include, filters, pagination, sort,
+             all_tenants, search, filter_id, **kwargs):
+        """List resource items"""
+        get_all_results = rest_utils.verify_and_convert_bool(
+            '_get_all_results',
+            request.args.get('_get_all_results', False)
+        )
+        request_schema = {'filter_rules': {'optional': False, 'type': list}}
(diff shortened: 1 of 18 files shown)


rest-service/manager_rest/rest/resources_v3_1/searches.py
```diff
@@ -0,0 +1,84 @@
...
+            sort=sort,
+            all_tenants=all_tenants,
+            get_all_results=get_all_results,
+            filter_rules=filter_rules
+        )
+
+        return ListResponse(items=result.items, metadata=result.metadata)
+
+
+def _get_filter_rules(raw_filter_rules, resource_model, filter_id):
```
YoniItzhak, line 42: This function, as `create_filter_rules_list`, doesn't validate the uniqueness of filter rules. I.e. the user **can** provide the same filter rule in the list (and I even created a test for it). For some reason, I'm still not convinced this is the right behavior and we might want to limit it. On the rest service side, I don't think having duplicated filter rules would cause an issue as this only means duplicated conditions in the query. @tehasdf what do you think?

tehasdf: I don't really see a reason to disallow it, we might deduplicate it

rest-service/manager_rest/test/endpoints/test_searches.py
```diff
@@ -0,0 +1,89 @@
...
+        deployments = self.client.deployments.list(
+            filter_rules=self.FILTER_RULES, filter_id=self.FILTER_ID)
+        self.assertEqual(len(deployments), 1)
+        self.assertEqual(deployments[0], dep1)
+        self.assert_metadata_filtered(deployments, 1)
+
+    def _test_list_resources_with_filter_rules_and_filter_id(self,
+                                                             filters_client,
+                                                             resource_client,
+                                                             cmp_resource):
```
glukhman, line 79: please rename `cmp_resource`. I can't understand what cmp stands for

rest-service/manager_rest/rest/resources_v3_1/searches.py
glukhman: I don't see a reason to limit it on the user side, but we can drop duplicates from the query before we send it

YoniItzhak: @tehasdf, @glukhman, thanks. Do you see any straightforward way of doing it? I came up with a method which you can see [here](https://github.com/cloudify-cosmo/cloudify-manager/pull/2748/commits/569ed810575b3275269a1dc0a025bf1300b81036#diff-90f5a7c8fde21a321436b1de427fe0c5ed49cbbddee075345edb57fa681599c2R129) and [here](https://github.com/cloudify-cosmo/cloudify-manager/pull/2748/commits/569ed810575b3275269a1dc0a025bf1300b81036#diff-53b96b3c8ef18ae614ee5f6489124011cab623cdb669ca26066e2b48dcc288aeR47), but it's not the most efficient one.

rest-service/manager_rest/rest/resources_v3_1/searches.py
```diff
@@ -0,0 +1,84 @@
...
+from manager_rest.storage import models, get_storage_manager
+
+from ..responses_v2 import ListResponse
+from .. import rest_decorators, rest_utils
+from ..filters_utils import (create_filter_rules_list,
+                             get_filter_rules_from_filter_id)
+
+
+class ResourceSearches(SecuredResource):
+    def post(self, resource_model, _include, filters, pagination, sort,
```
geokala, line 14: Why is this a post when it is not modifying state on the manager? It should be a get, surely?

rest-service/manager_rest/rest/resources_v3_1/searches.py
tehasdf: it'd have to go through a set, yeah. If we have a way to make them tuples..

rest-service/manager_rest/test/endpoints/test_searches.py
YoniItzhak: `cmp_resource` stands for `compared_resorce`.

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -796,3 +796,24 @@ def _verify_weekdays(weekdays, frequency):
                 "complex weekday expression {} can only be used with a months|"
                 "years frequency, but got {}.".format(weekday, frequency))
     return weekdays_caps
+
+
+def modify_blueprints_list_args(filters, _include):
```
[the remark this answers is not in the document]
YoniItzhak: The point of this function, and of the one below, is to serve as a point of indirection for any modification that needs to be done to the blueprints list requests' arguments. It's not specifically for the `filters` argument or the `_include`. Added a docstring to explain this.

rest-service/manager_rest/rest/resources_v3_1/searches.py
YoniItzhak: Updated the FilterRule object to be hashable.

tehasdf (approved):

PR #2754, opened 2021-03-09
RD-1488 Adding statuses and counter to sub service and envs

mabuaisha:
This PR is created in order to do the following:
1. Add sub statuses columns for both services and environments 
2. Add sub counters columns for both services and environments
3. Add support to check cyclic dependencies between deployment labels
4. Add support propagate deployment statuses in created graph
5. Add support to  propagate counters for deployment when add/remove node from grap

### DIFF
--- a/rest-service/manager_rest/rest/rest_utils.py
+++ b/rest-service/manager_rest/rest/rest_utils.py
@@ -1,4 +1,5 @@
 import re
+from collections import defaultdict, deque
 
 from ast import literal_eval
 from contextlib import contextmanager
@@ -28,6 +29,7 @@ from cloudify.models_states import (
     VisibilityState,
     BlueprintUploadState,
     ExecutionState,
+    DeploymentState,
 )
 
 from manager_rest.storage import models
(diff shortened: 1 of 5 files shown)


rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -575,6 +591,171 @@ def retrieve_and_display_dependencies(self, target_id,
...
+        :rtype Dict
+        """
+        if self.graph:
+            return
+        dependencies = self.sm.list(models.DeploymentLabelsDependencies)
+        self.graph = {}
+        for dependency in dependencies:
+            source = dependency.source_deployment_id
+            target = dependency.target_deployment_id
+            if target and target:
```
YoniItzhak, line 613: `if source and target`?

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -575,6 +591,171 @@ def retrieve_and_display_dependencies(self, target_id,
...
 
+class RecursiveDeploymentLabelsDependencies(BaseDeploymentDependencies):
+    cyclic_error_message = 'Deployments adding labels results in ' \
+                           'cyclic deployment-labels dependencies.'
+
+    def create_dependencies_graph(self):
+        """
+        Create deployment labels dependencies from the
+        `DeploymentLabelsDependencies` model where it contains dict for all
+        holds all the graph
```
YoniItzhak, line 602: Without `holds all`?

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -456,29 +470,128 @@ def latest_execution_status(self):
...
+        return deployment_status
+
+    @property
+    def is_environment(self):
+        target_key = 'csys-obj-type'
+        target_value = 'environment'
+        for label in self.labels:
+            if label.key == target_key and label.value.lower() == target_value:
+                return True
+        else:
```
YoniItzhak, line 593: No need for the `else` here

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: yes this should be source and target

rest-service/manager_rest/storage/resource_models.py
mabuaisha: yeah no need for this

PR #2766, opened 2021-03-10
RD-1648 Sync deployment creation: return errors

tehasdf:
In the synchronous deployment-create, if create-dep-env throws an error,
give it back to the user.
This is as opposed to just returning the deployment and acting as if
nothing is wrong.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/deployments.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/deployments.py
@@ -23,7 +23,7 @@ from flask_restful_swagger import swagger
 from flask_restful.reqparse import Argument
 
 from cloudify._compat import text_type
-from cloudify.models_states import VisibilityState
+from cloudify.models_states import VisibilityState, ExecutionState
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -105,8 +108,17 @@ def put(self, deployment_id, **kwargs):
...
+            if deployment.create_execution.status != ExecutionState.TERMINATED:
+                raise self._error_from_create(deployment.create_execution)
         return deployment, 201
 
+    def _error_from_create(self, execution):
+        """Map a failed create-dep-env execution to a REST error response"""
+        if execution.status != ExecutionState.FAILED or not execution.error:
+            return DeploymentEnvironmentCreationInProgressError()
+        error_message = execution.error.strip().split('\n')[-1]
+        return DeploymentCreationError(error_message)
```
tehasdf, line 120: for now just this, but in the future we'll look at the error and return some of the specialized error classes, eg. `UnknownDeploymentInputError` et al

PR #2781, opened 2021-03-15
RD-1640 execute_workflow refactor + queueing

tehasdf:
The point is the "RD-1640" commit, ie. new de-queueing mechanism.

But first, there's a LOT of refactors to be done!
Make sure to read each commit message, have fun!


rest-service/manager_rest/storage/resource_models.py
```diff
@@ -656,18 +661,123 @@ def _get_identifier_dict(self):
...
     @classproperty
     def resource_fields(cls):
         fields = super(Execution, cls).resource_fields
         fields.pop('token')
         return fields
 
+    @validates('deployment')
+    def _set_deployment(self, key, deployment):
+        self._set_parent(deployment)
+        if self.blueprint_id is None:
```
geokala, line 681: Nitpick: Any reason not to write this as: self.blueprint_id = self.blueprint_id or deployment.blueprint_id?

tehasdf: not really, but i think i prefer `is None` wherever possible. You know, performance!

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -656,18 +661,123 @@ def _get_identifier_dict(self):
...
+    def _validate_workflow_id(self, key, workflow_id):
+        if self.parameters is not None and self.deployment is not None:
+            self.merge_workflow_parameters(
+                self.parameters, self.deployment, workflow_id)
+        return workflow_id
+
+    @validates('parameters')
+    def _validate_parameters(self, key, parameters):
+        parameters = parameters or {}
+        if self.workflow_id is not None and self.deployment is not None:
```
geokala, line 698: I'm wondering if all of these checks wrapped around merge_workflow_parameters could just be rolled into merge_workflow_parameters, which might even remove the need for this _validate_parameters function entirely? (I'm not certain about that bit, haven't checked the decorator)

tehasdf: the point is, i want the main `merge` method to run only when all three are set. And these one-param-validate methods run in an indeterminate order. I tried toying with writing something more generic but I found that this was the cleanest after all

tehasdf: and, in the validate methods, the just-set parameter is passed in as an argument (not on `self` yet) while the others are possibly already set on `self`

tehasdf: moving it into the main method sounds like a plan but when I wrote it, it doesn't even save a line, so I'll just not do it instead

rest-service/manager_rest/resource_manager.py
```diff
@@ -201,38 +201,50 @@ def update_execution_status(self, execution_id, status, error):
...
+            for group in execution.execution_groups:
+                if group.id not in group_counts:
+                    group_counts[group.id] = group.concurrency -\
+                        len(group.currently_running_executions())
+        for execution in other_queued:
+            if execution.deployment in seen_deployments:
+                continue
+            if all(group_counts[g.id] > 0 for g in execution.execution_groups):
+                for group in execution.execution_groups:
+                    group_counts[group.id] -= 1
```
mateumann, line 245: why?

tehasdf: group_counts is the "how many executions can this group still run"

here, we're going to be running an execution, so we're reserving a "slot" in each of its groups

mateumann: just to clarify:
* `system_executions` do not need any slots (because are not part of any execution group?);
* `same_dep_executions` because the new execution is part of the same execution group as the finished one.
right?

tehasdf: oooh good point, same-dep needs to be under a group check as well

mateumann (approved):

rest-service/manager_rest/test/endpoints/test_executions.py
```diff
@@ -468,7 +468,8 @@ def _execution_resume_test(self, deployment, status):
             is_system_workflow=False,
             workflow_id='install',
             status=status,
-            blueprint_id=deployment.blueprint_id
+            blueprint_id=deployment.blueprint_id,
+            parameters={},
```
geokala, line 472: I thought the parameters defaulted themselves to {} if not supplied?

geokala (approved):

PR #2782, opened 2021-03-15
RD-1700 Add operation counts on the execution model

tehasdf:
Make the execution track its total and finished operation counts.
The operation endpoints deal with keeping that count up to date.


tehasdf:
this is very carefully going around the `None` part because in principle executions don't need to _have_ operations

mateumann (approved):

PR #2786, opened 2021-03-16
Rd 1485 update deployment apis

mabuaisha:
This PR created in order to update the following:
1. Add support to attach deployment(services)/environments to deployment object
2. Add support to include new extra fields as part of the deployment response related to environments
3. Add support to update deployments count for sub-deployments (environments && services) in API response

### DIFF
--- a/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
+++ b/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
@@ -80,7 +80,8 @@ def _add_deployment_sub_statuses_and_counters():
         sa.Column(
             'sub_environments_count',
             sa.Integer(),
-            nullable=True
+            nullable=False,
+            default=0,
         )
     )
     op.add_column(
@@ -101,7 +102,8 @@ def _add_deployment_sub_statuses_and_counters():
         sa.Column(
             'sub_services_count',
             sa.Integer(),
-            nullable=True
+            nullable=False,
+            default=0,
         )
     )
     op.add_column(
@@ -115,86 +117,6 @@ def _add_deployment_sub_statuses_and_counters():
                     ),
             nullable=True
         )
-    )
-
-
-def _create_deployment_labels_dependencies_table():
-    op.create_table(
-        'deployment_labels_dependencies',
-        sa.Column('_storage_id', sa.Integer(), autoincrement=True,
-                  nullable=False),
-        sa.Column('id', sa.Text(), nullable=True),
-        sa.Column('visibility', VISIBILITY_ENUM, nullable=True),
-        sa.Column('created_at', UTCDateTime(), nullable=False),
-        sa.Column('_source_deployment', sa.Integer(),
-                  nullable=False),
-        sa.Column('_target_deployment', sa.Integer(),
-                  nullable=False),
-        sa.Column('_tenant_id', sa.Integer(), nullable=False),
-        sa.Column('_creator_id', sa.Integer(), nullable=False),
-        sa.ForeignKeyConstraint(
-            ['_creator_id'], ['users.id'],
-            name=op.f('deployment_labels_dependencies__creator_id_fkey'),
-            ondelete='CASCADE'
-        ),
-        sa.ForeignKeyConstraint(
-            ['_tenant_id'], ['tenants.id'],
-            name=op.f('deployment_labels_dependencies__tenant_id_fkey'),
-            ondelete='CASCADE'
-        ),
-        sa.ForeignKeyConstraint(
-            ['_source_deployment'], ['deployments._storage_id'],
-            name=op.f(
-                'deployment_labels_dependencies__source_deployment_fkey'
-            ), ondelete='CASCADE'
-        ),
-        sa.ForeignKeyConstraint(
-            ['_target_deployment'], ['deployments._storage_id'],
-            name=op.f(
-                'deployment_labels_dependencies__target_deployment_fkey'
-            ), ondelete='CASCADE'
-        ),
-        sa.PrimaryKeyConstraint(
-            '_storage_id', name=op.f('deployment_labels_dependencies_pkey')
-        ),
-
-        sa.UniqueConstraint(
-            '_source_deployment',
-            '_target_deployment',
-            name=op.f(
-                'deployment_labels_dependencies__source_deployment_key'
-            )
-        )
-    )
-    op.create_index(
-        op.f('deployment_labels_dependencies__creator_id_idx'),
-        'deployment_labels_dependencies', ['_creator_id'], unique=False
-    )
-    op.create_index(
-        op.f('deployment_labels_dependencies__tenant_id_idx'),
-        'deployment_labels_dependencies', ['_tenant_id'], unique=False
-    )
-    op.create_index(
-        op.f('deployment_labels_dependencies_created_at_idx'),
-        'deployment_labels_dependencies', ['created_at'], unique=False
-    )
-    op.create_index(
-        op.f('deployment_labels_dependencies_id_idx'),
-        'deployment_labels_dependencies', ['id'], unique=False
-    )
-    op.create_index(
-        op.f('deployment_labels_dependencies__source_deployment_idx'),
-        'deployment_labels_dependencies', ['_source_deployment'],
-        unique=False
-    )
-    op.create_index(
-        op.f('deployment_labels_dependencies__target_deployment_idx'),
-        'deployment_labels_dependencies', ['_target_deployment'],
-        unique=False
-    )
-    op.create_index(
-        op.f('deployment_labels_dependencies_visibility_idx'),
-        'deployment_labels_dependencies', ['visibility'], unique=False
     )
 
 
--- a/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
+++ b/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
@@ -198,6 +198,78 @@ def _create_deployment_labels_dependencies_table():
     )
 
 
+def _create_deployment_labels_dependencies_table():
+    op.create_table(
+        'deployment_labels_dependencies',
+        sa.Column('_storage_id', sa.Integer(), autoincrement=True,
+                  nullable=False),
+        sa.Column('id', sa.Text(), nullable=True),
+        sa.Column('visibility', VISIBILITY_ENUM, nullable=True),
+        sa.Column('created_at', UTCDateTime(), nullable=False),
+        sa.Column('_source_deployment', sa.Integer(),
+                  nullable=False),
+        sa.Column('_target_deployment', sa.Integer(),
+                  nullable=False),
+        sa.Column('_tenant_id', sa.Integer(), nullable=False),
+        sa.Column('_creator_id', sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ['_creator_id'], ['users.id'],
+            name=op.f('deployment_labels_dependencies__creator_id_fkey'),
+            ondelete='CASCADE'
+        ),
+        sa.ForeignKeyConstraint(
+            ['_tenant_id'], ['tenants.id'],
+            name=op.f('deployment_labels_dependencies__tenant_id_fkey'),
+            ondelete='CASCADE'
+        ),
+        sa.ForeignKeyConstraint(
+            ['_source_deployment'], ['deployments._storage_id'],
+            name=op.f(
+                'deployment_labels_dependencies__source_deployment_fkey'
+            ), ondelete='CASCADE'
+        ),
+        sa.ForeignKeyConstraint(
+            ['_target_deployment'], ['deployments._storage_id'],
+            name=op.f(
+                'deployment_labels_dependencies__target_deployment_fkey'
+            ), ondelete='CASCADE'
+        ),
+        sa.PrimaryKeyConstraint(
+            '_storage_id', name=op.f('deployment_labels_dependencies_pkey')
+        ),
+    )
+    op.create_index(
+        op.f('deployment_labels_dependencies__creator_id_idx'),
+        'deployment_labels_dependencies', ['_creator_id'], unique=False
+    )
+    op.create_index(
+        op.f('deployment_labels_dependencies__tenant_id_idx'),
+        'deployment_labels_dependencies', ['_tenant_id'], unique=False
+    )
+    op.create_index(
+        op.f('deployment_labels_dependencies_created_at_idx'),
+        'deployment_labels_dependencies', ['created_at'], unique=False
+    )
+    op.create_index(
+        op.f('deployment_labels_dependencies_id_idx'),
+        'deployment_labels_dependencies', ['id'], unique=False
+    )
+    op.create_index(
+        op.f('deployment_labels_dependencies__source_deployment_idx'),
+        'deployment_labels_dependencies', ['_source_deployment'],
+        unique=False
+    )
+    op.create_index(
+        op.f('deployment_labels_dependencies__target_deployment_idx'),
+        'deployment_labels_dependencies', ['_target_deployment'],
+        unique=False
+    )
+    op.create_index(
+        op.f('deployment_labels_dependencies_visibility_idx'),
+        'deployment_labels_dependencies', ['visibility'], unique=False
+    )
+
+
 def _add_deployment_statuses():
     installation_status.create(op.get_bind())
     deployment_status.create(op.get_bind())
--- a/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
+++ b/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
@@ -237,6 +237,14 @@ def _create_deployment_labels_dependencies_table():
         sa.PrimaryKeyConstraint(
             '_storage_id', name=op.f('deployment_labels_dependencies_pkey')
         ),
+
+        sa.UniqueConstraint(
+            '_source_deployment',
+            '_target_deployment',
+            name=op.f(
+                'deployment_labels_dependencies__source_deployment_key'
+            )
+        )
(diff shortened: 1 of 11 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -56,8 +56,11 @@
                                 extract_host_agent_plugins_from_plan)
 from manager_rest.rest.rest_utils import (
     RecursiveDeploymentDependencies,
-    update_inter_deployment_dependencies, verify_blueprint_uploaded_state,
-    compute_rule_from_scheduling_params)
+    RecursiveDeploymentLabelsDependencies,
+    update_inter_deployment_dependencies,
+    verify_blueprint_uploaded_state,
+    compute_rule_from_scheduling_params
```
geokala, line 62: A trailing comma would be nice here, but not critical at all.

rest-service/manager_rest/resource_manager.py
```diff
@@ -1381,6 +1383,105 @@ def _finalize_create_deployment(self, deployment: models.Deployment):
...
+        ).items
+        _existing_parents = [_parent[0] for _parent in result]
+        for parent in parents:
+            if parent not in _existing_parents:
+                missing_parents.append(parent)
+
+        if missing_parents:
+            raise manager_exceptions.DeploymentParentNotFound(
+                'Deployment {0}: is referencing deployments'
+                ' using label `csys-obj-parent` that are not existed, '
```
geokala, line 1411: that are not existed -> that do not exist

rest-service/manager_rest/resource_manager.py
```diff
@@ -1381,6 +1383,105 @@ def _finalize_create_deployment(self, deployment: models.Deployment):
...
+                'make sure that deployments {1} are existed before create '
```
geokala, line 1412: make sure that deployments {1} are existing before create -> make sure that deployment(s) {1} exist before creating

rest-service/manager_rest/resource_manager.py
```diff
@@ -1381,6 +1383,105 @@ def _finalize_create_deployment(self, deployment: models.Deployment):
...
+                'deployment'.format(deployment_id, ','.join(missing_parents))
+            )
+
+    def _place_deployment_label_dependency(self, source, target):
+        now = utils.get_formatted_timestamp()
```
geokala, line 1417: I don't think you need to provide this- it already defaults to the current time, so probably best just to not provide it except when it's being overridden (which this isn't).

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -111,7 +170,7 @@ def put(self, deployment_id, **kwargs):
             skip_plugins_validation=skip_plugins_validation,
             site=site,
             runtime_only_evaluation=request_dict.get(
-                'runtime_only_evaluation', False),
```
geokala, line 114: Why remove the trailing comma here?

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -130,13 +189,6 @@ def put(self, deployment_id, **kwargs):
                 raise self._error_from_create(deployment.create_execution)
         return deployment, 201
 
-    def _error_from_create(self, execution):
```
geokala, line 133: Why was this moved nearer the start of the file?

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -644,57 +644,64 @@ def find_recursive_deployments(self, source_id):
...
-    def increase_deployment_counts_in_graph(self, target, source):
+    def increase_deployment_counts_in_graph(self, target_id, source):
         """
         Increase the deployment counts for target deployment that is
         referenced directly by the `source` deployment and then make sure to
         propagate that increase to all deployment that are referenced
         directly and indirectly by `target` deployment
-        :param target: Target Deployment node that we need to attach source to
-        :rtype Deployment
+        :param target_id: Target Deployment ID that we need to attach
```
geokala, line 653: I think it's better to continue to  pass the target deployment in here rather than the ID itself- we recently changed to this approach in several places.

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -644,57 +644,64 @@ def find_recursive_deployments(self, source_id):
...
+        source to
+        :rtype str
         :param source: Source deployment that reference target deployment
         :rtype Deployment
         """
-        total_services = source.sub_services_count or 0
```
geokala, line 658: This (and the following line's change) seems to just make some of the following code more complicated- is there any reason to take this approach?

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -644,57 +644,64 @@ def find_recursive_deployments(self, source_id):
...
+                self.sm.update(target)
 
-    def decrease_deployment_counts_in_graph(self, source):
+    def decrease_deployment_counts_in_graph(self, target_id, source):
         """
         Decrease the counter for each deployment that is referenced
         directly and indirectly by `source` deployment and the counts that
         need to be update are possibly for `sub_environments_count`
         & `sub_services_count`
+        :param target_id: Target Deployment ID that we need to de-attach
```
geokala, line 690: I think it'd be better to go with the approach of passing the deployment here (as noted above).

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -644,57 +644,64 @@ def find_recursive_deployments(self, source_id):
...
         :rtype Deployment
         """
-        target_group = self.find_recursive_deployments(source.id)
+        target_group = [target_id] + self.find_recursive_deployments(target_id)
         if target_group:
             total_services = source.sub_services_count or 0
             total_environments = source.sub_environments_count or 0
             if source.is_environment:
-                total_environments += 1
+                total_environments = total_environments + 1 if \
```
geokala, line 701: You don't need this extra logic I think, because you have the 'or 0' lines above.

geokala: (so you can just do total_environments += 1)

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -644,57 +644,64 @@ def find_recursive_deployments(self, source_id):
...
+                    total_environments else 1
             else:
-                total_services += 1
```
geokala, line 697: Similar to the previous comment, I don't see why this change is needed?

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -703,8 +710,12 @@ def decrease_deployment_counts_in_graph(self, source):
                 )
                 if target.sub_services_count:
                     target.sub_services_count -= total_services
+                    if not target.sub_services_count:
+                        target.sub_services_count = None
```
geokala, line 714: Out of interest, and bearing all the previous logic in mind: Why are we ever allowing these to be null values- why not instead require them to be integers, defaulting to 0- that seems like it would simplify a lot of this code.

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -703,8 +710,12 @@ def decrease_deployment_counts_in_graph(self, source):
...
                 if target.sub_environments_count:
                     target.sub_environments_count -= total_environments
+                    if not target.sub_environments_count:
```
geokala, line 717: Same note as previous comment- this should maybe not be allowing null values.

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -617,6 +631,25 @@ def make_create_environment_execution(self, **params):
...
+        return ''
+
+    @property
+    def finished_operations(self):
+        if not self.latest_execution:
+            return None
+        return self.latest_execution.finished_operations
+
+    @property
+    def total_operations(self):
```
geokala, line 648: These two _operations properties feel somewhat misleading as they only give you finished/total operations for the most recent execution. I'm unsure if this is a big issue, but it doesn't feel quite right.

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -0,0 +1,341 @@
...
+        deployment_2 = self.client.deployments.get('parent_2')
+        self.assertEqual(deployment_1.sub_services_count, 1)
+        self.assertEqual(deployment_1.sub_environments_count, 0)
+        self.assertEqual(deployment_2.sub_services_count, 1)
+        self.assertEqual(deployment_2.sub_environments_count, 0)
+
+    def test_deployment_with_invalid_parent_label(self):
+        error_message = 'Deployment invalid_label_dep:' \
+                        ' is referencing deployments using label' \
+                        ' `csys-obj-parent` that are not existed'
```
geokala, line 112: I recommend not asserting on the entire error message, but instead checking that specific sections are there, e.g. in this case you might check that the error message contained 'invalid_label_dep', 'not', 'exist'.
That allows some rejigging of error messages without triggering test failures, while still complaining if the generally important information isn't included in them.

rest-service/manager_rest/resource_manager.py
mabuaisha: done

rest-service/manager_rest/resource_manager.py
mabuaisha: done

rest-service/manager_rest/resource_manager.py
mabuaisha: done

rest-service/manager_rest/resource_manager.py
mabuaisha: @geokala The reason why I'm adding this because the time is not inserted by default unfortunately if you check other models you will notice that and if I omitted this the `created_at` is not null by default https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/storage/resource_models.py#L73 and I do not think so it set by default

rest-service/manager_rest/rest/resources_v3_1/deployments.py
mabuaisha: Bu mistake I think while was doing rebase and merge conflict but this should be fixed

rest-service/manager_rest/rest/resources_v3_1/deployments.py
mabuaisha: Well, just for formatting as all non resource methods aligned together

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: @geokala actually there is a reason for that which is to lock that deployment https://github.com/cloudify-cosmo/cloudify-manager/blob/RD-1485-update-deployment-apis/rest-service/manager_rest/rest/rest_utils.py#L669 here thats why `id` is passed instead of the object because I want to update the deployment counts and use `locking`

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: @geokala When I checked the default values for integer counter fields in all our models I see that default value is not zero so I sticked with the same approach and the reason why I'm putting this `or` is to help in summation below and prevent adding NoneType to integer. But I think I;m going to change the default value for all counters for deployment to fallback to zero instead which will make this code more cleaner

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: Answered above, the same reason for locking deployment

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: Yeah sure this is something I'm going to update now

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: sure will fix that

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: Yup will make default to zero, already left comment above

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: Sure

rest-service/manager_rest/resource_manager.py
tehasdf: if you follow your link, you'll see that it already has a default :P

rest-service/manager_rest/storage/resource_models.py
mabuaisha: Well, I added these two properties here because frontend needs to have the latest execution operation for this deployment which will make them to have these values ready to use instead of query each deployment again to get these operations because currently we exposed them,  So I think it would be better to change their names `latext_execution_ total_operations` for example so that we can tell these are for latest execution ?
@geokala

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
mabuaisha: Sure will update that

rest-service/manager_rest/resource_manager.py
mabuaisha: Ohh thats so impressing, @tehasdf this is because I'm using pycharm and it redirect me to old env that has a code with default value

resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
```diff
@@ -80,7 +80,8 @@ def _add_deployment_sub_statuses_and_counters():
         sa.Column(
             'sub_environments_count',
             sa.Integer(),
-            nullable=True
+            nullable=True,
```
geokala, line 83: Should it be nullable? With a default of 0 I think it can be made not nullable?

mabuaisha: I do not think so it should nullable if its default value is set to 0

rest-service/manager_rest/rest/resources_v3_1/deployments.py
geokala: Fair enough

rest-service/manager_rest/rest/rest_utils.py
geokala: OK, makes sense

rest-service/manager_rest/storage/resource_models.py
geokala: Yeah, I think naming them more accurately would probably be good- it's a bit more wordy, but it also says what it actually means.

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -0,0 +1,334 @@
...
+        )
+        deployment_1 = self.client.deployments.get('parent_1')
+        deployment_2 = self.client.deployments.get('parent_2')
+        self.assertEqual(deployment_1.sub_services_count, 1)
+        self.assertEqual(deployment_1.sub_environments_count, 0)
+        self.assertEqual(deployment_2.sub_services_count, 1)
+        self.assertEqual(deployment_2.sub_environments_count, 0)
+
+    def test_deployment_with_invalid_parent_label(self):
+        error_message = 'label `csys-obj-parent` that do not exist'
```
geokala, line 110: This error message should say something like "does not exist" rather than "do not exist"- did I miss it in this PR or is it elsewhere?

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -0,0 +1,334 @@
...
+                    {
+                        'csys-obj-parent': 'notexist'
+                    }
+                ],
+                resource_id='invalid_label_dep'
+            )
+
+    def test_deployment_with_valid_and_invalid_parent_labels(self):
+        self.put_deployment(deployment_id='parent_1')
+        error_message = 'label `csys-obj-parent` that do not exist'
```
geokala, line 123: "does not exist" again

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
mabuaisha: I think you miss it but will update it no worries

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
mabuaisha: fixed

rest-service/manager_rest/storage/resource_models.py
geokala: You appear to have removed the default when making them not nullable- I think that'll cause issues with restore?

mabuaisha: Sure this is back now and part of the migration file

rest-service/manager_rest/resource_manager.py
```diff
@@ -1356,6 +1358,103 @@ def _finalize_create_deployment(self, deployment: models.Deployment):
...
+        if not parents:
+            return
+        missing_parents = []
+        result = self.sm.list(
+            models.Deployment,
+            include=['id'],
+            filters={'id': lambda col: col.in_(parents)}
+        ).items
+        _existing_parents = [_parent[0] for _parent in result]
+        for parent in parents:
```
geokala, line 1379: This looks like an ideal candidate for using sets?
missing_parents = set(parents) - set(_existing_parents)

geokala: (I think that's the right way to do it anyway- you have a test for this?)

mabuaisha: Yes there is a test case of missing parents

geokala (approved):

PR #2807, opened 2021-03-19
RD-1677 Update node instances on snapshot restore

geokala:


tehasdf (approved):
note to self from the future: i was aware at the time of writing this, that this was the wrong way to do it, and the right way was to put it in the migration instead

PR #2813, opened 2021-03-22
RD-1655 dep-groups: create deployments: allow id, inputs

tehasdf:
Instead of `inputs` in the new-deployments list, allow `id`, `inputs`,
`labels`. The new deployment will be created from those, instead of
from just inputs.

Refactor the thing a bit so that it's split into separate methods.


tehasdf:
cloudify-cosmo/cloudify-common#728 is the restclient side of this

geokala (approved):

mateumann (dismissed):

mateumann (approved):

PR #2816, opened 2021-03-22
execute_workflow: queue if needed

tehasdf:
Make sure the execution status is queued, when the execution gets queued.
Otherwise, it'll never get dequeued.

Also in the other commit, rewrite the `_check` function, because it was just wrong,
and it also checked for `.deployment.id` etc for some reason, which mostly
stopped making sense after #2790


rest-service/manager_rest/resource_manager.py
```diff
@@ -902,7 +902,7 @@ def check_for_executions(self, execution, force, queue):
             and this is set, queue the execution. Otherwise, throw.
         """
         system_exec_running = self._check_for_active_system_wide_execution(
-            queue, execution)
+            execution, queue)
```
tehasdf, line 905: lol that was stupid

glukhman (approved):

YoniItzhak (approved):

PR #2820, opened 2021-03-23
RD-1603 Create schedules in create-dep-env

tehasdf:
Move creating the schedules to the workflow, so that they don't need to be
created up-front, and so that their definitions can use intrinsic functions

### DIFF
--- a/rest-service/manager_rest/test/mock_blueprint/blueprint_with_default_schedules.yaml
+++ b/rest-service/manager_rest/test/mock_blueprint/blueprint_with_default_schedules.yaml
@@ -3,6 +3,11 @@ tosca_definitions_version: cloudify_dsl_1_3
 imports:
   - cloudify/types/types.yaml
 
+inputs:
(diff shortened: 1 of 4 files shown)


rest-service/manager_rest/test/mock_blueprint/blueprint_with_default_schedules.yaml
```diff
@@ -12,6 +17,8 @@ deployment_settings:
       recurrence: 1w
       weekdays: [mo, fr]
       count: 5
+      workflow_parameters:
```
tehasdf, line 20: funnily, while parameters can use intrinsic functions indeed, other things not necessarily;  this is because parameters are declared (dsl-parser side) as a dict, and so they are considered for evaluating functions. But the other things (eg. workflow id) are just declared as a string, so it's invalid to give them a get-input.

glukhman (approved):

PR #2821, opened 2021-03-23
RD-1810 delete-dep-env: split out making the execution

tehasdf:
similar to what we did for create, also split out creating the
delete-dep-env execution, from actually running it.

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -669,21 +669,17 @@ class ResourceManager(object):
             )
 
         if not force:
-            node_instances = self.sm.list(models.NodeInstance, filters={
-                'deployment_id': deployment.id,
-            }, get_all_results=True)
             # validate either all nodes for this deployment are still
             # uninitialized or have been deleted
-            if any(node.state not in ('uninitialized', 'deleted') for node in
-                   node_instances):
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -670,41 +663,25 @@ def delete_deployment_environment(self, deployment_id,
...
-            }, get_all_results=True)
             # validate either all nodes for this deployment are still
             # uninitialized or have been deleted
-            if any(node.state not in ('uninitialized', 'deleted') for node in
-                   node_instances):
-                existing_instances = ','.join(
-                    node.id for node in node_instances
-                    if node.state not in ('uninitialized', 'deleted')
-                )
+            node_instances = self.sm.list(models.NodeInstance, filters={
```
geokala, line 674: I wonder if this should be something that can be retrieved from the deployment object itself?
deployment.node_instances?

geokala (approved):

rest-service/manager_rest/resource_manager.py
tehasdf: it would be good, but i dont know how to do it just yet. Remember that it's a dep->node->instance

additionally, this one also has the state filter on it

PR #2822, opened 2021-03-23
RD-1196 Add/remove deployments given by a group

tehasdf:
Add and remove deployments to a group, based on another group.

This gives us the "clone group" feature, because you can just create a
group, and set its deployments to ones defined by another group.

Removing isn't something that was asked for, but why not.


tehasdf:
The other parts are cloudify-cosmo/cloudify-common#730 and cloudify-cosmo/cloudify-cli#1253

PR #2826, opened 2021-03-24
RD-1487 Handle statuses propagation and delete deployment

mabuaisha:
This PR created in order to handle the following:
1. Handle statuses propagation to upper deployments and update the overall deployment status based on sub-services and sub-environments
2. Add validation on the blueprint creation and deployment when reference invalid parent inside DSL
3. Add validation on delete/stop/uninstall/update when deployments has  sub-services and sub-environments
4. Update counter for deployment when one its sub-services and sub-environments get deleted

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -133,6 +133,10 @@ class ResourceManager(object):
         dep.latest_execution = latest_execution
         dep.deployment_status = dep.evaluate_deployment_status()
         self.sm.update(dep)
+        if dep.deployment_parents:
+            graph = RecursiveDeploymentLabelsDependencies(self.sm)
+            graph.create_dependencies_graph()
+            graph.propagate_deployment_statuses(dep.id)
 
     def update_execution_status(self, execution_id, status, error):
         execution = self.sm.get(models.Execution, execution_id)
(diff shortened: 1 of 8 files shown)


rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -751,13 +759,53 @@ def propagate_deployment_statuses(self, source):
...
+                            'child':
+                                from_dependency.source_deployment_id
+                        })
+                        visited[from_dependency.source_deployment_id] = True
+        dependency_display = '  [{0}] Deployment `{1}` depends on `{2}`'
+        return '\n'.join(
+            [dependency_display.format(i + 1,
+                                       d['child'],
+                                       d['parent'])
+             for i, d in enumerate(dependencies)])
```
tehasdf, line 808:
```suggestion
            dependency_display.format(i + 1,
                                       d['child'],
                                       d['parent'])
             for i, d in enumerate(dependencies))
```
the [] do nothing other than waste memory

mabuaisha: @tehasdf Done and updated

rest-service/manager_rest/resource_manager.py
```diff
@@ -1393,12 +1428,13 @@ def _finalize_create_deployment(self, deployment: models.Deployment):
     @staticmethod
     def get_deployment_parents_from_labels(labels):
         parents = []
+        labels = labels or []
         for label, value in labels:
```
YoniItzhak, line 1432: You might to change the naming to:
`label` ->  `label_key` 
`value` -> `label_value`

rest-service/manager_rest/resource_manager.py
```diff
@@ -1408,6 +1444,10 @@ def verify_deployment_parent_labels(self, parents, deployment_id):
         ).items
         _existing_parents = [_parent[0] for _parent in result]
```
YoniItzhak, line 1445: `_parent[0]` -> `parent.id` ? 
Just for readability.

mabuaisha: no `parent.id` is available here its just normal list where each item is just string
@YoniItzhak

rest-service/manager_rest/resource_manager.py
mabuaisha: sure will change that

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -790,15 +812,40 @@ def propagate_deployment_statuses(self, source):
...
-                    target.sub_services_status = total_srv_status
-                target.deployment_status = target.evaluate_deployment_status()
-                self.sm.update(target)
+    def retrieve_and_display_dependencies(self, target):
+        self.create_dependencies_graph()
+        queue = [target.id]
+        visited = defaultdict(bool)
+        dependencies = []
+        while queue:
+            v = queue.pop()
```
geokala, line 823: I note that by doing this we can get into an infinite loop if there are recursive dependencies.

Do we guard elsewhere against that? If not, maybe guard against it here by keeping a separate list of visited nodes and just continuing if you hit a node you've already visited.

mabuaisha: @geokala we already have a visited nodes and I add an extra check after pop from the queue if that node is already visited to continue otherwise mark it as visited

geokala (approved):

geokala:
OK, so there's quite a lot here which means I've almost certainly missed /something/, but given time constraints I've greened it so that we can let the tests explode wherever they wish to.

geokala (approved):

PR #2834, opened 2021-03-29
RD-1693 Allow special chars in labels

YoniItzhak:
We should allow special characters in labels. The rules are: 

* A label key will support all alphanumeric basic chars (English only). No spaces, no special characters other than `-`, `.`, `_`.

* A label value may contain a single line of text, no newline, no tabs. It may start with any allowed character. Allowed characters will include all characters except quotes. The following are required: `-`,`_`, `.`, `’`.

* Labels should still be saved in lowercase. 

* Labels' keys and values are limited to 256 chars.

### DIFF
--- a/rest-service/manager_rest/rest/rest_utils.py
+++ b/rest-service/manager_rest/rest/rest_utils.py
@@ -1,4 +1,5 @@
 import re
+import unicodedata
 from collections import defaultdict, deque
 
 from ast import literal_eval
@@ -864,13 +865,33 @@ def get_labels_list(raw_labels_list):
             if ((not isinstance(key, text_type)) or
                     (not isinstance(value, text_type))):
                 _raise_bad_labels_list()
-            validate_inputs({'key': key, 'value': value})
-            if key.startswith(CFY_LABELS_PREFIX) and key not in CFY_LABELS:
+            parsed_key, parsed_value = _parse_label(key, value)
+            if (parsed_key.startswith(CFY_LABELS_PREFIX) and
+                    key not in CFY_LABELS):
                 _raise_labels_prefix_not_allowed()
-            labels_list.append((key.lower(), value.lower()))
+            labels_list.append((parsed_key, parsed_value))
 
     test_unique_labels(labels_list)
     return labels_list
+
+
+def _parse_label(label_key, label_value):
+    quoted_label_key = urlquote(label_key, safe='')
+    if quoted_label_key != label_key:
+        raise manager_exceptions.BadParametersError(
+            'The label\'s key {0} contains illegal characters. Only letters, '
+            'digits and the characters `-`, `.` and `_` are allowed'.format(
+                label_key)
+        )
+
+    parsed_label_value = unicodedata.normalize('NFKC', label_value).casefold()
+    if any(char in parsed_label_value for char in ['"', '\n', '\t']):
+        raise manager_exceptions.BadParametersError(
+            'The label\'s value {0} contains illegal characters. `"`, `\\n` '
+            'and `\\t` are not allowed'.format(label_value)
+        )
+
+    return label_key.casefold(), parsed_label_value
 
 
 def get_labels_from_plan(plan, labels_entry):
--- a/rest-service/manager_rest/rest/rest_utils.py
+++ b/rest-service/manager_rest/rest/rest_utils.py
@@ -48,6 +48,13 @@ states_except_private.remove('private')
 VISIBILITY_EXCEPT_PRIVATE = states_except_private
 
 
+class BadLabelsList(manager_exceptions.BadParametersError):
+    def __init__(self):
+        super().__init__(
+            'Labels must be a list of 1-entry dictionaries: '
+            '[{<key1>: <value1>}, {<key2>: [<value2>, <value3>]}, ...]')
+
+
 @contextmanager
 def skip_nested_marshalling():
     request.__skip_marshalling = True
@@ -857,7 +864,7 @@ def get_labels_list(raw_labels_list):
     labels_list = []
     for label in raw_labels_list:
         if (not isinstance(label, dict)) or len(label) != 1:
-            _raise_bad_labels_list()
+            raise BadLabelsList()
 
         [(key, raw_value)] = label.items()
         values_list = raw_value if isinstance(raw_value, list) else [raw_value]
@@ -872,7 +879,7 @@ def get_labels_list(raw_labels_list):
 def _parse_label(label_key, label_value):
     if ((not isinstance(label_key, text_type)) or
             (not isinstance(label_value, text_type))):
-        _raise_bad_labels_list()
+        raise BadLabelsList()
 
     if len(label_key) > 256 or len(label_value) > 256:
         raise manager_exceptions.BadParametersError(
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -864,15 +865,35 @@ def get_labels_list(raw_labels_list):
             if ((not isinstance(key, text_type)) or
                     (not isinstance(value, text_type))):
                 _raise_bad_labels_list()
-            validate_inputs({'key': key, 'value': value})
-            if key.startswith(CFY_LABELS_PREFIX) and key not in CFY_LABELS:
+            parsed_key, parsed_value = _parse_label(key, value)
+            if (parsed_key.startswith(CFY_LABELS_PREFIX) and
```
geokala, line 869: Why not handle this bit in the parsing?

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -864,15 +865,35 @@ def get_labels_list(raw_labels_list):
...
                 _raise_labels_prefix_not_allowed()
-            labels_list.append((key.lower(), value.lower()))
+            labels_list.append((parsed_key, parsed_value))
 
     test_unique_labels(labels_list)
     return labels_list
 
 
+def _parse_label(label_key, label_value):
+    quoted_label_key = urlquote(label_key, safe='')
```
geokala, line 879: If we want to validate that it contains only an allowed character set then we should probably validate that character set, not compare urlquoted with original.

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -864,15 +865,35 @@ def get_labels_list(raw_labels_list):
...
+    if quoted_label_key != label_key:
+        raise manager_exceptions.BadParametersError(
+            'The label\'s key {0} contains illegal characters. Only letters, '
+            'digits and the characters `-`, `.` and `_` are allowed'.format(
+                label_key)
+        )
+
+    parsed_label_value = unicodedata.normalize('NFKC', label_value).casefold()
+    if any(char in parsed_label_value for char in ['"', '\n', '\t']):
```
geokala, line 888: I'm not sure, but this might be more efficiently written as:
reserved_characters = '"\n\t'
for char in parsed_label_value:
___if char in reserved_characters:
________raise etcetc?

geokala: Ignore this, discussed separately and not to worry about.

rest-service/manager_rest/rest/rest_utils.py
YoniItzhak: Done.

rest-service/manager_rest/test/endpoints/test_deployments.py
```diff
@@ -944,6 +944,43 @@ def test_creation_failure_with_invalid_labels(self):
                                deployment_id=resource_id,
                                labels=self.INVALID_LABELS)
 
+    @attr(client_min_version=3.1,
+          client_max_version=base_test.LATEST_API_VERSION)
+    def test_creation_failure_with_invalid_label_key(self):
+        resource_id = 'i{0}'.format(uuid.uuid4())
+        err_label = [{'k ey': 'value'}]
+        error_msg = '400: .*The label\'s key {0} contains illegal ' \
+                    'characters'.format('k ey')
```
tehasdf, line 953: `.format` a literal? :P

tehasdf, line 953: This is way too exact
```suggestion
        error_msg = 'k ey.*illegal characters'
```
just this would be enough, ditto the other test

rest-service/manager_rest/test/endpoints/test_deployments.py
```diff
@@ -944,6 +944,43 @@ def test_creation_failure_with_invalid_labels(self):
...
+                                   self.put_deployment,
+                                   blueprint_file_name='blueprint.yaml',
+                                   blueprint_id=resource_id,
+                                   deployment_id=resource_id,
+                                   labels=[err_label])
+
+    @attr(client_min_version=3.1,
+          client_max_version=base_test.LATEST_API_VERSION)
+    def test_creation_success_with_special_label_value(self):
+        labels = [{'key': '&value$'}, {'key': 'val ue'}]
```
tehasdf, line 980: these values aren't THAT special.. but I'd also add a test for the normalization, ie. make a label using one way of writing some letter, and then assert that the label exists when checking via another way of writing that letter

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -861,18 +862,50 @@ def get_labels_list(raw_labels_list):
         [(key, raw_value)] = label.items()
         values_list = raw_value if isinstance(raw_value, list) else [raw_value]
         for value in values_list:
-            if ((not isinstance(key, text_type)) or
-                    (not isinstance(value, text_type))):
-                _raise_bad_labels_list()
-            validate_inputs({'key': key, 'value': value})
-            if key.startswith(CFY_LABELS_PREFIX) and key not in CFY_LABELS:
```
tehasdf, line 868: aha, so there was a bug before, this was checked before the `.lower()`

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -861,18 +862,50 @@ def get_labels_list(raw_labels_list):
...
+            labels_list.append((parsed_key, parsed_value))
 
     test_unique_labels(labels_list)
     return labels_list
 
 
+def _parse_label(label_key, label_value):
+    if ((not isinstance(label_key, text_type)) or
+            (not isinstance(label_value, text_type))):
+        _raise_bad_labels_list()
```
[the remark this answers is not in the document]
YoniItzhak: Agreed, I created a new class named `BadLabelsList(manager_exceptions.BadParametersError)` and raise it in case of an error.

rest-service/manager_rest/test/endpoints/test_deployments.py
YoniItzhak: You're absolutely right. Fixed.

YoniItzhak: Right.

rest-service/manager_rest/rest/rest_utils.py
YoniItzhak: You're absolutely right, I realized it while working on this task. You could put a `csys-` label id you used uppercase letters, but now you can't :)

rest-service/manager_rest/test/endpoints/test_deployments.py
YoniItzhak: Right, added.

tehasdf (approved):

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -856,23 +864,56 @@ def get_labels_list(raw_labels_list):
...
 
 
+def _parse_label(label_key, label_value):
+    if ((not isinstance(label_key, text_type)) or
+            (not isinstance(label_value, text_type))):
+        raise BadLabelsList()
+
+    if len(label_key) > 256 or len(label_value) > 256:
+        raise manager_exceptions.BadParametersError(
+            'The label\'s key or value is too long. Maximum allowed length is '
```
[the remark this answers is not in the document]
YoniItzhak: Thanks :)

PR #2836, opened 2021-03-30
RD-1909 Fix deployment creation with invalid labels

YoniItzhak:
The labels validation was done after the deployment was inserted into the DB.
This caused an issue where the deployment was created, even though invalid labels were provided.

The fix for this issue is to put the labels validation before the deployment creation.


tehasdf (approved):

YoniItzhak:
The labels validation was done after the deployment was inserted into the DB.
This caused an issue where the deployment was created, even though invalid labels were provided.
The fix for this issue is to put the labels validation before the deployment creation.

PR #2839, opened 2021-03-30
RD-1317 Apply group labels to deployments

tehasdf:
Copy group labels to the deployments, when the label is added.
Also remove it from the deployments, when the label is removed
from the group.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/deployments.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/deployments.py
@@ -684,10 +684,14 @@ class DeploymentGroupsId(SecuredResource):
                 sm.delete(dep_label)
 
     def _add_group_deployments(self, sm, group, request_dict):
+        rm = get_resource_manager()
+        group_labels = [(l.key, l.value) for l in group.labels]
+
         deployment_ids = request_dict.get('deployment_ids')
         if deployment_ids is not None:
             deployments = [sm.get(models.Deployment, dep_id)
                            for dep_id in deployment_ids]
+            self._create_deployments_labels(sm, rm, deployments, group_labels)
             for dep in deployments:
                 group.deployments.append(dep)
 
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -575,21 +575,23 @@ def put(self, group_id):
...
-            group = sm.get(models.DeploymentGroup, group_id)
-        except manager_exceptions.NotFoundError:
-            group = models.DeploymentGroup(
-                id=group_id,
-                description=request_dict.get('description'),
-                created_at=datetime.now()
-            )
         with sm.transaction():
+            try:
+                group = sm.get(models.DeploymentGroup, group_id)
```
geokala, line 580: Do we deliberately not update the description when performing a put on a deployment group?

tehasdf: we do, in `_set_group_attributes`. You're right, no point doing it twice.

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -99,7 +99,8 @@ def transaction(self):
 
         self._in_transaction = True
         try:
-            yield
+            with db.session.no_autoflush:
```
geokala, line 102: Out of interest, why this change?

rest-service/manager_rest/test/endpoints/test_deployment_groups.py
```diff
@@ -437,6 +437,86 @@ def test_group_labels_for_deployments(self):
...
+    def test_add_labels_already_exist(self):
+        self.client.deployments.update_labels('dep1', [{'label2': 'value2'}])
+        self.client.deployment_groups.put(
+            'group1',
+            deployment_ids=['dep1'],
+        )
+        self.client.deployment_groups.put(  # doesn't throw
+            'group1',
+            labels=[{'label2': 'value2'}],
+        )
```
geokala, line 486: Is it worth asserting on the what labels the deployment has here- a cheap operation but should alert us if we do something like duplicate labels with the same value?

geokala (dismissed):

rest-service/manager_rest/test/endpoints/test_deployment_groups.py
tehasdf: there's an unique index on them, but why not!

geokala (dismissed):

geokala (approved):

PR #2843, opened 2021-03-31
RD-1492 Update deployment update api to handle environment changes

mabuaisha:
This PR created in order to add deployments to their parents if there are new labels get added during deployment update

### DIFF
--- a/rest-service/manager_rest/deployment_update/manager.py
+++ b/rest-service/manager_rest/deployment_update/manager.py
@@ -250,9 +250,6 @@ class DeploymentUpdateManager(object):
             include_rel_order=True)
         dep_update.central_plugins_to_install = central_plugins_to_install
         dep_update.central_plugins_to_uninstall = central_plugins_to_uninstall
-        self.sm.update(dep_update)
-
-        # If this is a preview, no need to run workflow and update DB
         deployment = self.sm.get(models.Deployment, dep_update.deployment_id)
         labels_to_create = self._get_deployment_labels_to_create(dep_update)
         parents_labels = []
@@ -263,6 +260,8 @@ class DeploymentUpdateManager(object):
             rm.verify_deployment_parent_labels(
                 parents_labels, deployment.id
             )
+        self.sm.update(dep_update)
+        # If this is a preview, no need to run workflow and update DB
         if dep_update.preview:
             dep_update.state = STATES.PREVIEW
             dep_update.id = None
--- a/rest-service/manager_rest/deployment_update/manager.py
+++ b/rest-service/manager_rest/deployment_update/manager.py
@@ -45,11 +45,14 @@ from manager_rest.deployment_update.handlers import (
     DeploymentUpdateNodeInstanceHandler)
 from manager_rest.utils import get_formatted_timestamp
 
-from manager_rest.rest.rest_utils import (get_deployment_plan,
-                                          get_labels_from_plan,
-                                          get_parsed_deployment,
-                                          RecursiveDeploymentDependencies,
-                                          verify_blueprint_uploaded_state)
+from manager_rest.rest.rest_utils import (
+    get_deployment_plan,
+    get_labels_from_plan,
+    get_parsed_deployment,
+    RecursiveDeploymentDependencies,
+    RecursiveDeploymentLabelsDependencies,
+    verify_blueprint_uploaded_state,
+)
 
 
 class DeploymentUpdateManager(object):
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -251,6 +253,8 @@ def commit_deployment_update(self,
         self.sm.update(dep_update)
 
         # If this is a preview, no need to run workflow and update DB
+        deployment = self.sm.get(models.Deployment, dep_update.deployment_id)
+        labels_to_create = self._get_deployment_labels_to_create(dep_update)
```
YoniItzhak, line 257: Any reason why you changed the location of these variables?

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -295,7 +299,15 @@ def commit_deployment_update(self,
         )
 
         # Update deployment attributes in the storage manager
-        deployment = self.sm.get(models.Deployment, dep_update.deployment_id)
+        parents_labels = []
+        if labels_to_create:
```
YoniItzhak, line 303: This validation should be done before we check for preview https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/deployment_update/manager.py#L254. 

Otherwise, the user might get a "happy" preview and then a failure in the parents labels verification.

rest-service/manager_rest/deployment_update/manager.py
```diff
@@ -325,7 +336,15 @@ def commit_deployment_update(self,
 
         rm.create_resource_labels(models.DeploymentLabel, deployment,
                                   labels_to_create)
-
+        if parents_labels:
```
YoniItzhak, line 339: What this part is for?

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -257,6 +257,42 @@ def test_number_of_total_environments_deployed_inside_environment(self):
...
+
+        self.client.deployment_updates.update_with_existing_blueprint(
+            'sub_srv', blueprint_id='update_sub_srv'
+        )
+        self.client.deployment_updates.update_with_existing_blueprint(
+            'sub_env', blueprint_id='update_sub_env'
+        )
+        deployment = self.client.deployments.get('env')
+        deployment_1 = self.client.deployments.get('env_1')
+        self.assertEqual(deployment.sub_services_count, 1)
```
YoniItzhak, line 291: Don't you also want to check this value before the update? Just to make sure it was really updated.

rest-service/manager_rest/deployment_update/manager.py
mabuaisha: yes, because I want to have the reference  for deployment so that I can use it when verify deployment label parent id but I think I can drop the verification because its already happened on the blueprint upload

rest-service/manager_rest/deployment_update/manager.py
mabuaisha: This is in order to increase the deployment counters if `csys-obj-parent` is added as part of deployment update to that deployment so that we can increase the number of deployment for this parent since it has now extra deployment attached to it

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
mabuaisha: yup will add that as well no worries

rest-service/manager_rest/deployment_update/manager.py
mabuaisha: I think that validation is not needed since the validation happened on blueprint upload

rest-service/manager_rest/deployment_update/manager.py
YoniItzhak: But then, if an invalid list of labels was provided, the line `self.sm.update(dep_update)` will still be called, and I don't think we want that.

mabuaisha: Aha yeah that is a valid point then I will keep it and make it before preview

YoniItzhak: It is supposed to be before this line https://github.com/cloudify-cosmo/cloudify-manager/pull/2843/files#diff-e62ae4f438aaf3d7e37135a0065891a2814a47c4ffa74fa9077cfe049eaa3b08R253

mabuaisha: I think even before, there is another update before it

YoniItzhak: I don't think so. This line defines that the current status is `STATES.UPDATING`, and it might be used afterwards if something failed.

YoniItzhak (approved):

glukhman (approved):

PR #2857, opened 2021-04-01
Use target branch as default instead of master

geokala:


geokala:
This should mean that (for future releases and when backported), we use the target branch (e.g. 5.3.1-build) instead of master when creating a PR against an older branch such as when preparing for a patch release.

It won't handle the pip dependencies, though we could probably look at doing that with a little more code.

geokala:

PR #2865, opened 2021-04-04
RD-1950 Add support to update counters for deployment when deployment…

mabuaisha:
This PR created in order to fix issue when we changed the type of deployment by adding/removing `csys-obj-type` where the number of counters for parents deployments of the deployment that we delete/add `csys-obj-type` should be updated so that we can have the right counters

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -1399,16 +1399,16 @@ class ResourceManager(object):
         return parents
 
     def get_deployment_object_types_from_labels(self, deployment, labels):
-        new_labels = self.get_labels_to_create(deployment, labels)
         created_types = set()
         delete_types = set()
+        new_labels = self.get_labels_to_create(deployment, labels)
         for key, value in new_labels:
             if key == 'csys-obj-type' and value:
                 created_types.add(value.lower())
 
-        new_labels_set = set(new_labels)
-        for label in deployment.labels:
-            if (label.key, label.value) not in new_labels_set:
+        labels_to_delete = self.get_labels_to_delete(deployment, labels)
+        for label in labels_to_delete:
+            if label.key == 'csys-obj-type':
                 delete_types.add(label.value.lower())
         return created_types, delete_types
 
@@ -2326,11 +2326,9 @@ class ResourceManager(object):
         If an existing label is not in the new labels list, it will be deleted.
         """
         labels_to_create = self.get_labels_to_create(resource, new_labels)
-
-        new_labels_set = set(new_labels)
-        for label in resource.labels:
-            if (label.key, label.value) not in new_labels_set:
-                self.sm.delete(label)
+        labels_to_delete = self.get_labels_to_delete(resource, new_labels)
+        for label in labels_to_delete:
+            self.sm.delete(label)
 
         self.create_resource_labels(labels_resource_model,
                                     resource,
@@ -2344,6 +2342,15 @@ class ResourceManager(object):
             (label.key, label.value) for label in existing_labels)
 
         return new_labels_set - existing_labels_tup
+
+    @staticmethod
+    def get_labels_to_delete(resource, new_labels):
+        labels_to_delete = set()
+        new_labels_set = set(new_labels)
+        for label in resource.labels:
+            if (label.key, label.value) not in new_labels_set:
+                labels_to_delete.add(label)
+        return labels_to_delete
 
     def create_resource_labels(self,
                                labels_resource_model,
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -1398,6 +1398,14 @@ class ResourceManager(object):
                 parents.append(value)
         return parents
 
+    @staticmethod
+    def get_deployment_object_type_from_labels(labels):
+        types = set()
+        for label, value in labels:
+            if label == 'csys-obj-type' and value:
+                types.add(value.lower())
+        return types
+
     def verify_deployment_parent_labels(self, parents, deployment_id):
         if not parents:
             return
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -1398,13 +1398,19 @@ class ResourceManager(object):
                 parents.append(value)
         return parents
 
-    @staticmethod
-    def get_deployment_object_type_from_labels(labels):
-        types = set()
-        for label, value in labels:
-            if label == 'csys-obj-type' and value:
-                types.add(value.lower())
-        return types
+    def get_deployment_object_types_from_labels(self, deployment, labels):
+        new_labels = self.get_labels_to_create(deployment, labels)
+        created_types = set()
+        delete_types = set()
+        for key, value in new_labels:
+            if key == 'csys-obj-type' and value:
+                created_types.add(value.lower())
+
+        new_labels_set = set(new_labels)
+        for label in deployment.labels:
+            if (label.key, label.value) not in new_labels_set:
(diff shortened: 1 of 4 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -1398,6 +1398,14 @@ def get_deployment_parents_from_labels(labels):
                 parents.append(value)
         return parents
 
+    @staticmethod
+    def get_deployment_object_type_from_labels(labels):
+        types = set()
+        for label, value in labels:
```
YoniItzhak, line 1404: Seme comment as in last PR, it's better to call it `key` instead of `label`, as ` `label` defines the whole object :)

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -103,10 +103,23 @@ def _populate_direct_deployment_counts(deployment):
...
     @staticmethod
-    def _handle_deployment_labels(rm, deployment, raw_labels_list):
-        deployment_parents = deployment.deployment_parents
-        new_labels = rest_utils.get_labels_list(raw_labels_list)
+    def _update_deployment_counts(rm, deployment, new_labels):
+        if deployment.deployment_parents:
+            types = rm.get_deployment_object_type_from_labels(new_labels)
+            to_srv = deployment.is_environment and 'environment' not in types
+            to_env = not deployment.is_environment and 'environment' in types
+            _type = 'service' if to_srv else 'environment' if to_env else None
```
YoniItzhak, line 111: How would this be `None`?

rest-service/manager_rest/resource_manager.py
```diff
@@ -1398,6 +1398,14 @@ def get_deployment_parents_from_labels(labels):
...
+            if label == 'csys-obj-type' and value:
+                types.add(value.lower())
```
YoniItzhak, line 1406: Why happens if the user provides both `environment` and `service` as different values?

rest-service/manager_rest/rest/rest_utils.py
```diff
@@ -719,6 +719,39 @@ def decrease_deployment_counts_in_graph(self, target_id, source):
...
+        Update counter for deployment that is already connected to other
+        deployments where deployment could be changed from `service` to
+        `environment` and vice versa and we need to take care of deployment
+        counters so that count will not mess up
+        :param source: Deployment source instance that its type get changed
+        :param convert_type: The type we need to convert source deployment
+        to. It could be changed from `service` to `environment` and vice versa
+        """
+        if convert_type not in ('environment', 'service',):
+            return
```
YoniItzhak, line 735: Throw an error?

rest-service/manager_rest/resource_manager.py
mabuaisha: yeah, my bad will update that

rest-service/manager_rest/rest/resources_v3_1/deployments.py
mabuaisha: it would be None like if did not change the the `csys-type-obj` then it should be None, that means we did not remove/add `csys-type-obj` to the deployment

rest-service/manager_rest/resource_manager.py
mabuaisha: Well, the most important  value for us is `environment` that way we can tell if this is environment or not e otherwise it will be service

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: well, I do not want to throw error here because user could add values other than these two which requires no need to update the counter

rest-service/manager_rest/rest/resources_v3_1/deployments.py
YoniItzhak: But, unless I'm missing something, it would always get a boolean value.

mabuaisha: _type will be one of three values: `service` or `environment` or `None` and `None` means nothing is need to do where no add/remove happened to the `csys-type-obj` label

rest-service/manager_rest/resource_manager.py
YoniItzhak: Don't we want to throw an error in such a case? 
Let's say the user updates the deployments labels to be 
```
{csys-obj-type: environment}, 
{csys-obj-type: service}
```
What should we do? I think that because it is ambiguous, we should throw an error.

rest-service/manager_rest/rest/rest_utils.py
YoniItzhak: Unless I missed something, this would never get here with `None`.

rest-service/manager_rest/resource_manager.py
mabuaisha: If you remember the jira we raised regarding the restrictions for label values. Ofer does not want to add such limitation to raise error to users and on this case if the deployment originally was service and user added two of these the it will be converted to environment `not deployment.is_environment and 'environment' in types`  and if it was originally `environment` and user added two values then it will not be changed since its already `environment`

mabuaisha: The point is that we do not want to add more complexity to the values related to the labels values

YoniItzhak: I see. I don't think this the right behavior, but if this is the wanted behavior it's a different thing.

rest-service/manager_rest/rest/rest_utils.py
mabuaisha: Yeah, it will not and it could be removed

rest-service/manager_rest/resource_manager.py
mabuaisha: I agree, but for now we are going to stick with this as we only worried about added/removed labels so that we can update the counters

rest-service/manager_rest/resource_manager.py
```diff
@@ -1398,6 +1398,20 @@ def get_deployment_parents_from_labels(labels):
...
+        created_types = set()
+        delete_types = set()
+        for key, value in new_labels:
+            if key == 'csys-obj-type' and value:
+                created_types.add(value.lower())
+
+        new_labels_set = set(labels)
+        for label in deployment.labels:
+            if (label.key, label.value) not in new_labels_set:
+                delete_types.add(label.value.lower())
```
YoniItzhak, line 1412: This takes any label, and not only the ones with the key `csys-obj-type`. So basically also the label `a: environment`  will be counted.

YoniItzhak, line 1412: Plus, this is the exact same code from https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/resource_manager.py#L2316. 
So why not putting it in a different function, let's say `labels_to_delete`, and then run on the labels there and check if their key equals `csys-obj-type`?

mabuaisha: you are right, updated that and add check before add any deleted types

mabuaisha: This way cannot make it generic for other places need to call labels_to_delete
```
        new_labels_set = set(labels)
        for label in deployment.labels:
            if label.key == 'csys-obj-type':
                if (label.key, label.value) not in new_labels_set:
                    delete_types.add(label.value.lower())
 ```

mabuaisha: This get updated now, ignore my last comment

YoniItzhak (approved):

PR #2866, opened 2021-04-04
RD-1868 System filters infrastructure

YoniItzhak:
The system filters infrastructure follows these guidelines: 

- System filters will be defined in the DB using a boolean "flag" named `is_system_filter`. 
- System filters are `csys-` prefixed, and therefore a user won't be able to create a `csys-` prefixed filters.
- A user shouldn't be able to modify or delete a system filter.

I also changed in this PR the names of the constants `CFY_PREFIX` and `CFY_LABELS` to more accurate ones - `RESERVED_PREFIX` and `RESERVED_LABELS`.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/filters.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/filters.py
@@ -103,7 +103,7 @@ class FiltersId(SecuredResource):
         rest_utils.validate_inputs({'filter_id': filter_id})
         storage_manager = get_storage_manager()
         filter_elem = storage_manager.get(filters_model, filter_id)
-        _verify_not_a_system_filter(filter_elem)
+        _verify_not_a_system_filter(filter_elem, 'delete')
         storage_manager.delete(filter_elem, validate_global=True)
         return None, 204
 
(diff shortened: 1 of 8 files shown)


rest-service/manager_rest/rest/resources_v3_1/filters.py
```diff
@@ -158,6 +165,12 @@ def _get_filter_rules_by_type(filter_rules_list, filter_rule_type):
             filter_rules_list if filter_rule['type'] == filter_rule_type]
 
 
+def _verify_not_a_system_filter(filter_elem):
+    if filter_elem.is_system_filter:
+        raise manager_exceptions.IllegalActionError(
+            'Cannot delete a system filter')
```
glukhman, line 171: probably "update" instead of "delete" when called from patch?

YoniItzhak: Oh right, thanks!

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -797,6 +797,10 @@ class _Filter(CreatedAtMixin, SQLResourceBase):
```
glukhman, line 797: Just noticed: the class is called _Filter, rather than BaseFilter. I don't know how standard it is...

glukhman (approved):

rest-service/manager_rest/storage/resource_models.py
YoniItzhak: Yep, it might be worth changing it.

PR #2870, opened 2021-04-06
RD-1535 Resume an execution group

tehasdf:
Rewrite .reset_operations a bit to make bulk actions easier,
and implement a resume for a whole execution group.

It will only resume the executions that are indeed cancelled/failed.
If there are any executions still running, they will be unaffected.


tehasdf:
the other parts are cloudify-cosmo/cloudify-common#742 and cloudify-cosmo/cloudify-cli#1271

geokala (approved):

PR #2873, opened 2021-04-06
several integration test fixes

tehasdf:
Multiple, mostly unrelated fixes for integration tests.
Unfortunate that they're all in a single PR, but imagine waiting
for CI 18 times.

Most of them has to do with either:
1. using external ssl
2. the new queueing mechanism
3. async deployment create

### DIFF
--- a/tests/integration_tests/tests/agentless_tests/test_events.py
+++ b/tests/integration_tests/tests/agentless_tests/test_events.py
@@ -143,7 +143,7 @@ class EventsTest(AgentlessTestCase):
         else:
             self.fail("Expected logs to be found")
 
-    @pytest.mark.skipif(reason='causes the DB deadlock in snapshot')
+    @pytest.mark.skip(reason='causes the DB deadlock in snapshot')
     def test_snapshots_events(self):
         """ Make sure snapshots events appear when using the
          'cfy events list' command """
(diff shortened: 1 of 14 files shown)


mateumann (approved):

tests/integration_tests/tests/agentless_tests/test_events.py
```diff
@@ -143,6 +143,7 @@ def test_events_with_logs(self):
         else:
             self.fail("Expected logs to be found")
 
+    @pytest.mark.skipif(reason='causes the DB deadlock in snapshot')
```
geokala, line 146: You seem to have set this as a skipif without setting a condition for it to be skipped?

tests/integration_tests/tests/agentless_tests/test_snapshot.py
```diff
@@ -446,41 +446,38 @@ def _upload_and_restore_snapshot(
...
+            self._upload_and_validate_snapshot(snapshot_path,
+                                               snapshot_id)
+            self.logger.debug('Restoring snapshot...')
+            self.client.maintenance_mode.activate()
+            execution = self.client.snapshots.restore(
+                snapshot_id,
+                ignore_plugin_failure=ignore_plugin_failure)
+            # give the database some time to downgrade/upgrade before running
+            # requests to avoid the deadlock described in CY-1455
+            time.sleep(10)
```
geokala, line 459: Is there some sort of actual polling check we can reasonably do for this instead of an arbitrary sleep?

tehasdf: we'd have to look at the db directly i think; maybe so, but for now i want to just fix the tests, not rewrite them

geokala (approved):
Green, as the only remaining comment doesn't pertain to new code.

PR #2882, opened 2021-04-08
node.plugins is a list, not a dict

tehasdf:
default to the empty list, not the empty dict

Also rename the variable in dep-up. It's obvious that it's not a dict
there, because we immediately do .append and .remove on it.

Also change the getattr default in dep-up: I think it's actually
never used for .plugins, but might be for plugins_to_install

This fixes the inte-tests in test_deployment_update_misc.py


mateumann (approved):

tehasdf:
plugins is first set here, to a list: https://github.com/cloudify-cosmo/cloudify-common/blob/ad0fbfabd3bba5844bee16dd2301d2eda48a58d7/dsl_parser/elements/node_templates.py#L580

PR #2890, opened 2021-04-11
Rd 1987 stabilize integ tests

glukhman:


rest-service/manager_rest/rest/resources_v3_1/nodes.py
```diff
@@ -42,11 +42,11 @@ def _node_from_raw_node(self, raw_node):
...
-            number_of_instances=scalable.get('current_instances') or 1,
-            planned_number_of_instances=scalable.get('current_instances') or 1,
-            deploy_number_of_instances=scalable.get('default_instances') or 1,
-            min_number_of_instances=scalable.get('min_instances') or 1,
-            max_number_of_instances=scalable.get('max_instances') or 1,
+            number_of_instances=scalable.get('current_instances', 1),
+            planned_number_of_instances=scalable.get('current_instances', 1),
+            deploy_number_of_instances=scalable.get('default_instances', 1),
+            min_number_of_instances=scalable.get('min_instances', 1),
+            max_number_of_instances=scalable.get('max_instances', 1),
```
mabuaisha, line 49: what if the `scalable` contains `current_instances` with None value would not be that an issue ? it will never then get the default value 1.

One more thing, watch out to the spaces after `=`

mabuaisha (dismissed):

mabuaisha (approved):

PR #2905, opened 2021-04-14
RD-2046 Drop a query to avoid deadlocks

tehasdf:
For the dequeueing, and "same-deployment-first", we can avoid having
to do two (locking) queries, because we can just get the sorting done
in a single query.

Additionally, make sure we queue, if there are already queued executions
(FIFO semantics for executions).

See commit messages too.

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -998,7 +998,7 @@ class ResourceManager(object):
     def _check_for_active_executions(self, execution, queue):
         running = self.list_executions(
             filters={
-                'deployment_id': execution.deployment_id,
+                '_deployment_fk': execution._deployment_fk,
                 'id': lambda col: col != execution.id,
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -266,29 +266,20 @@ def _get_queued_executions(self, finished_execution):
...
-            other_queued = self.sm.list(
+            queued_executions = self.sm.list(
                 models.Execution,
                 filters={
                     'status': ExecutionState.QUEUED_STATE,
-                    'deployment_id': lambda col: col != deployment_id,
+                    'is_system_workflow': False
                 },
-                sort=sort_by,
+                sort=OrderedDict([
```
tehasdf: needs to be ordered, so that executions are sorted by deployment first, and only within those groups by the timestamp

geokala (approved):

PR #2907, opened 2021-04-15
RD-1969 Get rid of events.id and logs.id columns

mateumann:
Let's get rid of `events.id` and `logs.id` which were always empty anyway.

### DIFF
--- a/rest-service/manager_rest/storage/resource_models.py
+++ b/rest-service/manager_rest/storage/resource_models.py
@@ -1265,7 +1265,9 @@ class Event(SQLResourceBase):
             name='events__one_fk_not_null'
         ),
     )
-    id = db.Column(db.Text, index=True, default=lambda: str(uuid.uuid4()))
+    is_id_unique = False
+
+    id = None
     timestamp = db.Column(
         UTCDateTime,
         default=datetime.utcnow,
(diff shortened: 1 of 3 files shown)


tehasdf:
we note that inserting logs/events never even populated the id field anyway, https://github.com/cloudify-cosmo/cloudify-manager/blob/2313bd844ef74c6bddb177d6d29ab131de47d973/amqp-postgres/amqp_postgres/postgres_publisher.py#L21

tehasdf (dismissed):

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -1265,7 +1265,9 @@ class Event(SQLResourceBase):
             name='events__one_fk_not_null'
         ),
     )
-    id = db.Column(db.Text, index=True, default=lambda: str(uuid.uuid4()))
+    is_id_unique = False
+
+    id = None
```
tehasdf, line 1270: this is just to override the parent class attribute

geokala: That should maybe be a comment in the code itself?

mateumann: I'll add that comment in next PR

mateumann: (added in this one)

tehasdf (approved):

PR #2913, opened 2021-04-19
RD-1816 Deployment update workflow

tehasdf:
For making the dep-update async so that it can be called for a
deployment group: I give up. It's not possible to refactor it enough
in this timeframe.

Instead, I'm going to cheat!

Add a new builtin workflow that just creates a deployment update.
THAT workflow can well be run for a group (as a bonus, it can be also
used for scheduling and whatever else).

### DIFF
--- a/rest-service/manager_rest/storage/resource_models.py
+++ b/rest-service/manager_rest/storage/resource_models.py
@@ -976,16 +976,17 @@ class Execution(CreatedAtMixin, SQLResourceBase):
     def get_workflow(self, deployment=None, workflow_id=None):
         deployment = deployment or self.deployment
         workflow_id = workflow_id or self.workflow_id
+
+        if deployment and deployment.workflows\
+                and workflow_id in deployment.workflows:
+            return deployment.workflows[workflow_id]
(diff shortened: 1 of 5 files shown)


rest-service/manager_rest/storage/resource_models.py
```diff
@@ -967,22 +967,26 @@ def _system_workflow_task_name(self, wf_id):
             'delete_deployment_environment':
                 'cloudify_system_workflows.deployment_environment.delete',
             'update_plugin': 'cloudify_system_workflows.plugins.update',
-            'upload_blueprint': 'cloudify_system_workflows.blueprint.upload'
+            'upload_blueprint': 'cloudify_system_workflows.blueprint.upload',
+            'csys_update_deployment':
+            'cloudify_system_workflows.deployment_environment.'
```
geokala, line 972: This would be nicer with a bit of indentation (similar to the delete_deployment_environment one.

geokala (approved):
Green with one style related comment (which I consider optional, though it would be nice).

geokala (approved):

PR #2914, opened 2021-04-19
RD-2019 Update blueprint state as part of the migration

glukhman:


tehasdf (approved):
we want the state to be populated for already-existing blueprints, but we don't want to actually have a default on that column

set the state after adding the column, then

PR #2917, opened 2021-04-21
Update manager version in DB during upgrade

glukhman:


rest-service/manager_rest/update_managers_version.py
```diff
@@ -0,0 +1,31 @@
...
+import yaml
+
+from manager_rest.storage import models, get_storage_manager
+
+
+def update_managers_version(version):
+    logger.debug('Updating Cloudify managers version in DB...')
+    sm = get_storage_manager()
+    managers = sm.full_access_list(models.Manager)
+    for manager in managers:
```
mateumann, line 14: I'm not sure we should update ALL the managers.  I'd rather we do it one by one.  Please consider using something like `WHERE hostname='this_manager_hostname'`

tehasdf (approved):

PR #2918, opened 2021-04-21
RD-1417 In case of failed execution log only error

mateumann:
... instead of whole traceback (for execution_groups)


mateumann:
Actually let's log all the tracebacks' details because those might come handy some day.  Instead displaying only "readable part" will be handled by the cloudify-cli.

PR #2924, opened 2021-04-22
RD-2052 case sensitive label values

YoniItzhak:
We allow case-sensitive label values.


YoniItzhak:
This PE is waiting for an answer from Ofer regarding whether or not we should validate the `csys-obj-type` values.

PR #2928, opened 2021-04-22
Update manager version in db

mateumann:
* Update manager version in DB during upgrade

* Add rest config envver

* Update cloudify-rest-service.spec


mateumann:
Let's do it "properly", but first let's think about "how"

PR #2946, opened 2021-04-26
RD-2102 Add REST API to retrieve workflows

mateumann:


rest-service/manager_rest/rest/resources_v3_1/workflows.py
```diff
@@ -0,0 +1,66 @@
...
+
+        return ListResult([w.as_dict() for w in workflows],
+                          {'pagination': {
+                              'total': len(workflows),
+                              'size': len(workflows),
+                              'offset': 0,
+                          }})
+
+
+def _merge_workflows(w1: list('Workflow'),
```
tehasdf, line 61: the right syntax is 
```suggestion
def _merge_workflows(w1: 'List[Workflow]',
```
(with a `from typing import List`)

try toying with `mypy foo.py`, here's an example I tried with:

```python
from typing import List

class Workflow:
    pass

def f(a: 'List[Workflow]'):
    pass

f([Workflow()])
```

rest-service/manager_rest/rest/resources_v3_1/workflows.py
```diff
@@ -0,0 +1,78 @@
...
+    workflows = []
+    for dep in deployments:
+        workflows = _merge_workflows(
+            workflows,
+            models.Deployment._list_workflows(dep.workflows)
+        )
+    return workflows
+
+
+def _merge_workflows(w1: 'List[Workflow]',
```
tehasdf, line 73: actually there's a neat trick we can do, to avoid having to have this function:
implement hash+eq for Workflow:
```python
class Workflow(object):
  def __hash__(self):
     return hash(self.name)
  def __eq__(self, other):
     return self.name == other.name  # could also consider params but currently it just merges by name
```

then, `_extract_workflows` could be as simple as this:
```python
workflows = set()
for dep in deployments:
  workflows |= set(models.Deployment._list_workflows(dep.workflows)
```

tehasdf: i think that's better, because augmenting types feels more powerful than adding util functions

rest-service/manager_rest/test/endpoints/test_workflows.py
```diff
@@ -0,0 +1,98 @@
...
+            deployment_id='d3',
+            blueprint_id='b3',
+            blueprint_file_name='blueprint_with_workflows_with_parameters_'
+                                'types.yaml',
+            labels=[{'zxc': '2'}, {'asd': '2'}]
+        )
+
+    def test_workflows_list_default_workflows(self):
+        workflows = self.client.workflows.list(id='d1')
+        assert set(w.name for w in workflows.items) == {
```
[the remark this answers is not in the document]
tehasdf: i'm not very sure about this one, just plz consider if it's worth it

tehasdf (approved):

PR #2947, opened 2021-04-26
Lock the deployment when updating its attributes

tehasdf:
The deployment update as part of create-dep-env must be locking,
so that concurrent changes are not lost.

See commit messages

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/deployments.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/deployments.py
@@ -42,6 +42,7 @@ from manager_rest.manager_exceptions import (
 from manager_rest.resource_manager import get_resource_manager
 from manager_rest.maintenance import is_bypass_maintenance_mode
 from manager_rest.dsl_functions import evaluate_deployment_capabilities
+from manager_rest.execution_token import current_execution
 
 from manager_rest.rest.filters_utils import get_filter_rules_from_filter_id
 from manager_rest.rest import (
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -250,28 +273,29 @@ def patch(self, deployment_id):
...
+            for attrib in allowed_attribs:
+                if attrib not in request_dict:
+                    continue
+                previous = getattr(deployment, attrib, None)
+                if previous is not None:
+                    raise ConflictError(f'{attrib} is already set')
+                setattr(deployment, attrib, request_dict[attrib])
+            if 'labels' in request_dict:
+                raw_labels_list = request_dict.get('labels', [])
+                self._handle_deployment_labels(
```
mabuaisha, line 292: @tehasdf I think here we could have have database deadlock since inside `_handle_deployment_labels` we will use locking for Deployment regarding counts + statuses.

mabuaisha: especially in case we have multiple deployments running at the same time

tehasdf: i don't see the deadlock, this only locks the deployment

mabuaisha: do not forget we have another lock=True inside the statuses + counter propagation

mabuaisha (approved):

PR #2952, opened 2021-04-27
RD-1958 Add get_environment_capability function

mabuaisha:
Update the FunctionEvaluationStorage to add support to use `get_environment_capability` function.

[cloudify-common PR](https://github.com/cloudify-cosmo/cloudify-common/pull/755)


tehasdf (commented):
I suggest allowing get_environment_capability to also be able to traverse nested objects the same way get_capability can (eg. `{get_capability: [cap, a, b, 1]}` does `dep.capabilities['cap']['a']['b'][1]`)

mabuaisha:
@tehasdf sure will handle that

tehasdf (approved):

PR #2953, opened 2021-04-27
Migrations: handle labels + group events

tehasdf:
Two separate fixups, see each separately.

This allows migrations to actually run in cases:
- upgrade 5.2→5.3, if the system already has labels (eg. a 5.2 snapshot with deployment labels)
- downgrade 6.0→5.3 if there exist execution group logs/events

This allows downgrading like `/opt/manager/env/bin/alembic downgrade 9d261e90b1f3` (that id = 5.2) to run, and afterwards, `alembic upgrade head`

### DIFF
--- a/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
+++ b/resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
@@ -385,9 +385,29 @@ def _create_blueprints_labels_table():
                     unique=False)
 
 
+dl_table = sa.table(
+    'deployments_labels',
+    sa.Column('_labeled_model_fk'),
+    sa.Column('_deployment_fk')
+)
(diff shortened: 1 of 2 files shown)


resources/rest-service/cloudify/migrations/versions/396303c07e35_5_2_to_5_3.py
```diff
@@ -385,9 +386,29 @@ def _create_blueprints_labels_table():
                     unique=False)
 
 
+dl_table = sa.table(
```
tehasdf, line 389: can't use `models.DeploymentLabel.__table__` because that table never has both of these columns (it had one of them in 5.2, and the other in 5.3).

Instead, declare a synthetic "table" that pretends it has both

mateumann (approved):

geokala (approved):

PR #2976, opened 2021-05-05
RD-2249 deployment id template: more variables

tehasdf:
Additional template variables for the deployment ID template.

Note that `count` requires actually keeping a counter. So I add
a column.
The other two are trivial.


tehasdf:
this is about how currently when adding deployments to a group, you can do
`.add_deployments('g1', [{'id': 'xxx-{uuid}'}])` with the rest-client
or `deployment_settings: id_template: 'xxx-{uuid}'` in the blueprint.

However, `{uuid}` and `{group_id}` are currently the only allowed
variables in that template.

And this PR adds 3 more.

geokala (approved):

PR #3001, opened 2021-05-10
CY-2316 Prevent circular dependenciec in deployment update

mateumann:
In case the deployment update ends up having circular dependencies
(a.k.a. cycles within a nodes' relationship graph), throw an exception.


geokala:

PR #3005, opened 2021-05-11
RD-1419 Save exec-group results into dep-group

tehasdf:
Add a way to save exec-group results.

Deployments for which the execution succeeded, are added to the
success target group.
Deployments for which the execution failed, are added to the failure
target group.

Also see the docstring for the rest-client in cloudify-cosmo/cloudify-common#768

We store this in 2 ways: when setting the target groups, to catch
executions that were already finished by then; and also in
update-execution-status, so that other executions are processed
as they finish.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/executions.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/executions.py
@@ -300,7 +300,7 @@ class ExecutionGroupsId(SecuredResource):
 
     @authorize('execution_group_update')
     @rest_decorators.marshal_with(models.ExecutionGroup)
-    def put(self, group_id, **kwargs):
+    def patch(self, group_id, **kwargs):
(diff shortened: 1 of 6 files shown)


rest-service/manager_rest/rest/resources_v3_1/executions.py
```diff
@@ -296,3 +297,56 @@ def _resume_group(self, sm, group, action):
         amqp_client.add_handler(handler)
         with amqp_client:
             group.start_executions(sm, rm, handler)
+
+    @authorize('execution_group_update')
+    @rest_decorators.marshal_with(models.ExecutionGroup)
+    def put(self, group_id, **kwargs):
```
mateumann, line 303: Could you please make it PATCH?

mateumann: (also update in c-common)

PR #3012, opened 2021-05-11
Fix upgrade command

YoniItzhak:


tehasdf (approved):
moving the `upgrade_managers_version` script to AFTER the migration, because well, it is using the db, so the new code must use the updated db, with the migrated schema.
Before, it was new code + old schema in the db = exception

PR #3030, opened 2021-05-13
snapshot-restore: also pause the execution-scheduler

tehasdf:
Similar to pausing amqp-postgres, those db-using services must not
run while the db is being down/upgraded. The scheduler even more so,
because if it just happens to try and run something during that time,
oh boy is that going to break.


tehasdf:
the 2nd commit is unrelated but I don't want to wait on CI on that one separately. It just makes sure that the tests' events printer reconnects on any connection break.

geokala (approved):

PR #3044, opened 2021-05-16
RD-2248 Creating system filters during upgrade

YoniItzhak:
We would like to create default system filters that will allow the UI to filter environments and services easily. 
The first try was to just create them during the manager installation, but the created filters were gone after snapshot restore.  This PR fixes it.
The first idea for the fix was to create the system filters during the alembic upgrade. This doesn't work as `default_tenant` tenant and `admin` user are created only after the upgrade and are necessary for the creation of the filters. 
The second fix, that does work, is using a script for creating the system filters during installation, `cfy_manager upgrade`, and snapshot restore. 

This PR needs to be merged before its associated one, https://github.com/cloudify-cosmo/cloudify-manager-install/pull/1160, in order for its CI to pass.

### DIFF
--- a/packaging/rest-service/files/opt/manager/scripts/create_system_filters.py
+++ b/packaging/rest-service/files/opt/manager/scripts/create_system_filters.py
@@ -1,0 +1,72 @@
+#!/opt/manager/env/bin/python
+
+import os
+from datetime import datetime
+
+from manager_rest import constants
+from manager_rest.storage import models
+from manager_rest.storage.models_base import db
+from manager_rest.flask_utils import setup_flask_app
+
+
+def create_system_filters():
+    with setup_flask_app().app_context():
+        current_deployment_filters = db.session.query(models.DeploymentsFilter)
+        curr_dep_filters_ids = {dep_filter.id for dep_filter
+                                in current_deployment_filters}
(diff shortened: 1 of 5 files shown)


packaging/rest-service/files/opt/manager/scripts/create_system_filters.py
```diff
@@ -0,0 +1,72 @@
...
+from manager_rest.storage.models_base import db
+from manager_rest.flask_utils import setup_flask_app
+
+
+def create_system_filters():
+    with setup_flask_app().app_context():
+        current_deployment_filters = db.session.query(models.DeploymentsFilter)
+        curr_dep_filters_ids = {dep_filter.id for dep_filter
+                                in current_deployment_filters}
+        if ({'csys-environment-filter', 'csys-service-filter'} ==
```
geokala, line 17: This will lead to an interesting error if one of those exists but the other doesn't. Is that acceptable?

YoniItzhak: Changed the script to be idempotent. I.e. each filter will be created only if it does not exist.

geokala (approved):

PR #3053, opened 2021-05-18
RD-523 Optimize the operation-update call

tehasdf:
This makes the endpoint slightly faster, saving on the order of 3ms
per call (and one roundtrip, if the db is remote).

Instead of fetching-and-updating the execution operation counts,
use a sql UPDATE query. We never retrieve the former count, so
no need for locking.

Best see each commit separately

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/operations.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/operations.py
@@ -28,7 +28,8 @@ from manager_rest.rest.rest_decorators import (
 )
 from manager_rest.storage import (
     get_storage_manager,
-    models
+    models,
+    db
 )
 from manager_rest.security.authorization import authorize
 from manager_rest.resource_manager import get_resource_manager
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/resources_v3_1/operations.py
```diff
@@ -94,13 +95,32 @@ def patch(self, operation_id, **kwargs):
...
+        calls, for performance: the operation state-update call is on
+        the critical path for all operations in a workflow; this saves
+        about 3ms over the ORM approach (which requires fetching the
+        execution; more if the DB is not local).
+        """
+        exc_table = models.Execution.__table__
+        tg_table = models.TasksGraph.__table__
+        db.session.execute(
+            exc_table.update()
+            .where(db.and_(
```
tehasdf, line 123: note that the original had `execution.finished_operations is not None:` but we're not adding a `is null` check here: this is because null+1=null in sql, so they're effectively ignored anyway

(and, in practice, there aren't any executions who have operations, while not having the counts set to 0)

geokala (approved):

PR #3071, opened 2021-05-21
RD-2349 status reporter debugging hints

mateumann:


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -123,20 +123,32 @@ def get_cluster_status(detailed=False):
...
 
+    # find unexpected metrics
+    unexpected_metrics = [
+        result for result in global_results
+        if not _host_matches(result.get('metric'), cluster_nodes.keys())
+    ]
+    if unexpected_metrics:
+        current_app.logger.warning(
+            'These metrics do not match monitored IP address%s (%s): %s',
+            '' if len(cluster_nodes.keys()) == 1 else 'es',
```
tehasdf, line 143:
```suggestion
            '' if len(cluster_nodes) == 1 else 'es',
```

tehasdf (dismissed):

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -123,20 +123,32 @@ def get_cluster_status(detailed=False):
     return cluster_status
 
 
-def _add_monitoring_data(cluster_nodes):
-    # Get all the data in one fell swoop (thanks to the underlying Prometheus
-    # federation).
+def _add_monitoring_data(cluster_nodes: dict):
```
geokala, line 126: If you're adding the parameter type, it might be good to add the return type (which I think in this case is -> None?)

tehasdf (approved):

PR #3077, opened 2021-05-24
RD-1580 Refetch config on changes

tehasdf:
Add a before-request that reloads the config when it changes, based
on the updated-at fields


rest-service/manager_rest/server.py
```diff
@@ -88,6 +88,31 @@ def cope_with_db_failover():
...
+            db.session.query(models.Role.updated_at.label('updated_at'))
+        ).subquery()
+    )
+    config_last_updated = db.session.query(
+        db.func.max(last_updated_subquery.c.updated_at)
+    ).scalar()
+    current_app.logger.debug('Last updated locally: %s, in db: %s',
+                             config.instance.last_updated,
+                            config_last_updated)
+    if config_last_updated is not None and (
```
geokala, line 111: Not a big issue, but this might be clearer if config_last_updated was called db_config_last_updated- on first glance it looked like this line and the next line were referring to the same variable.

mateumann (approved):

geokala (approved):

PR #3083, opened 2021-05-25
Set execution status to FAILED if global/dependencies checks fail

glukhman:
This ports #3082 to 6.0.0


tehasdf (approved):

tehasdf:
I've reset this to contain the same code as #3082, together with the changes from #3084

PR #3090, opened 2021-05-26
RD-2499 Cluster status manager refactor

mateumann:
This is an attempt to have a little bit more readable `cluster_status_manager.py`.  Mostly cosmetic changes, except for the second commit, which modifies how we iterate over Prometheus' results.

### DIFF
--- a/rest-service/manager_rest/cluster_status_manager.py
+++ b/rest-service/manager_rest/cluster_status_manager.py
@@ -102,7 +102,7 @@ QUERY_STRINGS = {
 }
 
 
-def get_cluster_status(detailed=False):
+def get_cluster_status(detailed=False) -> typing.Dict[str, typing.Any]:
     cluster_nodes, cloudify_version = _get_cluster_details()
     _add_monitoring_data(cluster_nodes)
 
@@ -124,7 +124,7 @@ def get_cluster_status(detailed=False):
     return cluster_status
 
 
-def _add_monitoring_data(cluster_nodes: dict) -> None:
+def _add_monitoring_data(cluster_nodes: typing.Dict[str, dict]) -> None:
     """Add metrics data and information on services for the cluster nodes."""
     query_string = ' or '.join(QUERY_STRINGS.values())
     global_results = prometheus_query(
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -102,7 +102,7 @@
 }
 
 
-def get_cluster_status(detailed=False):
+def get_cluster_status(detailed=False) -> typing.Dict[str, typing.Any]:
```
tehasdf, line 105:
```
manager_rest/cluster_status_manager.py:122: error: Incompatible types in assignment (expression has type "str", target has type "Dict[Any, Dict[str, Any]]")
```
this isn't really on this line, and this type annotation is fiiiine, but it's about the `cluster_status['status'] = _get_overall_state(cluster_status)` line - mypy really doesn't like heterogenous dicts (and it's right, they're ugly)

I suggest doing this:
```python
def get_cluster_status(detailed=False) -> typing.Dict[str, typing.Any]:
    cluster_nodes, cloudify_version = _get_cluster_details()
    _add_monitoring_data(cluster_nodes)

    cluster_services = {
        CloudifyNodeType.BROKER:
        _get_broker_state(cluster_nodes, cloudify_version, detailed),
        CloudifyNodeType.DB:
        _get_db_state(cluster_nodes, cloudify_version, detailed),
        CloudifyNodeType.MANAGER:
        _get_manager_state(cluster_nodes, cloudify_version, detailed),
    }
    return {
        'services': cluster_services,
        'status': _get_overall_state(cluster_services)
    }
```
that way, there's no separate local variable for `cluster_status`, so just the typehint from the return value suggests that it's Any
also of course change `_get_overall_state` to take the services dict.

And I think it's a good change anyway, even though it's mostly just babysitting mypy

rest-service/manager_rest/cluster_status_manager.py
```diff
@@ -212,8 +369,11 @@ def _get_manager_state(cluster_nodes, cloudify_version, detailed):
     )
 
 
-def _get_cluster_service_state(cluster_nodes, cloudify_version, detailed,
-                               service_type):
+def _get_cluster_service_state(cluster_nodes: typing.Dict[str, dict],
+                               cloudify_version: str,
+                               detailed: bool,
+                               service_type: str) -> typing.Dict[str,
```
tehasdf, line 375: mypy also hates this heterogenous dict, I'll just send in my suggestion as a pr

PR #3104, opened 2021-05-31
WDR-35 Fix sorting by environment_type with filter (#3098)

YoniItzhak:
Description here https://github.com/cloudify-cosmo/cloudify-manager/pull/3098


YoniItzhak:
We don't push code to v6.0.0 anymore.

PR #3117, opened 2021-06-16
RD-518 Create events/logs via REST

tehasdf:
This just adds a POST endpoint. Removing amqp-postgres, etc, is for the future.


tehasdf:
needs cloudify-cosmo/cloudify-common#802 and cloudify-cosmo/cloudify-manager-install#1187

geokala (approved):

PR #3119, opened 2021-06-18
RD-515 Insert events when updating operation state

tehasdf:


mateumann (dismissed):

rest-service/manager_rest/rest/resources_v3_1/operations.py
```diff
@@ -84,21 +87,76 @@ def put(self, operation_id, **kwargs):
...
+            message = common_events.format_event_message(
+                operation.name,
+                operation.state,
+                result,
+                exception,
+                current_retries,
+                total_retries,
+            )
+            event_type = common_events.get_event_type(operation.state)
+        except RuntimeError:
```
geokala, line 140: What RuntimeErrors are we expecting that we don't want to log?

tehasdf: missing (ignored) event type, thrown in https://github.com/cloudify-cosmo/cloudify-common/blob/78c8d7c2e6283135ce51b8d98ab6de95cf1bffae/cloudify/workflows/events.py#L70

geokala: Hmm. Maybe we should've made that a more specific exception, but not a massive impact here at the moment.

geokala (approved):

PR #3120, opened 2021-06-18
snap-res refresh_roles: store timestamps in utc

tehasdf:
this sets to NOW(), but it needs to set in utc, because we use utc
everywhere, when using that column

### DIFF
--- a/tests/integration_tests/tests/agent_tests/test_events.py
+++ b/tests/integration_tests/tests/agent_tests/test_events.py
@@ -20,25 +20,17 @@ import pytest
 
 from integration_tests import AgentTestWithPlugins
 from integration_tests.tests.utils import get_resource as resource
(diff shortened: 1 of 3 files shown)


tehasdf:
this out to fix `test_change_role_permissions` which has been failing on jenkins recently.

The repro was to run the alternate timezone test, then the snapshot test, and then that test, eg.
```
pytest -svx \
    integration_tests/tests/agent_tests/test_events.py \
    integration_tests/tests/agent_tests/test_snapshots.py \ 
    integration_tests/tests/agentless_tests/security/test_authorization.py
```

tehasdf:
alxo fixed the test to clean up after itself. The agentless version of that test already did it, so I've used the same code.

geokala (approved):

PR #3124, opened 2021-06-23
inte-tests: replace the amqp events follower with a db one

tehasdf:
Events/logs are now inserted via REST and not via AMQP, and so
the inte-tests follower/printer needs to follow the db, rather than
rabbitmq.


tehasdf:
the output is unfortunately a bit worse, eg.
```
	2021-06-23T10:48:46.102643	Relationships established
	2021-06-23T10:48:46.211598	Node instance started
	2021-06-23T10:48:46.326853	Task started 'testmockoperations.tasks.mock_establish'
	2021-06-23T10:48:46.40809 	Node instance started
	2021-06-23T10:48:46.563224	Task succeeded 'testmockoperations.tasks.mock_establish'
	2021-06-23T10:48:46.608085	Relationships established
	2021-06-23T10:48:46.648184	Node instance started
	2021-06-23T10:48:46.767649	'install' workflow execution succeeded
```
This is because some entries are missing in the db-side event, eg. the deployment id, we'd have to join it. And I'm not really willing to do it right now, just in the name of tests-only display.

If we at some point decide that the display is important, we can amend it.

geokala (approved):

PR #3132, opened 2021-06-28
RD-2660 Operations update: return 204

tehasdf:
The returned operation is unused anyway. Skipping the serialization
saves a few ms.


tehasdf:
this requires cloudify-cosmo/cloudify-common#813

geokala (approved):

PR #3133, opened 2021-06-28
RD-2661 Detach current_execution/current_tenant from the session

tehasdf:
See the docstring.

For now, only the most-heavily-used endpoints are going to use
this, and other endpoints can still benefit from auto-loading if
they so wish.

### DIFF
--- a/rest-service/manager_rest/rest/rest_decorators.py
+++ b/rest-service/manager_rest/rest/rest_decorators.py
@@ -582,7 +582,9 @@ def detach_globals(f):
     """
     @wraps(f)
     def wrapper(*args, **kwargs):
-        db.session.expunge(current_execution)
-        db.session.expunge(current_tenant)
+        if current_execution:
+            db.session.expunge(current_execution)
+        if current_tenant:
+            db.session.expunge(current_tenant)
         return f(*args, **kwargs)
     return wrapper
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/rest_decorators.py
```diff
@@ -566,3 +567,22 @@ def wrapper(*args, **kwargs):
...
         return f(*args, **kwargs)
     return wrapper
+
+
+def detach_globals(f):
+    """Detach current_execution and current_tenant from the db session.
+
+    This means the current_execution and current_tenant objects can be
+    used in multiple transactions, and they won't be automatically reloaded
+    from the db - saving a query, but not loading additional relationships.
```
geokala, line 577: Does this have any implications, e.g. around updates for objects being performed after an execution is cancelled or similar?

tehasdf: In principle I guess it could? This means we won't auto-refresh current_execution, but it is effectively snapshotted to the state it was at the beginning of the request. For the two endpoints I'm adding it to, it's precisely what we want (because we only access things like `._storage_id` from it, and it's silly that this access did a query)

geokala: OK, based on this and it needing to happen within the same session anyway (which I think shouldn't be doable from outside- e.g. via rest client), I'll green this.

geokala (approved):

geokala (approved):

PR #3143, opened 2021-07-01
RD-2687 Postgres.encrypt_values() will encrypt(None) = None

mateumann:


mateumann:
We've decided to solve it on `cloudify-common` level: https://github.com/cloudify-cosmo/cloudify-common/pull/816

PR #3149, opened 2021-07-07
snapshot-restore: avoid using the db when downgraded

tehasdf:
Spend as little time with the db/restservice in an unusable state, and
don't attempt to use it during that time.

See each commit for details.


workflows/cloudify_system_workflows/snapshots/snapshot_restore.py
```diff
@@ -543,6 +543,8 @@ def _restore_db(
             self._tempdir, schema_revision,
             premium_enabled=self._premium_enabled,
             config=self._config, snapshot_version=self._snapshot_version)
+        if not self._license_exists(postgres):
+            postgres.restore_license_from_dump(self._tempdir)
         if self._snapshot_version <= V_4_6_0:
```
geokala, line 550: Not sure if you remove this in a later commit but noting here that we don't support pre-5.0.5 any more so this bit can go.

geokala (approved):

PR #3164, opened 2021-07-16
fix the kwarg name

tehasdf:
...whoops


mateumann (approved):

tehasdf:
the arg is actually without the leading `_`: https://github.com/cloudify-cosmo/cloudify-common/blob/master/cloudify_rest_client/node_instances.py#L232

PR #3173, opened 2021-07-27
RD-2828 events list: default sort to timestamp

tehasdf:
No real reason to go by storage_id, which isn't technically guaranteed
to have any application-level meaning. I'm not sure this will fix the
inte-test which is only failing on jenkins, but it's a good idea to
do this anyway.


geokala (dismissed):

geokala:
Green - I'm not sure if we should pre-empt someone managing to get exactly the same timestamp?

geokala (approved):

PR #3191, opened 2021-08-05
inte-tests: dep-update tests - use common wait_for_exec

tehasdf:
replace the dep-update-only wait_for_execution functions, with the
common base class wait_for_execution_to_end instead

also, instead of wait_for_successful_state, assert it synchronously,
there's no reason a wait would be needed there


tests/integration_tests/tests/agentless_tests/deployment_update/test_deployment_update_addition.py
```diff
@@ -32,8 +32,8 @@ class TestDeploymentUpdateAddition(DeploymentUpdateBase):
     def test_add_node_bp(self):
         self._test_add_node(archive_mode=False)
 
-    def test_add_node_archive(self):
-        self._test_add_node(archive_mode=True)
+    # def test_add_node_archive(self):
```
geokala, line 35: Commented code?

geokala (approved):

PR #3199, opened 2021-08-10
RD-2926 snap-res: avoid using deployments_contexts

tehasdf:
Instead of implicit rest calls via workflow-context, use explicit rest calls here.
Those workflow-context ones were broken ever since we've made it cache
the rest-client, and so the internal._switch_tenant stopped working.
Instead of fixing it, let's drop it entirely.


workflows/cloudify_system_workflows/snapshots/utils.py
```diff
@@ -243,7 +243,9 @@ def get_manager_version(client):
     return ManagerVersion(client.manager.get_version()['version'])
 
 
-def get_tenants_list():
+def get_tenants_list(version):
+    if version < snapshot_constants.V_4_0_0:
```
geokala, line 247: I don't think anyone anywhere is on anything earlier than 4.3, so we can probably just drop this conditional.

geokala (approved):

PR #3204, opened 2021-08-13
RD-2832 Don't overwrite the RabbitMQ user on agent_create

mateumann:
In case `agent_create` request comes in with a `create_rabbitmq_user`
flag enabled, we attempt to create a separate RabbitMQ user for that
agent.  In case the account already existed, we would have overwritten
it thus breaking the tenant's or the manager's interaction with the
broker.  This patch prevents that from happening.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/agents.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/agents.py
@@ -176,8 +176,13 @@ class AgentsName(SecuredResource):
             new_agent = get_storage_manager().get(models.Agent, name)
 
         if request_dict.get('create_rabbitmq_user'):
-            # Create rabbitmq user
-            self._get_amqp_manager().create_agent_user(new_agent)
+            try:
+                # Create rabbitmq user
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/rest/resources_v3_1/agents.py
```diff
@@ -176,8 +176,13 @@ def put(self, name):
...
 
         if request_dict.get('create_rabbitmq_user'):
-            # Create rabbitmq user
-            self._get_amqp_manager().create_agent_user(new_agent)
+            try:
+                # Create rabbitmq user
+                self._get_amqp_manager().create_agent_user(new_agent)
+            except manager_exceptions.ConflictError:
+                # Assuming the agent user was already created
+                current_app.logger.info(f'Not creating agent user "{name}" '
```
geokala, line 184: This would probably be better as a 'Not creating agent user  "%s"', name to avoid wasting (a tiny amount of) time when not hitting the conflict path.

geokala (approved):

geokala (approved):

PR #3205, opened 2021-08-13
RD-2976 Compat with updating operation state for old plugins

tehasdf:
Old common expects a 200, not 204, from this endpoint


rest-service/manager_rest/rest/resources_v3_1/operations.py
```diff
@@ -114,7 +114,7 @@ def patch(self, operation_id, **kwargs):
                     instance.state in common_constants.TERMINATED_STATES:
                 self._modify_execution_operations_counts(instance, 1)
             sm.update(instance, modified_attrs=('state',))
-        return None, 204
+        return {}, 200
```
geokala, line 117: Well, this makes me sad.

geokala (approved):

rest-service/manager_rest/rest/resources_v3_1/operations.py
tehasdf: me too!

PR #3212, opened 2021-08-23
RD-3009 Audit log DB triggers

mateumann:


mateumann:
Not to be merged into 6.2

rest-service/manager_rest/security/audit.py
```diff
@@ -10,6 +11,8 @@ def set_audit_method(method: str):
 
 def set_username(username: str):
     g.audit_username = username
```
geokala: Do we still need to set it in the flask session if it's in the DB session already? (I'm not sure where else that's used)

geokala: That's what I get for reviewing this a commit at a time- I see you removed this on the next commit- ignore this!

resources/rest-service/cloudify/migrations/versions/03ad040e6f78_6_1_to_6.2.py
```diff
@@ -139,67 +128,89 @@ def _drop_audit_log_table():
...
+        END;
+    $$ LANGUAGE plpgsql;
+
+
+    CREATE OR REPLACE FUNCTION audit_execution_id() RETURNS TEXT AS $$
+        BEGIN
+            RETURN current_setting('audit.execution_id');
+        EXCEPTION WHEN syntax_error_or_access_rule_violation THEN
+            IF (SELECT audit_username() IS NULL) THEN
+                RETURN 'administrative task';
```
geokala: This might be better as something like 'system' or 'execution'- administrative task sounds like it could've been done by the admin user's request, not anyone else's.

geokala: I note there's a risk of colliding with a username as well, and I'm not sure we can reasonably insert characters that can't be in a username since we allow external auth providers.

resources/rest-service/cloudify/migrations/versions/03ad040e6f78_6_1_to_6.2.py
```diff
@@ -139,67 +128,89 @@ def _drop_audit_log_table():
...
             END IF;
             RETURN NULL;
         END;
-        $func$ LANGUAGE plpgsql;""")
+    $$ LANGUAGE plpgsql;""")
 
 
-def _drop_function_write_audit_log():
+def _drop_functions_write_audit_log():
     op.execute("""DROP FUNCTION write_audit_log();""")
```
geokala: Noting that this is missing the other DROP FUNCTION calls it should have, just in case you didn't include them in a later commit.

resources/rest-service/cloudify/migrations/versions/03ad040e6f78_6_1_to_6.2.py
mateumann: There's a good chance username will not be valid if it contained spaces, so, let's put some of those here, like `system task`.

geokala (approved):

PR #3216, opened 2021-08-23
RD-3009 Set audit.execution_id parameter when installing

mateumann:
... because otherwise `audit_log`'s `audit_log_creator_or_user_not_null`
constraint is not met.


mateumann:
A different approach will be implemented in https://github.com/cloudify-cosmo/cloudify-manager/pull/3212

PR #3218, opened 2021-08-23
RD-3031 Extract update steps in the new dep-update workflow

tehasdf:
Toy with the step-extractor a bit, and call it from the workflow.
Creating the steps is going to be an operation of its own.

### DIFF
--- a/workflows/cloudify_system_workflows/deployment_update/workflow.py
+++ b/workflows/cloudify_system_workflows/deployment_update/workflow.py
@@ -44,7 +44,7 @@ def create_steps(*, update_id):
     nodes = client.nodes.list(deployment_id=dep_up.deployment_id)
 
     # step-extractor expects workflows in this format - this is the same format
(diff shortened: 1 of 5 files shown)


workflows/cloudify_system_workflows/deployment_update/workflow.py
```diff
@@ -17,11 +21,58 @@ def create_update(*, update_id, new_inputs):
...
+
+def create_steps(*, update_id):
+    """Given a deployment update, extracts the steps for it"""
+    client = get_rest_client()
+    dep_up = client.deployment_updates.get(update_id)
+    deployment = client.deployments.get(dep_up.deployment_id)
+    nodes = client.nodes.list(deployment_id=dep_up.deployment_id)
+
+    # step-extractor expects workflows in this format - this is the same format
+    # as retuend by prepare_deployment_plan
```
geokala, line 47: Many happy retuens.

geokala (approved):

PR #3219, opened 2021-08-24
RD-3033 Add a script to truncate audit log

mateumann:


mateumann:
Not to be merged into 6.2

mateumann:
We're going with `cfy audit truncate` instead of this script

PR #3222, opened 2021-08-25
RD-3036 FastAPI /audit endpoint (1)

mateumann:


tehasdf (dismissed):

api-service/cloudify_api/config.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import BaseSettings
```
geokala, line 1: What does pydantic.BaseSettings buy us in exchange for having this extra dependency?

api-service/cloudify_api/main.py
```diff
@@ -0,0 +1,17 @@
+from cloudify_api import CloudifyAPI
+from cloudify_api.routers import audit as audit_router
+
+
+TITLE = "Cloudify API"
+VERSION = "6.3.0.dev1"
```
geokala, line 6: Can we pull the version string from pkgutils or something, to avoid having it hardcoded in yet another place?

api-service/cloudify_api/models.py
```diff
@@ -0,0 +1,16 @@
+from sqlalchemy import MetaData
+from sqlalchemy.ext.declarative import declarative_base
+
+from manager_rest.storage.resource_models import AuditLog  # noqa
+
+
+BaseModel = declarative_base(metadata=MetaData(naming_convention={
+    # This is to generate migration scripts with constraint names
```
geokala, line 8: Hopefully all the migrations go in the same place. I'm assuming we can't easily share models?

api-service/cloudify_api/config.py
tehasdf: pydantic is required by fastapi anyway; it's the model/schema library that it uses for parsing request bodies etc

api-service/cloudify_api/models.py
tehasdf: on second thought, perhaps making this depend on restservice is a really really bad idea

api-service/cloudify_api/config.py
mateumann: Yeah, so at the moment we're not leveraging the parsing/validation features of `BaseSettings`.  That's the plan for the future, when we get rid of `manager_rest.config` dependency.
About the cost, it's probably negligible, because pydantic is already fastapi's dependency, and we cannot get rid of it anyway.

api-service/cloudify_api/models.py
mateumann: I know, yagni (at least for some time in the future)…  I guess I was using it in `db_list` which is now gone.  The other argument was the "future self" of the code.  I will get rid of that.

api-service/cloudify_api/config.py
geokala: OK, sounds good then.

api-service/cloudify_api/models.py
geokala: I think perhaps making the restservice and this depend on shared DB models would be good though? Otherwise we could easily end up with situations where we have to write the same model in two different places.

geokala (approved):

PR #3227, opened 2021-08-31
RD-3086 Allow updating relationships for node-instances

tehasdf:
Other than changing node-instance runtime-props, we can now also change
relationships too.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v1/nodes.py
+++ b/rest-service/manager_rest/rest/resources_v1/nodes.py
@@ -27,6 +27,7 @@ from manager_rest.rest.rest_utils import (
     get_args_and_verify_arguments,
     get_json_and_verify_params,
     verify_and_convert_bool,
+    is_deployment_update
 )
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/rest/resources_v1/nodes.py
```diff
@@ -197,36 +197,36 @@ def get(self, node_instance_id, _include=None, **kwargs):
...
+                raise manager_exceptions.ConflictError(
+                    'Node instance update conflict [current version={0}, '
+                    'update version={1}]'.format(instance.version, version)
+                )
+            if 'runtime_properties' in request_dict:
+                instance.runtime_properties = request_dict['runtime_properties']
+            if 'state' in request_dict:
+                instance.state = request_dict['state']
+        if 'relationships' in request_dict:
+            instance.relationships = instance.relationships + \
```
tehasdf, line 230: don't `+`, just set

(needs fixing in the workflow too)

geokala (approved):

PR #3233, opened 2021-09-01
RD-3085 + RD-3086 dep-up: apply changes to nodes+instances

tehasdf:
Figure out what changes need to be made, and then actually apply them.
All in the workflow, using the new PATCH endpoints.

### DIFF
--- a/workflows/cloudify_system_workflows/deployment_update/workflow.py
+++ b/workflows/cloudify_system_workflows/deployment_update/workflow.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from cloudify.decorators import workflow
 from cloudify.state import workflow_ctx
 from cloudify.manager import get_rest_client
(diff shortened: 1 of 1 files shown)


workflows/cloudify_system_workflows/deployment_update/workflow.py
```diff
@@ -68,6 +70,41 @@ def create_steps(*, update_id):
...
+        parts = step['entity_id'].split(':')
+        node_id = parts[1]
+        entity_type = step['entity_type']
+        if entity_type in modified_entity_type:
+            modified[node_id].append(modified_entity_type[entity_type])
+        if entity_type == 'operation':
+            if 'relationships' in parts:
+                modified[node_id].append('relationships')
+            else:
+                modified[node_id].append('operations')
```
geokala, line 95: So if entity_type is not in [plugin, relationship, property, operation] then we silently fall through? Should we perhaps scream instead?

tehasdf: no, this implementation is not complete yet

geokala (approved):

PR #3236, opened 2021-09-03
RD-2998 Add csys-k8s-filter for blueprints

mateumann:
The filter relies on blueprint's `obj-type` label having a value of
`k8s` (possibly among some others).


qooban (approved):
LGTM 👍🏻

PR #3240, opened 2021-09-07
RD-3121 Don't block blueprint deletion when other are being parsed

mateumann:


tests/integration_tests/tests/agentless_tests/test_blueprint_upload.py
```diff
@@ -284,3 +287,65 @@ def _assert_cleanup(self, blueprint_id):
...
+        # assert `upload_blueprint` execution failure for the `main` blueprint
+        for execution in get_executions(
+                self.client,
+                workflow_id='upload_blueprint',
+                parameters__blueprint_id=main_blueprint_id):
+            for event in get_events(self.client,
+                                    execution_id=execution['id'],
+                                    event_type='workflow_failed'):
+                assert "Requested blueprint import `imported` was not found" \
+                       in event['message']
```
tehasdf, line 350: also/instead, assert on the execution status as well

tests/integration_tests/tests/utils.py
```diff
@@ -296,3 +296,34 @@ def wait_for_executions(cls):
         attribute = _wait_for_executions_after_call(attribute)
         setattr(cls, name, attribute)
     return cls
+
+
+def get_executions(client, **kwargs):
+    for execution in client.executions.list(_get_all_results=True):
+        if _dict_matches(execution, **kwargs):
+            yield execution
```
tehasdf, line 304: tbh i'd prefer to avoid having a `_dict_matches` like that:
- `executions.list()` i'm pretty sure already would accept a `workflow_id='..'` filter?
- the parameters check would just be a simple `if` in the loop
- the events kwargs i think also can be just filters to `client.events.list()` already?

tests/integration_tests/tests/utils.py
```diff
@@ -296,3 +296,34 @@ def wait_for_executions(cls):
...
+    # Check if `d` dictionary matches `**kwargs`, e.g. if `d` is
+    # {'a': 1, 'b': 2}, then _dict_matches(d, a=1) returns True, and
+    # _dict_matches(d, c=3) returns False.  Nested matching is supported, e.g.:
+    # for d = {'a': {'q': 0, 'w': 1}, 'b': 'foo'}, _dict_matches(d, a__w==1)
+    # will return True.
+    for k, v in kwargs.items():
+        try:
+            head, tail = k.split('__', 1)
+        except ValueError:
+            head, tail = k, None
```
tehasdf, line 323:
```suggestion
        head, _, tail = k.partition('__')
```

tests/integration_tests/tests/agentless_tests/test_blueprint_upload.py
```diff
@@ -284,3 +287,65 @@ def _assert_cleanup(self, blueprint_id):
...
+        self.client.blueprints.upload(resource(imported_blueprint_path),
+                                      entity_id=imported_blueprint_id)
+        wait_for_blueprint_upload(imported_blueprint_id, self.client)
+        main_blueprint_id = 'main'
+        main_blueprint_path = \
+            os.path.join('dsl', 'simple_deployment_with_import.yaml')
+        self.client.blueprints.upload(resource(main_blueprint_path),
+                                      entity_id=main_blueprint_id,
+                                      async_upload=True)
+        wait_for_blueprint_upload(main_blueprint_id, self.client)
```
tehasdf, line 321: surely this is equivalent to just not having async_upload, and then you dont need the wait()?

tehasdf (commented):
so we dont actually need to really add anything, just remove the pre-upload check and thats it? nice

tests/integration_tests/tests/utils.py
mateumann: I've updated both `get_e*` functions.  As for the `_dict_matches` – I wanted to avoid `if`s in the test function body.  I guess it's just an aesthetic preference, but I kind of like it this way.

tests/integration_tests/tests/agentless_tests/test_blueprint_upload.py
mateumann: Yeah, that's updated too (looked at it with fresh eyes this morning)

tehasdf (approved):

geokala (approved):

PR #3243, opened 2021-09-09
CYBL-1248 Handle invalid blueprint syntax

mateumann:
... in case we're unable to determine `imports:` line position when
parsing a blueprint in substitute-import-lines scripts.


geokala:
We can't re-release 5.2.5, it's already in use for a major upgrade and the other patches we need to provide will need to be applied as updates to the managers that have already been upgraded.
Also, re-releasing something we already put in the wild is a terrible thing to think of as acceptable anyway.

PR #3259, opened 2021-09-15
RD-3168 lock db tables on restore

mateumann:


tehasdf (approved):

workflows/cloudify_system_workflows/snapshots/postgres.py
```diff
@@ -527,7 +527,13 @@ def _get_clear_tables_queries(self, snapshot_version,
         all_tables = [table for table in all_tables if
                       table not in self._TABLES_TO_KEEP]
 
+        tables_to_lock = ['users', 'roles']
+        if snapshot_version > V_4_6_0:
```
geokala, line 531: No need, we don't support anything before that on 6.3+ anyway.

geokala: Actually, I think we might still want 4.6 in some edge cases so let's leave this in.

PR #3264, opened 2021-09-16
RD-3183 Refactor cloudify_types/

tehasdf:
Starting with a major refactor of `cloudify_types/` so that I can start
fixing it without going insane.

So this starts with moving things out of the classes, just into the
operation functions directly. All the `inline` commits are about
taking one method of a Component class, and just copying over
into the operation directly.

This means the code is simpler, each operation just does what is
written right there in the operation function, we drop the whole
concept of whatever "proxy_operation" was, and life is great.

Then, some more refactors.


cloudify_types/cloudify_types/component/operations.py
```diff
@@ -124,10 +136,244 @@ def upload_blueprint(ctx, **kwargs):
     return True
 
 
+def _verify_secrets_clash(client, secrets):
```
geokala, line 133: This should probably be called _abort_if_secrets_clash- if it was really verifying that secrets clash I'd expect it to raise an exception if they didn't.

PR #3265, opened 2021-09-17
RD-3118 Audit streaming

mateumann:


api-service/cloudify_api/listener.py
```diff
@@ -0,0 +1,70 @@
...
+class Listener:
+    def __init__(self, dsn: str, logger: logging.Logger):
+        self.dsn = dsn
+        self.logger = logger
+        self.conn_listen = None
+        self.loop = asyncio.get_event_loop()
+        self.channels = {}
+
+    def listen_on_channel(self, channel: str):
+        if channel in self.channels.keys():
```
tehasdf, line 31:
```suggestion
        if channel in self.channels:
```
(ditto several more places)

resources/rest-service/cloudify/migrations/versions/8e8314b1d848_6_2_to_6_3.py
```diff
@@ -259,3 +261,24 @@ def _add_audit_log_indexes():
 def _drop_audit_log_indexes():
     op.drop_index(op.f('audit_log_creator_name_idx'), table_name='audit_log')
     op.drop_index(op.f('audit_log_execution_id_idx'), table_name='audit_log')
+
+
+# flake8: noqa
```
geokala, line 266: Please don't do a bare noqa, it should be targeted to the line it is intended to target.

api-service/cloudify_api/listener.py
```diff
@@ -0,0 +1,70 @@
...
+        if channel not in self.channels.keys():
+            raise TaskNotDefined(channel)
+        self.channels[channel]['queues'].append(queue)
+
+    def remove_queue(self, channel: str, queue: asyncio.Queue):
+        if channel not in self.channels.keys():
+            raise TaskNotDefined(channel)
+        try:
+            self.channels[channel]['queues'].remove(queue)
+            self.logger.debug(f"Queue {queue} removed from channel {channel}.")
```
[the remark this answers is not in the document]
geokala: (the same applies to the other logger lines, including .info ones)

geokala:
Presumably the approach here means that if someone requests to stream from a point in the past, we'll grab all logs from the past and attach a listener to get subsequent ones (sounds like opportunity for racing there though)?

api-service/cloudify_api/listener.py
```diff
@@ -0,0 +1,70 @@
...
+        self.conn_listen = None
+        self.loop = asyncio.get_event_loop()
+        self.channels = {}
+
+    def listen_on_channel(self, channel: str):
+        if channel in self.channels.keys():
+            raise TaskAlreadyExists(channel)
+        self.channels[channel] = {
+            'task': self.loop.create_task(self._listener(channel)),
+            'queues': [],
```
tehasdf, line 35: I'd make this a set, for faster .remove() (or .discard) if there's thousands of connected clients

api-service/cloudify_api/routers/audit.py
```diff
@@ -58,6 +61,26 @@ async def list_audit_log(
...
+    app.logger.warning("stream_audit_log queue=%s", queue)
+    return StreamingResponse(audit_log_streamer(queue, app))
+
+
+async def audit_log_streamer(queue: asyncio.Queue, app):
+    app.logger.warning("audit_log_streamer queue=%s", queue)
+    while True:
+        record = await queue.get()
+        app.logger.warning("audit_log_streamer record=%s", record)
+        response = f"{json.dumps(record)}\n".encode('utf-8', errors='ignore')
```
tehasdf, line 79: Needs double `\n` I think

tehasdf (commented):
still needs:
[] removing on disconnect
[] set the content-type to text/event-stream

api-service/cloudify_api/config.py
```diff
@@ -3,8 +3,9 @@
 
 class Settings(BaseSettings):
     cloudify_rest_config_file: str = '/opt/manager/cloudify-rest.conf'
-    sqlalchemy_database_dsn: str = ''
+    sqlalchemy_dsn: str = ''
     sqlalchemy_engine_options: dict = {}
+    asyncpg_dsn: str = ''
```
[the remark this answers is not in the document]
mateumann: `None` is not a valid `str`.  In which case we can leave it as is or define as `Optional[str]`.

resources/rest-service/cloudify/migrations/versions/8e8314b1d848_6_2_to_6_3.py
mateumann: I've reformatted the string in triple quotes (there was no way to put `# noqa` for the specific line since the triple quotes)

api-service/cloudify_api/listener.py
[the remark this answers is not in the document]
mateumann: Thanks, did that

api-service/cloudify_api/config.py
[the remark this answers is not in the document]
mateumann: Changed to `Optional[str]`

api-service/cloudify_api/listener.py
mateumann: Good point.  Done.  I left the `.remove()` because I think that might be interesting to log such situations.

api-service/cloudify_api/routers/audit.py
mateumann: Done

tehasdf (approved):

PR #3269, opened 2021-09-20
RD-3221 execution resume: only set pending, not started

tehasdf:
It's not quite started yet - it'll be started once it gets picked
up by the mgmtworker.

Also drop the additional .update(), it's not needed, because we commit
anyway - and it's not even all the attrs that changed (anymore)
(because now .render_message is what sets the new token)


tehasdf:
The other side of this is cloudify-cosmo/cloudify-common#886

mateumann (approved):

PR #3285, opened 2021-09-30
RD-3108 new dep-update: update deployment inputs

tehasdf:
Gotta handle no new-blueprint, because only the inputs might have
changed. In that case, use the old blueprint, and still update
the inputs.


tehasdf:
This will be switched over to master once #3284 is merged

geokala (approved):

PR #3287, opened 2021-10-05
RD-3161 Make resource_manager.cancel_execution readable

mateumann:


rest-service/manager_rest/resource_manager.py
```diff
@@ -1272,14 +1272,34 @@ def cancel_execution(self, execution_ids, force=False, kill=False):
...
         """
-        if kill:
-            force = True
-        if kill:
-            new_status = ExecutionState.KILL_CANCELLING
-        elif force:
-            new_status = ExecutionState.FORCE_CANCELLING
-        else:
-            new_status = ExecutionState.CANCELLING
+        def _new_status_force(_force: bool, _kill: bool) -> (str, bool):
```
geokala, line 1275: I'm not convinced the closure makes things more readable here compared to just cleaning up the if statements- specifically, just removing the second 'if kill' would probably make the entire thing clean?

geokala (approved):

PR #3295, opened 2021-10-11
RD-2996 Drop manager_rest dependency in integration tests

mateumann:


tests/integration_tests/tests/agent_tests/test_snapshots.py
```diff
@@ -18,7 +18,6 @@
 import pytest
 
 from integration_tests import AgentTestCase
-from cloudify.models_states import AgentState
```
tehasdf, line 21: this isn't `manager_rest`, it's c-common ;P

PR #3296, opened 2021-10-12
RD-2958 Clean up dependencies

mateumann:


rest-service/dev-requirements.txt
```diff
@@ -1,6 +1,32 @@
 git+https://github.com/cloudify-cosmo/cloudify-common@master#egg=cloudify-common[dispatcher]
-
-# For dealing with the binary leftovers of psycopg2 in the 2.7.x version
-psycopg2==2.7.4 --no-binary psycopg2
 # To match other repos' requirements
 jinja2>=2.10,<2.11
+
+
+
+Flask>1,<2
```
tehasdf, line 7: loose specifiers like `>1,<2` should go into setup.py's install_requires - but reqs.txt needs a concrete version, eg. `Flask==1.0.0`
(ditto for multiple entries in this file)

this is so that installing from reqs.txt is always 100% reproducible, ie. the version that we put here, is the one that's going to be installed, period - currently, if flask gets a new release, it'll still be updated

a simple way to do it, is to `pip install .` into a clean virtualenv, and get the current versions by `pip freeze > reqs.txt` (or pip-compile, which could also give you even sha256's of the packages!)

geokala (approved):

PR #3307, opened 2021-10-29
RD-2615 Deployment update: IDDs in the workflow

mateumann:


tehasdf (commented):
it's fine, but fix the tests; tbh just remove the log i think. Unfortunate that the tests dont set ctx, but, really, not worth a log...

tehasdf (approved):

PR #3308, opened 2021-11-01
RD-2869 Add is_prehashed flag to user create

glukhman:


rest-service/manager_rest/rest/resources_v3/users.py
```diff
@@ -85,17 +85,24 @@ def put(self, multi_tenancy):
                 },
             }
         )
-
+        is_prehashed = request_dict.get('is_prehashed')
+        if is_prehashed is not None:
+            is_prehashed = rest_utils.verify_and_convert_bool(
+                'is_prehashed', is_prehashed)
+            del request_dict['is_prehashed']
```
geokala, line 92: Maybe better would be is_prehashed = rest_utils.verify_and_convert_bool('is_prehashed', request_dict.pop('is_prehashed', False))?

geokala (approved):

PR #3317, opened 2021-11-09
restservice log_request: omit form data

tehasdf:
We don't keep anything interesting in the form data anyway, we almost
always use json instead. (which is request.json() rather
than request.form)

However, accessing request.form (which we did for
_every single request_) breaks one important usecase: the fileserver
(and cloudify-api) auth endpoint.

For auth_request, nginx sends through all the headers, but not the
request body (because the internal auth request is a GET request).

That means the auth request gets a non-zero content-length but an
empty request body, and so accessing `request.form` waits forever
for the body. If content-type was application/json, then accessing
request.form would not wait, because in that case flask always just
returns empty for .form

### DIFF
--- a/rest-service/manager_rest/app_logging.py
+++ b/rest-service/manager_rest/app_logging.py
@@ -36,27 +36,19 @@ def setup_logger(logger):
 
 
 def log_request():
-    # form and args parameters are "multidicts", i.e. values are not
-    # flattened and will appear in a list (even if single value)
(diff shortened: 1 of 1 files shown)


tehasdf:
(conversely, if we logged `request.json()`, then it WOULD block in case of `content-type: application/json`)

geokala (approved):

PR #3337, opened 2021-11-18
restservice tests: drop disused old-client tests

tehasdf:
The old-client tests haven't been run in years, and they are the reason
we keep all the `@attr` & `pytest -a` around.
Time to get rid of it!


geokala (approved):

tehasdf:
```(vv3) ➜  rest-service git:(master) ✗ pytest  -a '(client_min_version is None or client_min_version <= 3.1) and (client_max_version is None or 3.1 <= client_max_version)' manager_rest --collect-only | grep 'tests collected'

=============== 903/922 tests collected (19 deselected) in 1.28s ===============
```
after
```
(vv3) ➜  rest-service git:(remove-test-attr-old-ver-1) ✗ pytest -svx manager_rest --collect-only | grep 'tests collected'                                                                                                                         
========================= 903 tests collected in 1.50s =========================
```

same amount, I didnt drop any more than I wanted to, I think.

PR #3338, opened 2021-11-19
Events/logs: allow passing in timestamp

tehasdf:
`timestamp` is actually passed in normally by the client.
`reported_timestamp` is the time of inserting into the db.
Is it named the wrong way around? Perhaps.


tehasdf:
it is passed in here: https://github.com/cloudify-cosmo/cloudify-common/blob/3f1605ee6e2286db3a93da3f91f7a6dcb2d5641b/cloudify/logs.py#L306-L307

tehasdf:
Should we only have one field instead of two? Also perhaps.

geokala (approved):

PR #3340, opened 2021-11-22
New dep-up: handle missing default_schedules

tehasdf:
The field might well not be there, let's not break when that happens.


mateumann (approved):

workflows/cloudify_system_workflows/deployment_update/workflow.py
```diff
@@ -452,11 +452,18 @@ def delete_removed_nodes(*, update_id):
...
     dep_up = client.deployment_updates.get(update_id)
+    new_schedules = (
+        dep_up.deployment_plan
+        .get('deployment_settings', {})
+        .get('default_schedules')
+    )
+    if not new_schedules:
+        # we are not going to be deleting schedules that are now missing,
+        # because we can't tell if they were added by the user explicitly
+        return
```
tehasdf, line 463: not _strictly_ related but might as well return early to save some time.

PR #3348, opened 2021-11-25
RD-3573 Recursive query for dependency retrieval

tehasdf:
Note: this just implements the query, doesn't actually use it
yet - that'll come in the next PRs.

This is a way to retrieve a whole sub-tree of dependent
deployments, as opposed to the current approach, which
selects ALL THE DEPENDENCIES EVER and then formats them
into a tree.

There's a generic query on the InterDepDeps class itself,
and then a bunch of frontend interface methods on the
Deployment model.

### DIFF
--- a/rest-service/manager_rest/storage/resource_models.py
+++ b/rest-service/manager_rest/storage/resource_models.py
@@ -2022,6 +2022,66 @@ class BaseDeploymentDependencies(CreatedAtMixin, SQLResourceBase):
     source_deployment_id = association_proxy('source_deployment', 'id')
     target_deployment_id = association_proxy('target_deployment', 'id')
 
+    @classmethod
+    def _dependencies_adjacency(cls, deployment_ids, dependents=True):
+        """Select a dependency subgraph in an adjacency-list form.
+
+        Returns a query yielding (idd_id, source_id, target_id), which mean
+        "IDD with id idd_id, says that deployment source_id depends on the
+        deployment target_id".
+
+        This is recursive, so will return parents, then parents of parents,
+        etc. (or children, then children of children, etc).
+
+        :param deployment_ids: storage_ids of the root deployments
+        :param dependents: if set, return dependents, ie. children; otherwise,
+            return dependencies, ie. parents
+        """
+        select_cols = db.session.query(
+            cls._storage_id,
+            cls._source_deployment,
+            cls._target_deployment,
+        )
+
+        if dependents:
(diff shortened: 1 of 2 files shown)


tehasdf:
there will be more based off of this, so make sure you understand this well, when reviewing

geokala (approved):

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -2024,6 +2093,70 @@ def target_deployment(cls):
...
+            cls._target_deployment,
+        )
+
+        if dependents:
+            base = select_cols.filter(
+                cls._target_deployment.in_(deployment_ids))
+        else:
+            base = select_cols.filter(
+                cls._source_deployment.in_(deployment_ids))
+        base = base.cte(name='dependents', recursive=True)
```
tehasdf, line 2123: for a recap of recursive CTEs, check out https://www.postgresql.org/docs/14/queries-with.html

tehasdf:
So this query is going to be something like:
```sql
WITH RECURSIVE dependents(_storage_id, _source_deployment, _target_deployment) AS 
    (
        SELECT
            inter_deployment_dependencies._storage_id AS _storage_id,
            inter_deployment_dependencies._source_deployment AS _source_deployment,
            inter_deployment_dependencies._target_deployment AS _target_deployment
        FROM inter_deployment_dependencies 
        WHERE inter_deployment_dependencies._source_deployment IN (%(source_deployment_1)s)
    UNION SELECT
            inter_deployment_dependencies._storage_id AS inter_deployment_dependencies__storage_id, 
            inter_deployment_dependencies._source_deployment AS inter_deployment_dependencies__source_deployment,
            inter_deployment_dependencies._target_deployment AS inter_deployment_dependencies__target_deployment
        FROM inter_deployment_dependencies
        JOIN
            dependents ON inter_deployment_dependencies._source_deployment = dependents._target_deployment
    )
SELECT deployments._storage_id, [.... snip .... ]
FROM deployments
JOIN dependents ON deployments._storage_id = dependents._target_deployment
```
pay attention to the recursive part: we first select the root (the first `SELECT`, the one that is `WHERE _source in %()`, that is parametrized by the arguments passed in to the query), and then we join to that, the IDDs that target the result, as specified by the `JOIN .. ON` condition in the `UNION SELECT`. It is there that we give the rule for recursion: the next level's `source` must be the previous level's `target`
(or the other way around, depending which way do we want to go)
Then the database goes around, and joins to that again the IDDs that target the result, and again, and again.

PR #3350, opened 2021-11-25
RD-3573 Retrieve all data needed to recalc a tree

tehasdf:
This isn't used yet, but it will be (in the next PR, or one of them).

This adds a query that fetches all the data needed to recompute
a subtree: ie. given a deployment storage id, it will fetch details
required to compute that deployment's status(es), and those of
its parents, and those parents' parents, etc.

I'm keeping this query semi-private, because of how specific it
is, and not very useful for anything else than the one call site
(that it is yet to have).

It is a bit unfortunate that this writes the whole query from
scratch, rather than using the other methods, but I've tried,
and I haven't been able to do it in a way that made anything
cleaner.

### DIFF
--- a/rest-service/manager_rest/storage/resource_models.py
+++ b/rest-service/manager_rest/storage/resource_models.py
@@ -2157,6 +2157,99 @@ class BaseDeploymentDependencies(CreatedAtMixin, SQLResourceBase):
             query = query.with_for_update()
         return query.all()
 
+    @classmethod
+    def _get_update_tree(cls, deployment_ids):
+        """Fetch just enough data to recalculate statuses for a subtree.
+
+        To recalculate the status/sub_counts of each ancestor, we must
+        select all the ancestors in the tree, and all of their direct
+        children: for each ancestor, we must know how many sub-deployments
+        it has, and what are their statuses.
+
+        Eg. when computing the update-tree of a deployment, we fetch that
+        deployment's parent, and ALL of the children of that parent, ie.
+        all of the target deployment's siblings.
+
+        This method is very closely tied to its usage: we only select the
+        columns that are necessarily required; it's only useful for
+        recalculating the status of a subtree of deployments, PLEASE
+        do not use this for anything else.
+
+        Returns a a list of rows that contain:
+            - several columns from each deployment
+            - the edges, ie. source_id/target_id
+            - if the deployment is an environment
+            - the latest execution status for each deployment
+        """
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/storage/resource_models.py
```diff
@@ -2157,6 +2157,99 @@ def get_dependencies(cls, deployments, dependents=True, locking=False,
...
+                # these columns just happen to be enough to recalculate
+                # the statuses and counts; we avoid selecting anything
+                # that isn't necessarily required
+                Deployment._storage_id,
+                Deployment.deployment_status,
+                Deployment.installation_status,
+                Deployment.sub_services_count,
+                Deployment.sub_environments_count,
+                Deployment.sub_services_status,
+                Deployment.sub_environments_status,
```
tehasdf, line 2224: in principle we could select the whole deployment here, but eh. This is enough.

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -2157,6 +2157,99 @@ def get_dependencies(cls, deployments, dependents=True, locking=False,
...
+                depends.c.dependents__target_deployment.label(
+                    '_target_deployment'),
+                depends.c.dependents__source_deployment.label(
+                    '_source_deployment'),
+                db.session.query(DeploymentLabel.id)
+                .filter_by(_labeled_model_fk=Deployment._storage_id)
+                .filter_by(key='csys-obj-type')
+                .filter_by(value='environment')
+                .exists()
+                .label('is_env'),
```
tehasdf, line 2234: yes this is essentially a join of labels

rest-service/manager_rest/test/endpoints/test_inter_deployment_dependencies.py
```diff
@@ -62,6 +62,167 @@ def _label_dependency(self, source, target):
...
+        assert {r._storage_id for r in update_tree} == \
+            {d1._storage_id, d2._storage_id, d3._storage_id}
+        assert {
+            (r._source_deployment, r._target_deployment) for r in update_tree
+        } == {
+            (d1._storage_id, d2._storage_id),
+            (d3._storage_id, d2._storage_id)
+        }
+
+    def test_tree(self):
```
tehasdf, line 137: pay attention to this test, it shows explicitly what is it that we select

geokala (approved):

PR #3351, opened 2021-11-26
RD-3507 Transfer node instances

geokala:


rest-service/manager_rest/rest/resources_v3_1/nodes.py
```diff
@@ -131,36 +135,71 @@ def post(self):
...
+            raw_instance['_node_fk'] = nodes[raw_instance.pop('node_id')]
+            creator = raw_instance.get('creator')
+            if creator:
+                if creator in user_lookup:
+                    user = user_lookup[creator]
+                else:
+                    rest_utils.valid_user(creator)
+                    user = sm.get(models.User, None, include=['id'],
+                                  filters={'username': creator})
+                    user_lookup[creator] = user
```
tehasdf, line 184:
```suggestion
                if creator not in user_lookup:
                    user_lookup[creator] = rest_utils.valid_user(creator)
                user = user_lookup[creator]
```
The caching stuff, I prefer to write reversed like this, with a `not in`, I think it's nicer, no? To me, it reads like a "fetch from cache, or otherwise if there's a cache miss, then do this", which underlines that cache misses are in principle rare.

Additionally, I note that `valid_user` already returns a user

rest-service/manager_rest/rest/resources_v3_1/nodes.py
```diff
@@ -131,36 +135,71 @@ def post(self):
...
+
+        valid_params = ['id', 'runtime_properties', 'state', 'version',
+                        'relationships', 'scaling_groups', 'host_id',
+                        'index', 'visibility',
+                        '_tenant_id', '_node_fk', '_creator_id']
+
+        user_lookup = {}
+
         for raw_instance in raw_instances:
             node_id = raw_instance['node_id']
```
tehasdf, line 170: you could just `.pop()` here

tehasdf (dismissed):

rest-service/manager_rest/rest/resources_v3_1/nodes.py
geokala: Good point, missed that when I was merging this stuff together.

tehasdf (approved):

PR #3352, opened 2021-11-26
RD-3573 Recalculate all ancestors sub-counts&statuses 

tehasdf:
A single tree-traversing function to recalculate all statuses & counts
for parents of the given deployments.

This uses the `_get_update_tree` query from before, and it is not
yet used itself, but it will be!

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -63,6 +63,65 @@ from . import workflow_executor
 from . import manager_exceptions
 
 
+class _SubDepSummary(object):
+    """A container for sub-counts and sub-statuses of a deployment.
+
+    This is only to be used by the recalc_ancestors machinery. This
+    class defines how counts and statuses should be added together.
+    """
+    __slots__ = (
+        'sub_services_count',
+        'sub_environments_count',
+        'sub_services_status',
+        'sub_environments_status'
+    )
+
+    def __init__(
+        self,
+        *,
+        sub_services_count=0,
+        sub_environments_count=0,
+        sub_services_status=None,
+        sub_environments_status=None,
+    ):
+        self.sub_services_count = sub_services_count
+        self.sub_environments_count = sub_environments_count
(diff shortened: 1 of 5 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -63,6 +63,65 @@
...
+            other.sub_environments_count,
+            sub_services_status=models.Deployment.compare_statuses(
+                self.sub_services_status, other.sub_services_status
+            ),
+            sub_environments_status=models.Deployment.compare_statuses(
+                self.sub_environments_status, other.sub_environments_status
+            )
+        )
+
+    def __iadd__(self, other):
```
tehasdf, line 106: we could let python fall back to just `__add__`, but this is an optimization, it'll allow `a += b` to not create an additional object

rest-service/manager_rest/resource_manager.py
```diff
@@ -2640,6 +2699,142 @@ def _get_utc_datetime_from_sched_plan(time_expression, base_datetime=None):
             return parse_utc_datetime_relative(time_expression, base_datetime)
         return datetime.strptime(time_expression, time_fmt)
 
+    def _reset_sub_deployments(self, deployment_ids):
+        """Set sub-counts to 0 and sub-statuses to None, for deployment_ids.
+
+        We know that deployment_ids have no children, so we can clear their
```
geokala, line 2705: I assume this is intended to say that "This should only be called on deployment_ids that we know to have no children"?

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -572,6 +572,8 @@ def compare_between_statuses(
         }
         return max(statuses, key=lambda st: importance.get(st, 0))
 
+    compare_statuses = compare_between_statuses
```
geokala, line 575: Any reason not to do this as an actual rename?

geokala (approved):

PR #3354, opened 2021-11-26
RD-3753 Run recalc_ancestors rather than update_exec_status

tehasdf:
Instead of recalcing the ancestors using the old way,
in update_exec_status, run the new method.
This is just the first call-site of recalc_ancestors, but the most
important one, because this runs more often than other places

delete_deployment is going to return the affected parent ids, but
it doesn't just yet.

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -177,17 +177,15 @@ class ResourceManager(object):
         dep.latest_execution = latest_execution
         dep.deployment_status = dep.evaluate_deployment_status()
         self.sm.update(dep)
-        if dep.deployment_parents:
-            graph = RecursiveDeploymentLabelsDependencies(self.sm)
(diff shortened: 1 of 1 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -177,17 +177,15 @@ def update_deployment_statuses(self, latest_execution):
         dep.latest_execution = latest_execution
         dep.deployment_status = dep.evaluate_deployment_status()
         self.sm.update(dep)
-        if dep.deployment_parents:
```
geokala, line 180: I take it this doesn't need to happen because of some part of the previous PR I already forgot?

tehasdf: it doesn't need to happen because of THIS pr, `graph.propagate_deployment_statuses(dep.id)` ~ `recalc_ancestors()`

geokala (approved):

PR #3356, opened 2021-11-27
Rewrite test_deployment_labels_dependencies

tehasdf:
This makes this test module WAY faster. Like ten times.

Eventually I'd like to see all the tests rewritten this way.

### DIFF
--- a/rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
+++ b/rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
@@ -284,7 +284,7 @@ class DeploymentLabelsDependenciesTest(BaseServerTestCase):
 
         exc = models.Execution(
             id='exc1',
-            deployment=dep1,
+            _deployment_fk=dep1._storage_id,
             workflow_id='create_deployment_environment',
         )
         db.session.add(exc)
@@ -309,7 +309,7 @@ class DeploymentLabelsDependenciesTest(BaseServerTestCase):
         for dep in [env1, env2, srv1]:
             exc = models.Execution(
                 id=f'create_{dep.id}',
-                deployment=dep,
+                _deployment_fk=dep._storage_id,
                 workflow_id='create_deployment_environment',
             )
             db.session.add(exc)
--- a/rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
+++ b/rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
@@ -286,6 +286,8 @@ class DeploymentLabelsDependenciesTest(BaseServerTestCase):
             id='exc1',
             _deployment_fk=dep1._storage_id,
             workflow_id='create_deployment_environment',
+            tenant=self.tenant,
+            creator=self.user,
         )
         db.session.add(exc)
         self.client.executions.update(exc.id, status='terminated')
@@ -311,6 +313,8 @@ class DeploymentLabelsDependenciesTest(BaseServerTestCase):
                 id=f'create_{dep.id}',
                 _deployment_fk=dep._storage_id,
                 workflow_id='create_deployment_environment',
+                tenant=self.tenant,
+                creator=self.user,
             )
             db.session.add(exc)
             self.client.executions.update(exc.id, status='terminated')
--- a/rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
+++ b/rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
@@ -1,613 +1,420 @@
-
-from mock import patch
-
 from cloudify_rest_client.exceptions import CloudifyClientError
 from cloudify.models_states import DeploymentState
 
-from manager_rest.rest.rest_utils import RecursiveDeploymentLabelsDependencies
+from manager_rest.storage import models, db
 from manager_rest.test.base_test import BaseServerTestCase
 
 
 class DeploymentLabelsDependenciesTest(BaseServerTestCase):
 
-    def _create_deployment_objects(self, parent_name, deployment_type, size):
-        for service in range(1, size + 1):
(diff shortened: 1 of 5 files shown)


rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -1,613 +1,420 @@
...
-    @patch('manager_rest.resource_manager.ResourceManager'
-           '.handle_deployment_labels_graph')
-    @patch('manager_rest.resource_manager.ResourceManager'
-           '.verify_attaching_deployment_to_parents')
-    def test_deployment_with_non_parent_labels(self,
-                                               verify_parents_mock,
-                                               handle_labels_graph_mock):
-        self.put_deployment_with_labels([{'env': 'aws'}, {'arch': 'k8s'}])
-        verify_parents_mock.assert_not_called()
-        handle_labels_graph_mock.assert_not_called()
```
tehasdf, line 74: removing those because they're tightly coupled to methods that I'm going to be removing in RD-3573

Like, no need for a test that only checks that something _wasn't_ called.

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -1,613 +1,420 @@
...
+                'dep',
+                labels=[{'csys-obj-parent': 'parent'}]
             )
 
-    def test_upload_blueprint_with_valid_parent_id_on_dsl(self):
-        self.put_deployment('valid-id')
-        self.put_blueprint(
-            blueprint_id='bp1',
-            blueprint_file_name='blueprint_with_valid_parent_labels.yaml'
-        )
```
tehasdf, line 99: the `on_dsl` tests, which test stuff based on a blueprint with an actual .yaml file, NEED TO GTFO.
If you want to test the whole flow, from a .yaml, to behaviour, there's inte-tests for that.

On this level, we just check the behaviour of specific rest endpoints.

What happens with a yaml file, well, that's to be defined by the dsl-parser, and the create-dep-env workflow.
In another commit in this PR, I add tests for create-dep-env, who check that the labels from the dsl, are passed to the restservice. And since we know they're indeed passed in, here we just need to check the behaviour of that endpoint.

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -1,613 +1,420 @@
...
-                        'csys-obj-parent': 'parent_1'
-                    },
-                    {
-                        'csys-obj-parent': 'notexist'
-                    }
-                ],
-                resource_id='invalid_label_dep'
+        parent = self._deployment(id='parent')
+        self._deployment(id='dep')
+        with self.assertRaisesRegex(CloudifyClientError, 'does not exist'):
```
tehasdf, line 67: this will be later (not in this pr) replace with a check on a 404 exception class

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -1,613 +1,420 @@
...
-        )
-        self.client.deployment_updates.update_with_existing_blueprint(
-            'sub_env', blueprint_id='update_sub_env'
-        )
-        deployment = self.client.deployments.get('env')
-        deployment_1 = self.client.deployments.get('env_1')
-        self.assertEqual(deployment.sub_services_count, 1)
-        self.assertEqual(deployment.sub_services_count, 1)
-        self.assertEqual(deployment_1.sub_environments_count, 1)
-        self.assertEqual(deployment_1.sub_environments_count, 1)
```
tehasdf, line 325: deployment-update tests can f.. right off. Especially ones that use .yaml blueprints, on this level. If you want to test that, that goes in inte-tests.

Deployment update (the new one) works in terms of updating labels via the restservice, anyway, which is covered by all the other tests already.

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -1,613 +1,420 @@
...
+        exc = models.Execution(
+            id='exc1',
+            deployment=dep1,
+            workflow_id='create_deployment_environment',
         )
-        self.assertIsNone(deployment.sub_services_status)
-        self.assertIsNone(deployment.sub_environments_status)
+        db.session.add(exc)
+        self.client.executions.update(exc.id, status='terminated')
+        assert dep1.deployment_status == DeploymentState.GOOD
```
tehasdf, line 292: adding this check too, for no good reason, but might as well - check that it's require_attn up front, and `good` after running create-dep-env.

rest-service/manager_rest/test/endpoints/test_deployment_labels_dependencies.py
```diff
@@ -1,613 +1,420 @@
...
         )
-        dep_graph = RecursiveDeploymentLabelsDependencies(self.sm)
-        dep_graph.create_dependencies_graph()
-        targets = dep_graph.find_recursive_deployments(['dep_0'])
-        self.assertEqual(len(targets), 5)
-        self.assertIn('dep_1', targets)
-        self.assertIn('dep_11', targets)
-        self.assertIn('dep_12', targets)
-        self.assertIn('dep_13', targets)
-        self.assertIn('dep_14', targets)
```
tehasdf, line 518: tests for `RecursiveDeploymentLabelsDependencies` ie. the graph, need to go: the graph is going to be unused very soon, and I don't want to keep them around.

tehasdf: also the recently-added `recalc_ancestors` tests, cover checking the "graph"/"tree" in more depth than those

mateumann (approved):

PR #3358, opened 2021-11-29
RD-3573 Rewrite label-dependency infra

tehasdf:
Look, this is going to be mostly impossible to review. I've tried to
split it into somewhat readable commits, but let's be realistic here...

Anyway, this rewrites my previous `recalc_ancestors` with an approach
that is way faster (see the `Rewrite recalc_ancestors` commit - if you
don't read anything else, at least read that one).

Then I make sure all the places that used to create the RecursiveGraph
stuff, use recalc_ancestors instead.

And then I drop a bunch of now-disused stuff. Nice.


tehasdf:
given another week, I'd split this into multiple smaller PRs. But alas.

tehasdf:
The `sorted()` (and `.order_by`) calls in this are sponsored by my realization that yes, a single query can still deadlock.
That is, `SELECT _storage_id from deps WHERE _storage_id IN (1, 2, 3)` and another one with a `IN (3, 2, 1)` MIGHT STILL DEADLOCK.

It's not about putting everything in a single query. Locks are taken on a per-row basis, not a per-query basis.

Instead of trying to put as much as you can into a query, you need to make sure that items are always accessed in the same order.

rest-service/manager_rest/resource_manager.py
```diff
@@ -2676,140 +2617,76 @@ def _get_utc_datetime_from_sched_plan(time_expression, base_datetime=None):
...
-                    deployment_status=new_status,
-                    sub_services_count=dep_summary.sub_services_count,
-                    sub_environments_count=dep_summary.sub_environments_count,
-                    sub_services_status=dep_summary.sub_services_status,
-                    sub_environments_status=dep_summary.sub_environments_status
+        with self.sm.transaction():
+            deps = models.DeploymentLabelsDependencies.get_dependencies(
+                deployment_ids, dependents=False, locking=True)
+            if not deps:
+                # no deps, means there's no tree to speak of, we just need to
```
geokala, line 2499: This should be "we only need to update deployment_ids" or "we just need to update deployment_ids", but I'm just commenting on it in case it wasn't a braino- I don't think it's necessarily worth fixing for this PR.

tehasdf: `we just need to update deployment_ids only` is the wording i changed this to, for extra emphasis

geokala (approved):

PR #3365, opened 2021-12-01
RD-3680 Reduce memory usage when fetching labels

tehasdf:
Fetching visibility/tenant_id like that didn't actually use the
declared relationship (the one with lazy=joined).
No, it used a separate join, and it actually emitted a cross join.
This is why the restservice starts using tens of gigabytes of ram at
some point.
To use the declared relationship, we have to use it explicitly,
like this.

### DIFF
--- a/rest-service/manager_rest/storage/resource_models.py
+++ b/rest-service/manager_rest/storage/resource_models.py
@@ -868,14 +868,6 @@ class LabelBase(CreatedAtMixin, SQLModelBase):
     def creator(cls):
         return one_to_many_relationship(cls, User, cls._creator_id, 'id')
 
-    @declared_attr
-    def visibility(cls):
-        return cls.labeled_model.visibility
-
-    @declared_attr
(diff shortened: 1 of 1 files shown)


rest-service/manager_rest/storage/resource_models.py
```diff
@@ -910,6 +905,9 @@ def blueprint(cls):
             'Blueprint', lazy='joined',
```
mateumann, line 905: Why did you leave quotes here?

tehasdf: removed now!

It doesn't really matter, anyway, sqlalchemy fetches them by string as well (it does it specifically so that it's possible to have two models reference each other, it'd be a NameError in one of them otherwise)
(but in this case, BlueprintLabel is defined after Blueprint, so it doesn't need to go by string)

mateumann (approved):

PR #3368, opened 2021-12-01
RD-3663 Limit iteration in dequeueing

tehasdf:
Rather than `while True`, make sure to limit it to a finite count.
This makes me feel safer.

Also, don't just fetch the same executions in every loop - repeating
the same and hoping for different results, is the definition of
insanity, after all.
Instead, increase offset in each iteration, to allow some crawling
through the results.

This doesn't directly fix RD-3663, but it's a stopgap for now


tehasdf:
Scratch the offset idea for now (see the last commit). So, this is just about replacing the `while True` with a finite loop

geokala (approved):

PR #3369, opened 2021-12-01
RD-3630 Allow to change blueprint's ownership

mateumann:
We're going to use blueprint update but explicitely check if user is
allowed to modify ownership.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/blueprints.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/blueprints.py
@@ -266,6 +266,10 @@ class BlueprintsId(resources_v2.BlueprintsId):
             blueprint.description = request_dict['description']
         if 'main_file_name' in request_dict:
             blueprint.main_file_name = request_dict['main_file_name']
+        if 'creator' in request_dict:
+            check_user_action_allowed('set_owner', None, True)
+            creator = rest_utils.valid_user(request_dict['creator'])
+            blueprint.creator = creator
         provided_labels = request_dict.get('labels')
 
         if request_dict.get('plan'):
--- a/rest-service/manager_rest/rest/resources_v3_1/blueprints.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/blueprints.py
@@ -267,8 +267,6 @@ class BlueprintsId(resources_v2.BlueprintsId):
         if 'main_file_name' in request_dict:
             blueprint.main_file_name = request_dict['main_file_name']
         if 'creator' in request_dict:
-            check_user_action_allowed('set_owner', None, True)
-            creator = rest_utils.valid_user(request_dict['creator'])
             blueprint.creator = creator
         provided_labels = request_dict.get('labels')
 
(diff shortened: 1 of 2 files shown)


tests/integration_tests/tests/agentless_tests/test_chown.py
```diff
@@ -0,0 +1,59 @@
...
+        blueprint = self.client.blueprints.get(self.BLUEPRINT_ID)
+        assert blueprint.get('created_by') == 'user1'
+
+        try:
+            self.client.users.delete('user1')
+        except CloudifyClientError as ex:
+            assert ex.status_code == 400
+            assert 'cannot delete user' in str(ex).lower()
+        else:
+            assert False
```
tehasdf, line 31: probably better written using `assertRaises`? the bare `assert False` will produce misleading exceptions

here's an example, also using `cm.exception` (which you can `str()`): https://github.com/cloudify-cosmo/cloudify-manager/blob/f2ecc3bef310d57284a1e707c19406017100f3a4/tests/integration_tests/tests/agentless_tests/multi_tenancy/test_users.py#L110-L112

tests/integration_tests/tests/agentless_tests/test_chown.py
```diff
@@ -0,0 +1,59 @@
...
+            username='user1', password='password1', tenant='test_tenant')
+        a_client.blueprints.upload(
+            resource('dsl/{}'.format(self.BLUEPRINT_FILENAME)),
+            entity_id=self.BLUEPRINT_ID)
+        try:
+            a_client.blueprints.update(self.BLUEPRINT_ID, {'creator': 'admin'})
+        except ForbiddenError as ex:
+            assert ex.status_code == 403
+            assert 'not permitted' in str(ex).lower()
+            assert 'set_owner' in str(ex).lower()
```
tehasdf, line 52: i'd say you don't need to assert for the "not permited", and for the 403, because that's already well implied by the ForbiddenError exception type; then you'll be able to just use `with self.assertRaisesRegexp(ForbiddenError, 'set_owner'):`

rest-service/manager_rest/rest/resources_v3_1/blueprints.py
```diff
@@ -266,6 +266,10 @@ def patch(self, blueprint_id, **kwargs):
             blueprint.description = request_dict['description']
         if 'main_file_name' in request_dict:
             blueprint.main_file_name = request_dict['main_file_name']
+        if 'creator' in request_dict:
+            check_user_action_allowed('set_owner', None, True)
+            creator = rest_utils.valid_user(request_dict['creator'])
+            blueprint.creator = creator
```
tehasdf, line 272: but we already figure out the user here: https://github.com/cloudify-cosmo/cloudify-manager/blob/f2ecc3bef310d57284a1e707c19406017100f3a4/rest-service/manager_rest/rest/resources_v3_1/blueprints.py#L227-L229

..but we didn't set it before? huh? so presumably now we just need to add the `blueprint.creator = creator` line and we don't need the others?

mateumann: oh, indeed

tehasdf (approved):

PR #3387, opened 2021-12-07
RD-3703 Dequeue create-dep-env before other workflows

tehasdf:
Make sure that create-dep-env is dequeued before executions of
any other workflows that use that deployment.

The "only dequeue executions that belong to deployments who
have no active executions" check is just not enough.
Instead, only dequeue executions, who:
- are either create-dep-env (so their deployment will NOT possibly 
  have any other executions running)
- belong to deployments that have:
  - no other executions _running_
  - no create-dep-env _queued_

That is to say, if a create-dep-env is even queued, then no other
executions for that deployment can possibly run.
Trying to run them would result in just an error, and indeed we
used to just throw an error.

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -65,8 +65,6 @@ class ResourceManager(object):
 
     def __init__(self, sm=None):
         self.sm = sm or get_storage_manager()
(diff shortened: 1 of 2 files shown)


tehasdf:
this fixes a bug, but the performance work will be continued in RD-3663

PR #3391, opened 2021-12-09
RD-3663 Dequeue more executions

tehasdf:
Make sure to check non-active exec-groups, when dequeuing executions.
Otherwise, already-at-capacity groups could be all that's returned from
our query, and we'd dequeue nothing.

This is the reason we used to always de-generate to running, like, 5
executions at a time, tops, even if there were thousands queued.

With this change, we'll dequeue up to... well, up to the new setting.
By default it's 20, which is still quite a lot, but if you know you'll be
running very simple workflows, you might as well crank it up.

### DIFF
--- a/rest-service/manager_rest/resource_manager.py
+++ b/rest-service/manager_rest/resource_manager.py
@@ -337,6 +337,43 @@ class ResourceManager(object):
         else:
             return self._cached_queued_execs_query
 
+    def _report_running(self):
+        """Report currently-running executions.
+
+        This returns the amount of currently-running executions total,
+        and a dict of {group_id: [active in the group, group concurrency]}
+        """
+        exgrs = models.executions_groups_executions_table
+        active_execs = (
+            db.session.query(
+                models.Execution._storage_id,
+                exgrs.c.execution_group_id,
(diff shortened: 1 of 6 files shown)


rest-service/manager_rest/resource_manager.py
```diff
@@ -417,27 +417,25 @@ def _get_queued_executions(self, deployment_storage_id):
...
                 # this execution cannot run, because it would exceed one
                 # of its' groups concurrency limit
                 continue
             if execution._deployment_fk in seen_deployments:
                 continue
 
-            for group in execution.execution_groups:
-                group_can_run[group] -= 1
+            for g in execution.execution_groups:
+                groups[g._storage_id][0] += 1
```
mateumann, line 440: are you sure of `+`?

tehasdf: yeah because before it used to be a "how many can the group still run" counter, but now it's a "how many is the group currently running" counter

and the check used to be something like `<= 0`, but now the check is `if any(groups[g._storage_id][0] >= groups[g._storage_id][1]`

mateumann (approved):

PR #3392, opened 2021-12-09
RD-3559 A basic string intrinsic functions test

mateumann:


mateumann:
String intrinsic functions are not complicated enough to justify adding more integration tests (which are relatively time-consuming)

PR #3409, opened 2021-12-15
RD-3805 Lock starting execution-groups

tehasdf:
Use The Big Lock™ here as well, to avoid deadlocks


geokala (dismissed):

geokala (dismissed):

geokala (changes requested):
This is wrong, can you fix it?

geokala (approved):

PR #3410, opened 2021-12-16
Speed up: rewrite test_nodes.py and test_agents.py

tehasdf:
Use the less-slow way for those tests.


rest-service/manager_rest/test/endpoints/test_nodes.py
```diff
@@ -31,146 +31,176 @@ class NodesTest(base_test.BaseServerTestCase):
...
-                nodes[i].deployment_id >= nodes[i + 1].deployment_id)
+        nodes = self.client.nodes.list(sort='deployment_id',
+                                       is_descending=True)
+        assert [n.id for n in nodes] == ['node2', 'node1']
 
     def test_bad_patch_node(self):
         """Malformed node instance update requests return an error."""
         response = self.patch('/node-instances/1234', 'not a dictionary')
-        self.assertEqual(400, response.status_code)
+        assert response.status_code == 400
```
tehasdf, line 117: this also rewrites all the `self.assertX` as just `assert`. Not a very important reason for that, but why not.

Tbh, I was mostly using this as a way to see which tests I have already gone through.

rest-service/manager_rest/test/endpoints/test_nodes.py
```diff
@@ -31,146 +31,176 @@ class NodesTest(base_test.BaseServerTestCase):
...
                                   'version': 4
                               })
-        self.assertEqual(200, response.status_code)
-        self.assertEqual('ddd', response.json['runtime_properties']['ccc'])
-        self.assertEqual('b-state', response.json['state'])
+        assert response.status_code == 200
+        assert response.json['runtime_properties']['ccc'] == 'ddd'
+        assert response.json['state'] == 'b-state'
 
-    @skip('Deprecated since using sqlalchemy locking mechanism')
```
tehasdf, line 129: we use the sqlalchemy built-in way, but there still is a way to test it!

rest-service/manager_rest/test/endpoints/test_nodes.py
```diff
@@ -205,178 +231,157 @@ def test_patch_node_runtime_props_overwrite(self):
...
             runtime_properties={}
         )
-        self.assertEqual('1234', response['id'])
-        self.assertEqual(0, len(response['runtime_properties']))
+        assert response['id'] == '1234'
+        assert len(response['runtime_properties']) == 0
 
     def test_patch_node_conflict(self):
         """A conflict inside the storage manager propagates to the client."""
-        # patch the storage manager .update_node_instance method to throw an
```
tehasdf, line 244: ha-ha! use mock.patch instead!

rest-service/manager_rest/test/endpoints/test_nodes.py
```diff
@@ -205,178 +231,157 @@ def test_patch_node_runtime_props_overwrite(self):
...
+        assert all(ni.node_id == node2.id for ni in dep1_n2_instances)
 
-        def assert_dep_and_node(expected_len, dep, node_id, instances):
-            self.assertEqual(expected_len, len(instances))
-            for instance in instances:
-                self.assertEqual(instance.deployment_id, dep)
-                self.assertEqual(instance.node_id, node_id)
+        assert len(dep2_n3_instances) == 2
+        assert all(ni.deployment_id == dep2.id for ni in dep2_n3_instances)
+        assert all(ni.node_id == node3.id for ni in dep2_n3_instances)
```
tehasdf, line 361: those all written out seem silly, but I prefer this as it is simpler than a nested function

rest-service/manager_rest/test/endpoints/test_nodes.py
```diff
@@ -385,39 +390,46 @@ def test_patch_before_put(self):
...
-        for instance in instances:
-            # test list/get/patch endpoints
-            tested_instances = [
-                instance,
-                self.client.node_instances.get(instance.id),
-                self.client.node_instances.update(instance.id)]
-            for tested_instance in tested_instances:
-                scaling_groups = tested_instance.scaling_groups
-                self.assertEqual(1, len(scaling_groups))
-                self.assertDictContainsSubset({'name': 'group1'},
```
tehasdf, line 414: checking that the _instance_ has the groups, is testing instance creation in create-dep-env. I don't want to test that here.
Instead, this test is about figuring out the actual_planned_number_of_instances per node, which is something that we have specific logic for in the get nodes endpoint, so it'd be good for it to be tested

geokala (approved):

PR #3412, opened 2021-12-16
Speed up: rewrite several infrastructure/ tests

tehasdf:
This speeds them up MASSIVELY, the whole directory used to be on the
order of 4 minutes, but it's on the order of 1 minute now.

The idea is, instead of creating actual real plugins etc, just to
test listing and filtering, only create the models directly.

### DIFF
--- a/rest-service/manager_rest/test/infrastructure/test_list_filters.py
+++ b/rest-service/manager_rest/test/infrastructure/test_list_filters.py
@@ -13,6 +13,8 @@
 #  * See the License for the specific language governing permissions and
 #  * limitations under the License.
 
+from datetime import datetime
+
 from cloudify_rest_client.exceptions import CloudifyClientError
 from manager_rest import manager_exceptions
 from manager_rest.storage import models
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/test/infrastructure/test_list_filters.py
```diff
@@ -276,7 +275,6 @@ def test_blueprints_list_with_filters(self):
                                       'values {0}, got {1}'
                                       .format(filter_params, blueprint))
         self.assertEqual(self.first_blueprint_id, blueprint['id'])
-        self.assertIsNotNone(response[0]['plan'])
```
geokala, line 279: Why did you remove these checks and/or why were we asserting them before? They look unrelated to anything this test needs to test?

tehasdf: oh yeah, forgot to mention. Indeed I'm not going to check that plan specifically is set. We check that the right blueprint are returned by the filters, I dont see much reason to check for plan specifically. And in this case it is None becaue I don't mock it. I could but why.

geokala (approved):

PR #3448, opened 2022-01-04
Rd 3728 clone service component must be able to use lables

EarthmanT:


cloudify_types/cloudify_types/component/operations.py
```diff
@@ -81,6 +81,7 @@ def upload_blueprint(**kwargs):
     blueprint_id = blueprint.get('id') or ctx.instance.id
     blueprint_archive = blueprint.get('blueprint_archive')
     blueprint_file_name = blueprint.get('main_file_name')
+    blueprint_labels = blueprint.get('labels', [])
```
mateumann, line 84: `blueprint_labels` is something different than `labels` (or deployment labels as we sometimes call them) https://docs.cloudify.co/latest/developer/blueprints/spec-blueprint-labels/ and https://docs.cloudify.co/latest/developer/blueprints/spec-labels/.  Just make sure it's really `labels` you're after.

cloudify_types/cloudify_types/component/operations.py
```diff
@@ -282,6 +301,17 @@ def create(timeout=EXECUTIONS_TIMEOUT, interval=POLLING_INTERVAL, **kwargs):
...
+    if not validate_labels(deployment_labels):
+        raise NonRecoverableError(
+            "The provided deployment labels are not valid. "
+            "Labels must be a list of single-entry dicts, "
+            "e.g. [{\'foo\': \'bar\'}]. "
+            "This value was provided: %s." % deployment_labels
+        )
+
+    if not any('csys-obj-parent' in item for item in deployment_labels):
+        deployment_labels.append({'csys-obj-parent': ctx.deployment.id})
```
mateumann, line 314: Is it safe to assume that all components should get `csys-obj-parent` label set by default to `ctx.deployment.id` (unless specified otherwise)?

cloudify_types/cloudify_types/component/tests/test_blueprint_handling.py
```diff
@@ -128,3 +128,46 @@ def test_upload_blueprint_use_not_existing_external(self):
                 'Blueprint ID "test" does not exist'
             ):
                 upload_blueprint(**self.resource_config)
+
+    def test_upload_blueprint_bad_labels(self):
+        with mock.patch('cloudify.manager.get_rest_client') as mock_client:
+            blueprint_labels = {'foo': 'bar'}
```
mateumann, line 134: It's not exactly `blueprint_labels` (see the comment above).  It's just naming thing though, but might be confusing for someone looking at your code some time later.

resources/rest-service/cloudify/types/types.yaml
```diff
@@ -1065,6 +1065,11 @@ data_types:
         type: string
         default: ""
         required: true
+      labels:
+        description: >
+          A list of single-entry dicts describing labels to apply to the blueprint.
```
mateumann, line 1070: `labels` are applied to the deployments.

`blueprint_labels` apply to the blueprints themselves.

PR #3450, opened 2022-01-05
RD-3748 Use tenant_name when creating IDDs

mateumann:
... because of times when multiple tenants use the same deployment_id
for different deployment and we should be able to find out which one
should be used.


rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -598,6 +598,19 @@ def _verify_and_get_source_and_target_deployments(
             )
         return source_deployment, target_deployment
 
+    @staticmethod
+    def _get_deployment(sm, deployment_id, tenant_name):
+        if tenant_name:
+            return sm.get(models.Deployment,
+                          None,
+                          filters={'id': deployment_id,
```
geokala, line 606: Maybe make a filters variable which always has 'id', and then add 'tenant_name' to it if tenant_name, then you can just have one sm.get call?
That's pretty much a style thing though, not a big deal either way.

tehasdf (changes requested):
I don't really understand. `sm.get` already adds a tenant filter, by `current_tenant`. So surely applying another tenant filter as well, will just return none, because no deployment can ever belong to two tenants?

If this does indeed have some effect, then this for sure needs tests to show how does this work

geokala:
@tehasdf 1. There's other things about this that we were discussing on the channel, so I think this is probably not the form it wants to take anyway.

2. You're really bad at being on holiday.

tehasdf (approved):

PR #3452, opened 2022-01-09
RD-3929 Allow uninstall of inactive source get-capa

glukhman:


rest-service/manager_rest/resource_manager.py
```diff
@@ -2436,6 +2436,12 @@ def _verify_dependencies_not_affected(self, execution, force):
         deployment = execution.deployment
         idds = self._get_blocking_dependencies(
             deployment, skip_component_children=True)
+        # allow uninstall of get-capability dependencies
+        # if the dependent deployment is inactive
+        if execution.workflow_id == 'uninstall':
+            idds = [idd for idd in idds if not (
+                    'get_capability' in idd.dependency_creator and
```
tehasdf, line 2443: for sure needs to be `.get_capability.` or something like that, to avoid catching cases where the node instance id is eg. "xxxget_capabilityxxx"

tehasdf (approved):

PR #3455, opened 2022-01-10
RD-3887 Synchronous dep-update preview

tehasdf:
It used to be synchronous before, let's keep it this way for now.
This also allows me to revert the test to the previous state, in which
it didn't wait for previews.


tehasdf:
this is pretty trivial, because we use the same way of waiting-for-execution in blueprint-upload and in create-dep-env

glukhman (approved):

PR #3459, opened 2022-01-11
RD-3952 dep-update: bump deployment.updated_at

tehasdf:
Apparently this doesn't happen automatically!

### DIFF
--- a/rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
+++ b/rest-service/manager_rest/rest/resources_v2_1/deployment_update.py
@@ -14,6 +14,7 @@
 #  * limitations under the License.
 
 import uuid
+from datetime import datetime
 
 from flask import request
 from flask_restful_swagger import swagger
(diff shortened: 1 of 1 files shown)


tehasdf:
I'm not gonna push the inte-test I used to test this, because adding ~20 seconds to the test run, to test something as insignificant as this, is not a good trade in my book.

Instead, I'll make sure to test this using lower-level tests, after RD-3859

Anyway, the test I did use to try this out, is:
```python

    def test_updated_at_timestamp(self):
        # any blueprint works here, let's just pick a small one
        deployment, modified_bp_path = \
            self._deploy_and_get_modified_bp_path('add_output')
        self.client.blueprints.upload(modified_bp_path, BLUEPRINT_ID)
        wait_for_blueprint_upload(BLUEPRINT_ID, self.client)

        deployment = self.client.deployments.get(deployment.id)
        assert deployment['updated_at'] == deployment['created_at']

        self._do_update(deployment.id, BLUEPRINT_ID)

        deployment = self.client.deployments.get(deployment.id)
        assert deployment['updated_at'] > deployment['created_at']

```

geokala (approved):

PR #3489, opened 2022-01-26
RD-3907 New dep-update: implement updating operation inputs

tehasdf:
This is like #1919 (CY-1710), but for the new dep-update.
For the most part, it copies the code, just to this place.

Instead of the whole dep-update happening on the REST side, it now
happens in the workflow; but some parts must still happen in
restservice, of course.

Anyway, this is a bit heavy-handed, but I'll just move the whole update
thing to a pretty-much-RPC call in POST /operations.

This call is going to do the same thing that the old dep-up did:
examine stored operations, and update them with the new inputs.


tehasdf:
requires cloudify-cosmo/cloudify-common#3489

tehasdf:
the old-update impl is https://github.com/cloudify-cosmo/cloudify-manager/blob/9769287be1e0501b375e9db8f8cf1ef559eea26f/rest-service/manager_rest/deployment_update/handlers.py#L214

glukhman (approved):

PR #3497, opened 2022-01-28
RD-4096 Name the new dep-up workflow "update"

tehasdf:
The `csys-new-deployment-update` idea is wrong, because
people do rely on the update workflow's name being actually
"update".
So, let's rename it to `update`, then. This also, hopefully, keeps
back-compat with old blueprints.


tehasdf:
This also requires cloudify-cosmo/cloudify-common#977 for compat w/ old blueprints

geokala (approved):

PR #3499, opened 2022-01-30
Rd 3892 consumer intrinsic func

glukhman:


rest-service/manager_rest/dsl_functions.py
```diff
@@ -211,7 +211,7 @@ def get_group_capability(self, capability_path):
             return [caps for dep_id, caps in
                     sorted(capabilities.items(), key=lambda item: item[0])]
 
-    def get_sys(self, entity, prop):
+    def get_sys(self, entity, prop, subprop=None):
         deployment = self.sm.get(Deployment, self._deployment_id)
```
geokala, line 215: We should probably complain if someone tries to pass subprop on something that doesn't support subprop.

geokala (dismissed):

tehasdf (dismissed):

geokala (approved):

PR #3502, opened 2022-02-01
Fix bp upload inte test 404

glukhman:


tehasdf:
this test only started being correct at all after cloudify-cosmo/cloudify-manager-install#1220

tehasdf (approved):

tehasdf (approved):

PR #3509, opened 2022-02-04
RD-2617 Implement deleting node-instances

tehasdf:
This is going to be only useful from dep-update. Don't you dare
use it otherwise.

I mean, you can, if you want to break your shit. I'll still try to stop
you, via the `@only_deployment_update`.


tehasdf:
Requires cloudify-cosmo/cloudify-common#982 and cloudify-cosmo/cloudify-manager-install#1306

geokala (approved):

PR #3513, opened 2022-02-08
RD-4131 Intrinsic functions in resource_tags test

mateumann:


tests/integration_tests/tests/agentless_tests/test_plugin_labels.py
```diff
@@ -123,12 +126,32 @@ def test_resource_tags_propagation(self):
...
+            entity_id='bp')
+        self.client.deployments.create('bp', 'dep')
+
+        execution = self.client.executions.start(
+            deployment_id='dep',
+            workflow_id='execute_operation',
+            parameters={'operation': 'test.context'}
+        )
+
+        self.wait_for_execution_status(execution.id, 'completed')
```
tehasdf, line 150: this needs to be "terminated", not "completed".  "Completed" is just the display text, but the actual value is "terminated".

Or really, just:
```suggestion
        self.wait_for_execution_to_end(execution)
```
(this already throws if the execution fails)

tests/integration_tests/resources/dsl/scripts/workflows/context.py
```diff
@@ -0,0 +1,6 @@
+from cloudify.workflows import ctx
```
tehasdf, line 1:
```suggestion
from cloudify import ctx
```
we can't use the workflow context here, this needs to be operation ("regular") context. This is in an operation, not in a workflow
`from cloudify.state import ctx` also works.

This should really error out in a nicer way, I've made RD-4209 about it

tehasdf (approved):

PR #3530, opened 2022-02-15
RD-4117 `deployment_id` data type validation

mateumann:
Implements methods which allow to list deployments of all tenants,
filtered by deployment.id.

Pass these methods to dsl_parser's `prepare_deployment_plan` function.


workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -158,6 +168,11 @@ def create(ctx, labels=None, inputs=None, skip_plugins_validation=False,
     idd.create(ctx, client, deployment_plan)
 
 
+def _get_deployments(client: 'CloudifyClusterClient',
```
tehasdf, line 171: doesn't need to be cluster, just `CloudifyClient` is fine (which is a superclass of cluster)

workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -158,6 +168,11 @@ def create(ctx, labels=None, inputs=None, skip_plugins_validation=False,
...
+                     deployment_id: str) -> 'ListResponse':
```
tehasdf, line 172: and ListResponse is probably not great either. I think we can do something like Iterable[Deployment]

PR #3537, opened 2022-02-18
Allow retrieving password hash with filter

geokala:
Without this, applying an _include to the query makes _include_hash unusable.


tehasdf (approved):

geokala:
For context: When building the new snapshots I want to retrieve only the necessary data (e.g. not calculated values or the like) in order to reduce snapshot size and time.

PR #3538, opened 2022-02-21
RD-4306 Allow migrations to be rendered to SQL

tehasdf:
We used to be able to render migrations to just SQL statements,
but recently we've broken it. This brings it back.

It's not used just yet, but eventually I'd like to render the SQL
at build time, and apply it when installing the manager (instead
of applying python-level migrations).

This will be important for being able to split services into separate
containers, because now the DB will be able to create itself, as
opposed to the restservice creating it.

### DIFF
--- a/resources/rest-service/cloudify/migrations/versions/1fbd6bf39e84_4_5_to_4_5_5.py
+++ b/resources/rest-service/cloudify/migrations/versions/1fbd6bf39e84_4_5_to_4_5_5.py
@@ -50,7 +50,6 @@ def upgrade():
     op.add_column('executions',
                   sa.Column('scheduled_for', UTCDateTime(), nullable=True))
 
-    op.execute('COMMIT')
     # Add new execution status
     op.execute("alter type execution_status add value 'scheduled'")
     op.add_column(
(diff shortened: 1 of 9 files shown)


tehasdf:
There are style-only commits here (the whitespace one), so I do recommend only reading this commit-by-commit

resources/rest-service/cloudify/migrations/versions/1fbd6bf39e84_4_5_to_4_5_5.py
```diff
@@ -50,7 +50,6 @@ def upgrade():
     op.add_column('executions',
                   sa.Column('scheduled_for', UTCDateTime(), nullable=True))
 
-    op.execute('COMMIT')
```
geokala, line 53: Was this commit never required then?

tehasdf: Can't see why would it be, other than to break the "single transaction" semantics and show a warning

mateumann (approved):

resources/rest-service/cloudify/migrations/versions/1fbd6bf39e84_4_5_to_4_5_5.py
geokala: Fair enough- no additional comments, and I see @mateumann greened it, so merge away.

PR #3543, opened 2022-02-22
RD-2358 Pass existing NI ids to node graph

glukhman:


workflows/cloudify_system_workflows/deployment_environment.py
```diff
@@ -104,12 +104,21 @@ def create(ctx, labels=None, inputs=None, skip_plugins_validation=False,
            display_name=None, **_):
     client = get_rest_client(tenant=ctx.tenant_name)
     bp = client.blueprints.get(ctx.blueprint.id)
+
+    plan_node_ids = [n['id'] for n in bp.plan['nodes']]
+    existing_ni_ids = {}
+    for ni in client.node_instances.list(node_id=plan_node_ids):
```
tehasdf, line 110: should be able to just `deployment_id=`

glukhman: umm, no. We want NI's from _any_ deployment with identical names

tehasdf: aha! then this needs to traverse pagination

tehasdf (approved):

PR #3554, opened 2022-02-28
deployment create-with-workdir: use makedirs

tehasdf:
In case the tenant dir is missing (removed by.. something? maybe
removed by the reset-storage!), let's just create it. This also
mirrors the way we do it in create-dep-env, where we do use makedirs.

I'm not 100% happy to change the behaviour, because maybe
reset-storage could've just been more intelligent instead, but...
well, I think it's a good change anyway.


tehasdf:
FWIW create-dep-env: https://github.com/cloudify-cosmo/cloudify-manager/blob/1b66be5d4a2a7c50b58d841b81c12d6ceb636be2/workflows/cloudify_system_workflows/deployment_environment.py#L229

geokala (approved):

PR #3555, opened 2022-02-28
RD-4328 Allow filtering graphs by just execution-id

tehasdf:
No need to require the name. We can allow to just list all graphs
for the given execution.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/operations.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/operations.py
@@ -348,16 +348,20 @@ class TasksGraphs(SecuredResource):
     def get(self, _include=None, pagination=None, **kwargs):
         args = get_args_and_verify_arguments([
             Argument('execution_id', type=text_type, required=True),
-            Argument('name', type=text_type, required=True)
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/resources_v3_1/operations.py
```diff
@@ -348,16 +348,20 @@ class TasksGraphs(SecuredResource):
...
+        filters = {'execution': execution}
+        if name:
+            filters['name'] = name
         return sm.list(
             models.TasksGraph,
-            filters={'execution': execution, 'name': name},
-            pagination=pagination
+            filters=filters,
+            pagination=pagination,
+            include=_include,
```
tehasdf, line 364: also adding `_include` here. It's unrelated but looks like we forgot it before. Also, tasks graphs don't have anything interesting in them, so no reason to use it. But let's have it.

tehasdf:
requires the client from cloudify-cosmo/cloudify-common#1002

geokala (dismissed):

geokala (approved):

PR #3561, opened 2022-03-03
RD-3903 Keep custom IDDs after dep-update

tehasdf:
Just look at the idd's dependency_creator, and if it's not coming
from a function, it means it's created by the user explicitly.
And in that case, we keep it.

I do realize that this isn't bulletproof, ie. someone might've created
an idd with "get_capability" in the name, but at that point we're so
deep into direct inner workings of the manager, that I don't even.

Like, the user knows what they're doing at that point, because tbh
that user is likely me anyway. So, suck it, me: better stay in line
and name your dependency creators in a smart way!

### DIFF
--- a/workflows/cloudify_system_workflows/idd.py
+++ b/workflows/cloudify_system_workflows/idd.py
@@ -5,6 +5,7 @@ from cloudify.workflows.workflow_context import CloudifyWorkflowContext
 from cloudify_rest_client.client import CloudifyClient
 from dsl_parser import constants as dsl_constants
 from dsl_parser import models as dsl_models
+from dsl_parser import functions as dsl_functions
 
 
 def create(ctx: CloudifyWorkflowContext,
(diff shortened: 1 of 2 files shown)


workflows/cloudify_system_workflows/idd.py
```diff
@@ -28,6 +29,33 @@ def create(ctx: CloudifyWorkflowContext,
...
+    Custom IDDs, created by the user via the rest-client, ie. not coming from
+    an intrinsic functions, should be kept as well.
+    """
+    creator = idd['dependency_creator']
+    if creator.startswith(
+        ('component.', 'sharedresource.')
+    ):
+        return False
+
+    _, _, last_segment = creator.rpartition('.')
```
glukhman, line 47: maybe just
`last_segment = creator.split[-1]` ? can we be sure there are always 3 parts here?

tehasdf: the point of partition is that there's always 3 parts yea (possibly the empty string)

I started with split, wrote it wrong (the right way is `creator.rsplit('.', 1)[-1]`) and said "fuck that I'll just always use partition ever, that one is impossible to get wrong" :/

PR #3562, opened 2022-03-03
RD-4412 Add new tokens model

geokala:


rest-service/manager_rest/storage/management_models.py
```diff
@@ -474,6 +481,19 @@ def is_locked(self):
         return False
 
 
+class Token(SQLModelBase, CreatedAtMixin):
```
tehasdf, line 484: expiration date? execution id?

rest-service/manager_rest/storage/management_models.py
```diff
@@ -474,6 +481,19 @@ def is_locked(self):
...
+    __tablename__ = 'tokens'
+
+    id = db.Column(db.String(10), primary_key=True, index=True)
+    secret = db.Column(db.String(255), nullable=False)
```
tehasdf, line 488: string? maybe binary?

resources/rest-service/cloudify/migrations/versions/272e61bf5f4a_6_3_to_6_4.py
```diff
@@ -0,0 +1,45 @@
...
+    op.create_index(op.f('tokens_created_at_idx'),
+                    'tokens', ['created_at'], unique=False)
+    op.create_index(op.f('tokens_id_idx'), 'tokens', ['id'], unique=True)
+
+
+def downgrade():
+    op.drop_index(op.f('tokens_last_used_idx'), table_name='tokens')
+    op.drop_index(op.f('tokens_id_idx'), table_name='tokens')
+    op.drop_index(op.f('tokens_created_at_idx'), table_name='tokens')
+    op.drop_table('tokens')
```
tehasdf, line 45: might as well just drop the table only, i mean, indexes will be gone too

rest-service/manager_rest/storage/management_models.py
tehasdf: if it's a string, i assume it's base64?

geokala: It's the same as the password field on the users table.

resources/rest-service/cloudify/migrations/versions/272e61bf5f4a_6_3_to_6_4.py
geokala: Ah- I assumed alembic had good reason for doing both. I'll let it implicitly drop indices then.

rest-service/manager_rest/storage/management_models.py
tehasdf: oh, right, it's a hash. Maybe `secret_hash`?

rest-service/manager_rest/storage/management_models.py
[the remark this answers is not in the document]
geokala: OK, will resolve that with the same commit introducing the other columns.

geokala: Yup, adding.

resources/rest-service/cloudify/migrations/versions/272e61bf5f4a_6_3_to_6_4.py
tehasdf: Nah it's just always emitting drops for corresponding creates. It doesn't have an optimizer

rest-service/manager_rest/storage/management_models.py
geokala: Sure, why not

tehasdf (approved):

PR #3563, opened 2022-03-03
Rd 4420 ext component

glukhman:


cloudify_types/cloudify_types/component/operations.py
```diff
@@ -66,16 +66,52 @@ def _get_desired_operation_input(key, args):
 def _get_client(kwargs):
     client_config = _get_desired_operation_input('client', kwargs)
     if client_config:
-        return CloudifyClient(**client_config)
+        client = CloudifyClient(**client_config)
+
+        # for determining an external client:
+        manager_ips = [mgr.private_ip for mgr in
```
tehasdf, line 72: I think let's actually do both private_ip + public_ip into the set. It's our local manager in either case!

cloudify_types/cloudify_types/component/operations.py
```diff
@@ -66,16 +66,52 @@ def _get_desired_operation_input(key, args):
...
+                       manager.get_rest_client().manager.get_managers()]
+        internal_hosts = ({'127.0.0.1', 'localhost'} | set(manager_ips))
+        host = {client.host} if type(client.host) == str \
```
tehasdf, line 75:
```suggestion
        host = {client.host} if isinstance(client.host, str) \
```
as a general rule, always prefer isinstance over `type()==`, because `type()==` ignores subclasses (breaking the liskov substitution principle). Doesn't matter in this case, and I'm aware this is just copied, but let's fix it while we're at it.

cloudify_types/cloudify_types/component/operations.py
```diff
@@ -66,16 +66,52 @@ def _get_desired_operation_input(key, args):
...
+            else set(client.host)
+        is_external_host = not (host & internal_hosts)
```
tehasdf, line 77: you know what would be funny, if `host` was several entries and only _some_ matched our own manager, and others didn't. What would that mean? Who knows. Maybe we want to at least log a warning.

cloudify_types/cloudify_types/component/operations.py
```diff
@@ -66,16 +66,52 @@ def _get_desired_operation_input(key, args):
...
     else:
-        return manager.get_rest_client()
+        client = manager.get_rest_client()
+        is_external_host = False
+    return client, is_external_host
+
+
+def _get_idd(deployment_id, is_external_host, kwargs):
```
tehasdf, line 84: imo let's factor this out to some utils module & from shared_resource as well

cloudify_types/cloudify_types/component/operations.py
```diff
@@ -66,16 +66,52 @@ def _get_desired_operation_input(key, args):
...
+        local_dependency_params['external_target'] = {
+            'deployment': deployment_id,
+            'client_config': client_config
+        }
+        inter_deployment_dependency['external_source'] = {
+            'deployment': ctx.deployment.id,
+            'tenant': ctx.tenant_name,
+            'host': manager_ips
+        }
+    return inter_deployment_dependency, local_dependency_params
```
tehasdf, line 107:
```suggestion
    inter_deployment_dependency = create_deployment_dependency(
        dependency_creator_generator(COMPONENT, ctx.instance.id),
        source_deployment=ctx.deployment.id,
        target_deployment=deployment_id
    )
    local_dependency = None
    if is_external_host:
        client_config = _get_desired_operation_input('client', kwargs)
        manager_ips = [mgr.public_ip for mgr in
                       manager.get_rest_client().manager.get_managers()]
        local_dependency = create_deployment_dependency(
            dependency_creator_generator(COMPONENT, ctx.instance.id),
            source_deployment=ctx.deployment.id,
            target_deployment=' ',  # ??? can't it be just none?
            external_target={
                'deployment': deployment_id,
                'client_config': client_config
            },
            external_source={
                'deployment': ctx.deployment.id,
                'tenant': ctx.tenant_name,
                'host': manager_ips
            }
        )
    return inter_deployment_dependency, local_dependency
```

I suggest like this, so that both sides (inter_depl_dep, and local_dependency) are just a result of calling `create_deployment_dependency`. That function already takes `external_target` & `external_source` as arguments.

Then, this function, doesn't need to make the assumption that the dependency is a dict. (or on the key names)

tehasdf (dismissed):

tehasdf (approved):

PR #3568, opened 2022-03-07
Rd 4412 create and delete tokens

geokala:


rest-service/manager_rest/rest/endpoint_mapper.py
```diff
@@ -108,7 +109,6 @@ def setup_resources(api):
         'SummarizeExecutions': 'summary/executions',
         'SummarizeBlueprints': 'summary/blueprints',
         'SummarizeExecutionSchedules': 'summary/execution_schedules',
-        'UserTokens': 'user-tokens/<string:user_id>',
```
tehasdf, line 111: so of course this is not back-compatible, as you are aware; it'll also need rest-client updates

rest-service/manager_rest/security/authentication.py
```diff
@@ -181,7 +181,10 @@ def _authenticate_token(self, token):
         elif not user:
             raise NoAuthProvided()
         else:
-            self._verify_token(token=data[1], user=user)
+            # If we get here, we already validated the new token. Only legacy
+            # tokens need an extra check.
+            if not token.startswith('ctok-'):
+                self._verify_token(token=data[1], user=user)
```
tehasdf, line 187: Now this makes me a bit uneasy, because _right now_ we indeed have already validated the new token, but I can imagine the possibility of the code above being refactored in such a way, that we get here, and end up not verifying the token in one of the cases.

In other words: I don't like how this tree of elifs is non-complete.

And, we already have an `else` (non-ctok) branch inside of `get_token_status`! Let's just move the `_verify_token` call to there, no?

It's going to be easy to do, except for the `token_verified_cache`. But the new tokens don't use it. So you can either 1) just get rid of the cache (and don't we all love -lines),  2) add a cache to the new tokens as well

rest-service/manager_rest/rest/resources_v3_1/tokens.py
```diff
@@ -1,35 +1,110 @@
...
+def _can_manage_token(token):
+    if token._user_fk == current_user.id:
+        return True
+    elif is_user_action_allowed('manage_others_tokens'):
+        return True
+    return False
+
+
+# A bit more user friendly than uuids
+def _random_string(length=10):
```
tehasdf, line 108:
```suggestion
def _random_string(length=10):
    """A random string that is a bit more user friendly than uuids"""
```

rest-service/manager_rest/rest/resources_v3_1/tokens.py
```diff
@@ -1,35 +1,110 @@
...
+            sm.delete(token)
+            return None, 204
+        else:
+            raise NotFoundError()
+
+    @marshal_with(responses.Tokens)
+    @authorize('list_token')
+    def get(self, token_id):
+        sm = get_storage_manager()
+        token = sm.get(models.Token, token_id, fail_silently=True)
```
tehasdf, line 85: the alternative is `filters = {'id': token_id};   if can_manage_others: filters['_user_fk'] = current_user.id` and then no `silently` in the get call. I'm not sure if it's much better, other than it handles the 404 case for you already

rest-service/manager_rest/rest/resources_v3_1/tokens.py
```diff
@@ -1,35 +1,110 @@
...
+    @authorize('delete_token')
+    def delete(self, token_id):
+        """Delete an existing token."""
+        sm = get_storage_manager()
+        token = sm.get(models.Token, token_id, fail_silently=True)
+        if token and _can_manage_token(token):
+            sm.delete(token)
+            return None, 204
+        else:
+            raise NotFoundError()
```
tehasdf, line 79: at least include the id in the 404 messages please

rest-service/manager_rest/rest/resources_v3_1/tokens.py
```diff
@@ -1,35 +1,110 @@
...
+            id=_random_string(),
+            description=request_dict.get('description', ''),
+            secret_hash=hash_password(secret),
+            expiration_date=expiration_date,
+            _user_fk=current_user.id,
+        )
+        sm.put(token)
+
+        # Keeping the same structure as old tokens to make migration easier,
+        # but adding relevant extra details.
```
tehasdf, line 60: no need to comment this, IMO comments should be about code as-is, not about the changes and process

rest-service/manager_rest/rest/resources_v3_1/tokens.py
```diff
@@ -1,35 +1,110 @@
...
+
+        expiration_date = request_dict.get('expiration_date')
+        if expiration_date:
+            expiration_date = parse_utc_datetime(expiration_date)
+            if is_expired(expiration_date):
+                raise BadParametersError("Expiration date was in the past.")
+
+        token = models.Token(
+            id=_random_string(),
+            description=request_dict.get('description', ''),
```
tehasdf, line 52: currently `description` isn't optional (in the verify_params) call, so 2-arg get isn't necessary.
There's several ways we can go about this, but I think if description is nullable (and it is), then nulls should be allowed?

rest-service/manager_rest/rest/resources_v3_1/tokens.py
```diff
@@ -1,35 +1,110 @@
...
+        return True
+    elif is_user_action_allowed('manage_others_tokens'):
+        return True
+    return False
+
+
+# A bit more user friendly than uuids
+def _random_string(length=10):
+    charset = string.ascii_uppercase + string.ascii_lowercase + string.digits
+    return ''.join(random.choice(charset) for i in range(length))
```
tehasdf, line 110: this is also used for the token secrets, so I think just out of principle, I'd use SystemRandom here.

And actually, we're on 3.6, so:
```suggestion
    return ''.join(secrets.choice(charset) for i in range(length))
```
(and import secrets)

rest-service/manager_rest/rest/resources_v3_1/tokens.py
```diff
@@ -1,35 +1,110 @@
...
+        else:
+            raise NotFoundError()
+
+
+def _can_manage_token(token):
+    if token._user_fk == current_user.id:
+        return True
+    elif is_user_action_allowed('manage_others_tokens'):
+        return True
+    return False
```
tehasdf, line 104: I think `return token.user == curr.id or is_allowed('')` is a cleaner way to write this

rest-service/manager_rest/security/user_handler.py
```diff
@@ -102,22 +90,45 @@ def get_token_status(token):
...
+
+    if token.startswith('ctok-'):
+        token_parts = token.split('-')
+        if len(token_parts) == 3:
+            _, tok_id, tok_secret = token_parts
+
+            sm = get_storage_manager()
+            token = sm.get(Token, tok_id, fail_silently=True)
+
+            error = 'Unauthorized.'
```
tehasdf, line 104: The full stop in error messages feels a bit weird, don't it?

rest-service/manager_rest/security/user_handler.py
```diff
@@ -102,22 +90,45 @@ def get_token_status(token):
...
+            if token:
+                if verify_password(tok_secret, token.secret_hash):
+                    user = user_datastore.find_user(id=token._user_fk)
+                    # Legacy data structure
+                    data = [str(token._user_fk), 'no_hash_needed']
```
tehasdf, line 109: this would be unnecessary, if we did the thing I suggested in one of the above comments, and pulled old-token `verfiy_token` into here. Then the result of this method wouldn't need `data` at all.

Actually, then we realize, instead of returning the `expired, invalid, error` from this, just for the caller to examine them and throw, this function could just throw. Then, `Authentication._authenticate_token` could be just removed.

Now my mouth is watering at the refactor possible here. Do ittttttt.

rest-service/manager_rest/utils.py
```diff
@@ -454,3 +439,10 @@ def _get_timestamp(value):
         return date_parser.parse(value, ignoretz=True)
     else:
         return value
+
+
+def is_expired(expiry):
+    """Check whether an expiry time has passed."""
+    if isinstance(expiry, str):
```
tehasdf, line 446: does this _really_ need a case for strings?

rest-service/manager_rest/rest/endpoint_mapper.py
geokala: Yup, it's on the radar.

rest-service/manager_rest/security/authentication.py
geokala: My intent was to just delete this once we delete old tokens, but we need the stage part to be done before that.

rest-service/manager_rest/rest/resources_v3_1/tokens.py
geokala: I think I'll leave it at the current way for the moment. Will keep thinking about it though.

rest-service/manager_rest/rest/resources_v3_1/tokens.py
geokala: Updated.

rest-service/manager_rest/rest/resources_v3_1/tokens.py
geokala: Updated.

rest-service/manager_rest/rest/resources_v3_1/tokens.py
geokala: I actually forgot I made that nullable. I'll update this to allow null descriptions.

rest-service/manager_rest/rest/resources_v3_1/tokens.py
geokala: Updating.

rest-service/manager_rest/rest/resources_v3_1/tokens.py
geokala: I think I briefly considered refactoring it but put it aside for something else. Will refactor it now.

rest-service/manager_rest/security/user_handler.py
geokala: Hmm. It's also duplicating what the actual error is- I put them in because of the function needing to return them. However, I note your refactor comment, so I'll mark this as resolved and they'll likely just disappear if I do that.

rest-service/manager_rest/utils.py
geokala: I think it's passed in as a string when creating the token, but let me double check whether I fixed that.

geokala: It's a string when it's used in the token auth. I could pull the logic out to there, but this way it'll work if passed a datetime or if passed a datetime-string formatted in the way we usually do, so I'm inclined to leave it in.

geokala:
Tests require rest client modifications. Therefore, tests to be introduced once that is done: https://cloudifysource.atlassian.net/browse/RD-4441

tehasdf (approved):

PR #3589, opened 2022-03-18
RD-4497 Update mgmtworker token on snap restore

geokala:
Syncthing will copy it to the other mgmtworkers and it's only needed when
mgmtworker service starts so shouldn't see any problems.


geokala:
As @tehasdf just pointed out to me elsewhere, we're already doing this in a different place in this file, so I'll close this.

PR #3596, opened 2022-03-22
RD-4332 Replace PUT /secrets/{key} v3 with v3.1 + nicer error

glukhman:
The old v3 endpoint created a plain secret value, which can't be read by the GET endpoint, which expects an encrypted value.
Here, the code from v3.1 is moved to v3 to override the old one, so now both v3 and v3.1 endpoints do the same.

Additionally, trying to GET a plain secret value will now result with a friendly error instead of an obscure 500 one


rest-service/manager_rest/manager_exceptions.py
```diff
@@ -122,6 +122,11 @@ class NoTokenGeneratorError(ManagerException):
     status_code = 401
 
 
+class InvalidTokenFormatError(ManagerException):
```
geokala, line 125: Can we maybe call this something like InvalidFernetFormatError to differentiate from tokens used for auth?

rest-service/manager_rest/rest/resources_v3_1/secrets.py
```diff
@@ -7,18 +7,16 @@
                                          generate_key_using_password)
```
geokala, line 7: Any reason why you've not just copied this over the 3.0 one and putting in the extra decrypt-fail catching there?

rest-service/manager_rest/manager_exceptions.py
```diff
@@ -122,6 +122,11 @@ class NoTokenGeneratorError(ManagerException):
     status_code = 401
 
 
+class InvalidFernetTokenFormatError(ManagerException):
+    error_code = 'invalid_fernet_token_format_error'
+    status_code = 400
```
geokala, line 127: Sorry, should've noticed this before: This is a server side error, so it should be a 500.

geokala (approved):

geokala:
I note that this slightly changes the behaviour when creating a secret with a naming collision- before, hitting api/v3/secrets would result in the secret being silently replaced- with a broken (unenencrypted thus unretrievable) secret.

Now, it will result in an immediate error instead of waiting until someone tries to retrieve the secret to scream, so I don't believe this is something that we need to worry about.

PR #3609, opened 2022-03-25
RD-4588 Allow unlicensed user.get call

geokala:
This is making no modifications so should be safe, and helps out the Stage team.


geokala:
I note that this says it only allows user.get unlicensed- but the user (singular) endpoint has only .get- it gets the current user.

tehasdf (approved):

PR #3610, opened 2022-03-25
RD-4575 components: consider different tenant an external host

tehasdf:
Yeah, I don't know what else to do. 
It doesn't seem there's a REST call we can do that'd make sense, to
attach those IDDs - neither of the clients (local or external) is guaranteed
to have access to both tenants. In fact, that might very well be the reason
we use a separate client, because our current user doesn't have access
to the other tenant.

So, I just give up, and treat the "other tenant" client as an external client.
This means no blocking IDDs between them. Too bad.

### DIFF
--- a/cloudify_types/cloudify_types/utils.py
+++ b/cloudify_types/cloudify_types/utils.py
@@ -67,24 +67,30 @@ def get_desired_operation_input(key, args):
             ctx.node.properties.get(key))
 
 
+def _is_client_external(client, client_config):
+    tenant = client_config.get('tenant')
+    if tenant and tenant != ctx.tenant_name:
+        return True
+
+    manager_ips = set()
+    for mgr in manager.get_rest_client().manager.get_managers():
+        manager_ips |= {mgr.private_ip, mgr.public_ip}
+    internal_hosts = {'127.0.0.1', 'localhost'} | manager_ips
+    host = {client.host} if isinstance(client.host, str) else set(client.host)
+    is_internal = host & internal_hosts
+    if is_internal and (host - internal_hosts):
+        ctx.logger.warning(
+            'Only partial match between the destination client host and '
+            'the local manager: the following IPs are external: %s. '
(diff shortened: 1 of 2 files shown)


cloudify_types/cloudify_types/utils.py
```diff
@@ -67,24 +67,30 @@ def get_desired_operation_input(key, args):
...
+    if tenant and tenant != ctx.tenant_name:
+        return True
+
+    manager_ips = set()
+    for mgr in manager.get_rest_client().manager.get_managers():
+        manager_ips |= {mgr.private_ip, mgr.public_ip}
+    internal_hosts = {'127.0.0.1', 'localhost'} | manager_ips
+    host = {client.host} if isinstance(client.host, str) else set(client.host)
+    is_internal = host & internal_hosts
+    if is_internal and (host - internal_hosts):
```
tehasdf, line 81: I'm also changing this condition, because the previous impl also emitted the warning in case of no match at all. Whoops.

glukhman: Nice catch!

glukhman:
@tehasdf with the local client (after we checked that the host is not external), we can pass the tenant here: https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/rest/resources_v3_1/deployments.py#L575 through the REST endpoint, then use _all_tenants and filter by tenant, so we won't get the wrong deployment.
It will of course require adding both source and target tenants as optional params to the EP, which is a bit messy

tehasdf:
@glukhman the problem is having access to those deployments. One of the users (the currently-running user, or the username/password in client_config) would need to have access to both tenants, and we currently don't require that. Can we? Do we want to?

I mean: this needs to be consistent inside of a single request. Considering this as part of a workflow is one thing, but ultimately, there's a REST request fired, and the REST request would be: "create an IDD between dep1, and dep2, but I don't have access to dep2". I don't think we can allow that.

PR #3612, opened 2022-03-25
RD-4171 DSL data type `node_type`

mateumann:


rest-service/manager_rest/rest/resources_v3_1/searches.py
```diff
@@ -187,16 +193,36 @@ class NodesSearches(ResourceSearches):
...
 
+class NodeTypesSearches(ResourceSearches):
+    @swagger.operation(**_swagger_searches_docs(models.Node, 'nodes'))
+    @authorize('node_list', allow_all_tenants=True)
+    @rest_decorators.marshal_with(models.Node)
+    @rest_decorators.paginate
+    @rest_decorators.sortable(models.Node)
+    @rest_decorators.all_tenants
+    @rest_decorators.search('type')
+    def post(self, _include=None, pagination=None, sort=None,
```
geokala, line 213: Why is the search a post rather than a get? I'm guessing something to do with the size of the parameters?

tehasdf: yeah, all the searches are :/

tehasdf (approved):

rest-service/manager_rest/rest/resources_v3_1/searches.py
mateumann: yeah, that's the case (sometimes)

PR #3613, opened 2022-03-27
RD-3385 Cope with node names containing ':' in dep-update

glukhman:
This was encountered by TD.


glukhman:
Since up to this point we don't validate that node names don't contain ":", we need to cope with them.
That said, maybe we can add a deprecation warning in the DSL validation, and disallow such naming in 6.5/7

tehasdf (approved):

PR #3614, opened 2022-03-27
Rd 4586 refactor set visibility

glukhman:


rest-service/manager_rest/storage/resource_models_base.py
```diff
@@ -137,3 +137,7 @@ def list_labels(labels):
                       created_at=label.created_at,
                       creator_id=label.creator.id)
                 for label in labels]
+
+    def check_unique_query(self):
+        if self.is_id_unique:
```
tehasdf, line 142: we now have a good chance to get rid of `.is_id_unique` entirely - just implement this method in the relevant classes, and do `def check_unique_query(self): return`

Beause:
1. I'd rather we had only one thing related to uniqueness (this method) rather than two (method + flag)
2. I've always felt the flag is named backwards anyway :P

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -475,18 +475,16 @@ def _validate_pagination(pagination_size):
...
         # Only relevant for resources that have unique IDs and are connected
         # to a tenant
-        if not instance.is_resource or not instance.is_id_unique:
+        if not instance.is_resource or not instance.check_unique_query():
             return
 
-        query = self._get_unique_resource_id_query(instance.__class__,
-                                                   instance.id,
-                                                   instance.tenant)
+        query = self._get_unique_resource_query(instance)
```
tehasdf, line 487: it'd be good to only get the query once (ie. only call `check_unique_query` once), instead of here, and in `_get_unique_resource_query` as well

I'm thinking just something like `if not is_resource: return;  query = instance.check_query();  if not query: return`

Then, `_get_unique_resource_query` can just be inlined here, or even into the base class method

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -498,26 +496,28 @@ def _validate_unique_resource_id_per_tenant(self, instance):
...
             self._safe_commit()
 
-            raise manager_exceptions.ConflictError(
-                '{0} already exists on {1} or with global visibility'.format(
-                    instance,
-                    self.current_tenant
-                )
-            )
+            if instance.visibility == VisibilityState.GLOBAL:
+                error_msg = "Can't set or create the resource {0}, it's " \
```
tehasdf, line 500:
```suggestion
                error_msg = "Can't set or create the resource {0}, its " \
```

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -5,6 +5,7 @@
 from os import path
 from datetime import datetime
 from collections import namedtuple
+from sqlalchemy import and_ as sql_and
```
tehasdf, line 8: no need to import it:
1. `db.and_` (ditto `db.or_`)
2. just multiple `.filter().filter()` calls are ANDed anyway

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -229,6 +230,12 @@ def installation_state(cls):
         return db.relationship('_PluginState', cascade='delete',
                                passive_deletes=True)
 
+    def check_unique_query(self):
+        return self.__class__.query.filter(sql_and(
+            self.__class__.package_name == self.package_name,
+            self.__class__.archive_name == self.archive_name
+        ))
```
tehasdf, line 237: might as well check the id too

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -1930,6 +1937,12 @@ def set_deployment(self, deployment):
     def set_actual_planned_node_instances(self, num):
         self.actual_planned_number_of_instances = num
 
+    def check_unique_query(self):
+        return self.__class__.query.filter(sql_and(
+            self.__class__.id == self.id,
+            self.__class__._deployment_fk == self._deployment_fk
+        ))
```
tehasdf, line 1944:
```suggestion
        return super().check_unique_query().filter(
            self.__class__._deployment_fk == self._deployment_fk
        )
```

then, no need for `sql_and`, and we can take advantage of the superclass' possibly setting other filters (tenant) too

rest-service/manager_rest/storage/resource_models_base.py
```diff
@@ -137,3 +137,7 @@ def list_labels(labels):
...
+            return self.__class__.query.filter(self.__class__.id == self.id)
```
tehasdf, line 143: if other classes use `super()`, then this could apply the tenant filter, and sm's `_get_unique_resource_query` could be entirely removed

tehasdf: also the tenant/visibility filter can be written as `self.__class__.query.tenant(instance.tenant).filter(self.__class__.id == self.id)`

rest-service/manager_rest/resource_manager.py
```diff
@@ -2352,79 +2349,37 @@ def set_deployment_visibility(self, deployment_id, visibility):
...
-        max_resource_number = 0 if create_resource else 1
-        if self.sm.count(
-            model_class, unique_filter, all_tenants=True
-        ) > max_resource_number:
-            raise manager_exceptions.IllegalActionError(
-                "Can't set or create the resource `{0}`, it's visibility "
-                "can't be global because it also exists in other tenants"
-                .format(resource_id)
-            )
-
```
tehasdf, line 2427: nice!

tehasdf (approved):

PR #3618, opened 2022-03-29
RD-4626 create_auth_token: replace db.session with sm.put

tehasdf:
Avoid committing explicitly, because that could break the encapsulating
transaction. Instead, sm.put commits only when necessary (ie. when
we're not in a transaction).

Also, `user=self,` rather than `_user_fk`, because providing a fk
means that `t.user` will only be set after committing (when the token
object gets repopulated from the database).


tehasdf:
An alternative is to do the bigger refactor, ie. prepare_message+send 2-step process for all messages (hooks, service tasks). Not yet, but I've made RD-4627 about it

PR #3631, opened 2022-04-07
Make dep ID optional on most searches

geokala:


rest-service/manager_rest/rest/resources_v3_1/searches.py
```diff
@@ -363,20 +369,20 @@ def post(self, search=None, _include=None,
         )
 
 
-def retrieve_deployment_id_and_constraints(constraints_optional=True):
+def retrieve_deployment_id_and_constraints(dep_id_required=False):
```
mateumann, line 370: `deployment_id` is required for searches based on constraints for: capability values, node ids, node types, node instance ids and (soon) scaling groups.

mateumann: … It does not really matter if it comes as one of the constraints or query parameter, but it absolutely must be there.

rest-service/manager_rest/rest/resources_v3_1/searches.py
```diff
@@ -363,20 +367,20 @@ def post(self, search=None, _include=None,
...
+        {'constraints': {'optional': True, 'type': dict}})
     constraints = request_dict.get('constraints', {})
     if args.get('deployment_id') and 'deployment_id' in constraints:
         raise manager_exceptions.BadParametersError(
             "You should provide either a valid '_deployment_id' parameter "
             "or have a 'deployment_id' key in the constraints, not both.")
     deployment_id = args.get('deployment_id') \
         or constraints.get('deployment_id')
-    if not deployment_id:
+    if dep_id_required and not deployment_id:
```
mateumann, line 383: maybe then something like:
```
if (dep_id_required or constraints) and not deployment_id:
    ...
```

rest-service/manager_rest/rest/resources_v3_1/searches.py
geokala: This breaks the old behaviour of the node instances search, will discuss separately and update here with conclusions.

geokala: After discussion: deployment ID is required when constraints are provided. I will update the PR for this.

rest-service/manager_rest/rest/resources_v3_1/searches.py
geokala: Updating shortly.

rest-service/manager_rest/rest/resources_v3_1/searches.py
```diff
@@ -363,20 +367,21 @@ def post(self, search=None, _include=None,
...
     constraints = request_dict.get('constraints', {})
     if args.get('deployment_id') and 'deployment_id' in constraints:
         raise manager_exceptions.BadParametersError(
             "You should provide either a valid '_deployment_id' parameter "
             "or have a 'deployment_id' key in the constraints, not both.")
     deployment_id = args.get('deployment_id') \
         or constraints.get('deployment_id')
-    if not deployment_id:
+    # A deployment ID is necessary if constraints are provided
+    if (constraints or dep_id_required) and not deployment_id:
```
mateumann, line 384: based on our conversation on Slack maybe it should be:
```
if (constraints is not None or dep_id_required) and not deployment_id:
```
?

geokala: constraints will never be none- it is retrieved as a dict above. I could make the check more complicated, if it's really worth it?

geokala: (line 372/376)

mateumann: constraints could be None if it's not our REST client we're using (but UI for example

mateumann: I am writing to fast and not reading fast enough, sorry, sec

geokala: It still can't be- see line 372/376 (and the line before it):
We try to get constraints from the request dict. If it's there, it must be a dict. If it's not there, we set it to an empty dict.

geokala: So we can put a check in for whether 'constraints' is in request_dict, but I'm not sure if that's worth doing. We enforce the dep ID being set for several of the searches, and that's presumably enough for most cases?

mateumann: What if we dropped even `dep_id_required`?

```python
    request_dict = rest_utils.get_json_and_verify_params(
        {'constraints': {'optional': True, 'type': dict}})
    have_constraints = 'constraints' in request_dict
    constraints = request_dict.get('constraints', {})
    if args.get('deployment_id') and 'deployment_id' in constraints:
        raise manager_exceptions.BadParametersError(
            "You should provide either a valid '_deployment_id' parameter "
            "or have a 'deployment_id' key in the constraints, not both.")
    deployment_id = args.get('deployment_id') \
        or constraints.get('deployment_id')
    # A deployment ID is necessary if constraints are provided
    if have_constraints and not deployment_id:
        raise manager_exceptions.BadParametersError(
            "Please provide a valid '_deployment_id' parameter or have "
            "a 'deployment_id' key in the constraints.")
```

geokala: dep_id_required I left in because the capabilities does need to have a dep ID- so even if constraints were not supplied for whatever reason it would want to end up with a dep ID. I could put in the extra constraints check though, yes- you think it's worthwhile?

mateumann: ok, yeah, maybe.  I think I'll rewrite the capabilities endpoint one day, maybe even sooner than later :)

geokala: Fair enough!

PR #3634, opened 2022-04-08
RD-4712 Stop supervisord giving spurious CRIT log

geokala:


rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -67,6 +67,7 @@ def __init__(self, path):
         self._path = path
 
     def make_connection(self, host):
+        _, self._extra_headers, _ = self.get_host_info(host)
```
geokala, line 70: This is copied from the xmlrpc transport we are overriding.

rest-service/manager_rest/rest/resources_v3_1/status.py
```diff
@@ -210,7 +211,7 @@ def _lookup_supervisor_service_status(service_name):
     service_status = None
     is_optional = True if service_name in OPTIONAL_SERVICES else False
     server = xmlrpclib.Server(
-        'http://',
+        'http://stopspurious:logmessages@localhost',
```
geokala, line 214: Creds set in c-m-i. Hardcoded because they're utterly meaningless.

geokala:
We can't use creds on this because node_exporter won't support them, I'll find another way.

PR #3635, opened 2022-04-08
RD-4615 Extend `valid_values` based on types hierarchy

mateumann:


rest-service/manager_rest/rest/resources_v3_1/searches.py
```diff
@@ -551,3 +554,21 @@ def scaling_group_name_matches(scaling_group_name, constraints, search_value):
...
+        models.Node,
+        filters={'deployment_id': deployment_id},
+        include='id,type,type_hierarchy',
+        get_all_results=True)
+    results = []
+    for value in valid_values:
+        for node in nodes:
+            if value in node.type_hierarchy:
+                th = node.type_hierarchy
+                results.extend(th[th.index(value):])
```
mateumann, line 573: Luckily `type_hierarchy` will preserve the order of inheritance with the root node being the first in the list and the actual node.type - the last.  For example:

```json
"id": "myserver",
"type": "webserver",
"type_hierarchy": [
    "cloudify.nodes.Root",
    "cloudify.nodes.SoftwareComponent",
    "cloudify.nodes.WebServer",
    "webserver"
]
```

tehasdf (approved):

PR #3642, opened 2022-04-13
RD-4711 Default `deployment_id` constraint value

mateumann:
... when running workflows


rest-service/manager_rest/rest/search_utils.py
```diff
@@ -322,6 +329,14 @@ def get_node_instances(self, node_instance,
             filter_rules=filter_rules
         )
 
+    def update_deployment_id_constraint(self, data_type, **kwargs):
+        if data_type not in TYPES_WHICH_REQUIRE_DEPLOYMENT_ID_CONSTRAINT:
+            return kwargs
+        params = copy(kwargs)
+        if 'deployment_id' not in kwargs.keys():
```
tehasdf, line 336:
```suggestion
        if 'deployment_id' not in kwargs:
```

PR #3643, opened 2022-04-14
RD-4627 Use a long-running connection for amqp

tehasdf:
Instead of creating the amqp client when necessary, create it on first
access, and store it. That way, we only pay the high cost of setting up
a TLS connection to rabbitmq once.


geokala:
I'm maybe missing something, but what happens if the long running connection drops during a period of inactivity?

tehasdf:
if it's dropped, it gets reconnected as usual by amqp_client, up to the deadline set by connect_timeout (which I set to None here). amqp_client does the backoff and it'll eventually retry every 30 seconds, by way of https://github.com/cloudify-cosmo/cloudify-common/blob/5b60129d7d033640e0e069dd46cf482021b98d08/cloudify/amqp_client.py#L195-L198

geokala:
OK, sounds great then- I saw you put no connect timeout but didn't realise it behaved like that. Will approve.

geokala (approved):

PR #3649, opened 2022-04-27
RD-4817 Support include on plugin updates

geokala:
Without this, including both blueprint_id and temp_blueprint_id causes two
outer joins against the same table, which results in sadness and exception.

### DIFF
--- a/rest-service/manager_rest/storage/storage_manager.py
+++ b/rest-service/manager_rest/storage/storage_manager.py
@@ -149,11 +149,7 @@ class SQLStorageManager(object):
             query = query.distinct(*distinct)
 
         if joins:
-            query = query.options(db.joinedload(join) for join in joins
-                                  if join.can_joinedload)
-            outer_joins = [join for join in joins if not join.can_joinedload]
-            if outer_joins:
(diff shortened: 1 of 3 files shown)


mateumann (approved):

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -149,7 +149,11 @@ def _get_base_query(self, model_class, include, joins, distinct=None,
             query = query.distinct(*distinct)
 
         if joins:
-            query = query.outerjoin(*joins)
+            query = query.options(db.joinedload(join) for join in joins
+                                  if join.can_joinedload)
+            outer_joins = [join for join in joins if not join.can_joinedload]
+            if outer_joins:
+                query.outerjoin(*outer_joins)
```
tehasdf, line 156:
```suggestion
                query = query.outerjoin(*outer_joins)
```

geokala: Whoops, nice catch.

geokala:
Rebased and force pushed

glukhman (approved):

PR #3651, opened 2022-05-02
RD-4826 Deduplicate user loader

tehasdf:
The main point of this is to clean up the confusion between
user_handler.py and authenticate.py - out of those files, which one
does what?

Well, now it's cleaner - user_handler loads the user object, based on the
request, and authenticate is just a decorator that requires the user to
be logged in.

All of the user-loading ("authentication") logic is now in user_loader.
You are very much invited to point out any discrepancies between this
implementation and the previous one, but I think it does the same
thing. But it's for sure clearer to see what are the sources of the user
creds, and what's the ordering between them.

However! To do this, I needed tests to behave. We used to set
current_user in tests from just the test level. That's stupid, because
that allows the tests to have a very different auth flow than the regular
REST calls. So I'm doing away with having request context in test
functions themselves.

...but the test functions do use the StorageManager, and that required
request context. So in the first commits, I'm making it not require
request context anymore.

And so, make sure to read commits separately, lest you go mad.


tehasdf:
The accompanying c-premium PR is cloudify-cosmo/cloudify-premium#855

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -293,11 +295,11 @@ def _add_tenant_filter(self, query, model_class, all_tenants):
         if all_tenants:
             # If a user that is allowed to get all the tenants in the system
             # no need to filter
-            if all_tenants_authorization():
+            if all_tenants_authorization(self.current_user):
```
geokala, line 298: Is there actually a current_user attribute on here?

geokala: Ignore that, I didn't scroll down far enough.

tehasdf: there is now!

rest-service/manager_rest/security/user_handler.py
```diff
@@ -1,67 +1,159 @@
...
+    current_execution,
+)
+from manager_rest.security import audit
+from manager_rest.utils import (
+    is_expired,
+    is_sanity_mode,
+)
 
 
-ENCODED_ID_LENGTH = 5
```
geokala, line 20: Nice catch- I noticed on friday that I forgot to remove that, was thinking of doing it at some point so I'm glad you nixed it.

rest-service/manager_rest/security/user_handler.py
```diff
@@ -1,67 +1,159 @@
...
-    """Attempt to retrieve the current user from the request
-    Either from request's Authorization attribute, or from the token header
+    Try several ways of retrieving the user: each of these retrieval functions
+    are expected to decide whether they should be the one to get the user,
+    and return one, or None. The functions can also throw, or return an error
+    response immediately.
+    """
+    for getter, method_label in [
+        (_get_user_from_execution_token, 'execution_token'),
+        (_get_user_from_auth, 'http_basic'),
```
geokala, line 49: Doesn't this ordering result in ldap breaking as it uses http basic auth then passes through?

rest-service/manager_rest/security/user_handler.py
```diff
@@ -1,67 +1,159 @@
...
 
-def user_loader(request):
-    """Attempt to retrieve the current user from the request
-    Either from request's Authorization attribute, or from the token header
+    Try several ways of retrieving the user: each of these retrieval functions
+    are expected to decide whether they should be the one to get the user,
+    and return one, or None. The functions can also throw, or return an error
+    response immediately.
+    """
+    for getter, method_label in [
```
geokala, line 47: This is nice- much clearer ordering than the old method.

rest-service/manager_rest/utils.py
```diff
@@ -25,15 +25,21 @@
 from manager_rest import constants, config, manager_exceptions
 
 
+def check_unauthenticated_endpoint():
+    """Is the accessed request unauthenticated?"""
+    if request.endpoint is None:
+        return False
+    request_endpoint = _endpoint_strip_version(request.endpoint)
+    return request_endpoint in constants.UNAUTHENTICATED_ENDPOINTS
```
geokala, line 33: Can't this be determined by seeing which endpoint flask thinks it's routing to, given that said endpoint is not descended from securedresource? (just thinking to avoid having to remember to add things to a list in the unlikely event that we add more unauthed endpoints)

rest-service/manager_rest/security/user_handler.py
tehasdf: Good point. The `7da7b126b Only allow the bootstrap admin to bypass ext auth` commit should address this, and keep the previous behaviour. Now, if ext-auth is configured, this function will be a no-op for all but the bootstrap admin.

rest-service/manager_rest/utils.py
tehasdf: Right, I thought about that, but I didn't find a good way to do this.

First: at this point, we don't yet know the endpoint directly. We could figure it out, like this:
```python
# this returns the .as_view method of the resource
viewfunc = current_app.view_functions[request.endpoint]
# now this gives us the resource
view_class = viewfunc.view_class

if isinstance(view_class, SecuredResource): ...
```

but I don't quite like this, because:
1. I'm not sure checking for SecuredResource to be the right place. Ideally we'd check for the existence of the authenticate decorator, but there's no good way to do that, especially in face of...
2. this then assumes all our viewfunctions are indeed descended from Resource, and are class-based indeed. That's not inevitable, we might have eventually some simple function-based views
3. the obvious way of writing this `if` is unauthenticated-default. `return isinstance(x, SecuredResource)` means that the default, eg. if the view is just a function, we default to unauthenticated. I think auth here should be opt-out, not opt-in.

And given how critical this is (failure here is catastrophic, you allow unauthenticated access!!), I'm not all that averse to a very-explicit "just a list of endpoints", even at the cost of increased burden while implementing new unauthenticated endpoints

geokala: We could dynamically build the list of endpoints when something has the authenticate decorator added I think?

But yeah, fair point, given the impact of failure I agree, let's make it very explicit.

geokala (approved):

PR #3653, opened 2022-05-04
plugins update workflow: remove debug label from the event

tehasdf:
Hi @mateumann


geokala (approved):
Yeah, this makes sense. If it's not logging to /tmp/ilikecake then it's clearly wrong.

PR #3656, opened 2022-05-05
RD-4640 Set NI status & drift when installing it

tehasdf:
Automatically set those properties. Values TBD, but for right now,
they're this.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/operations.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/operations.py
@@ -263,7 +263,7 @@ class OperationsId(SecuredResource):
                     'task': None,
                     'timestamp': datetime.utcnow().isoformat(),
                 })
-        if state == 'started':
+        elif state == 'started':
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/rest/resources_v3_1/operations.py
```diff
@@ -253,8 +253,28 @@ def _on_success_SetNodeInstanceStateTask(self, sm, operation):
...
+            node_instance.system_properties = {}
+        if state == 'configured':
+            node_instance.system_properties.setdefault(
+                'configuration_drift', {
+                    'ok': True,
+                    'result': None,
+                    'task': None,
+                    'timestamp': datetime.utcnow().isoformat(),
+                })
+        if state == 'started':
```
mateumann, line 266: `elif`?

tehasdf: sure

mateumann (approved):

PR #3674, opened 2022-05-17
RD-4612 Add drifted/unavailable instance counts

tehasdf:
I'll hide these behind a querystring flag. It's not ideal, but we need to
examine all instances, and there might be quite a few of them.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v2/nodes.py
+++ b/rest-service/manager_rest/rest/resources_v2/nodes.py
@@ -17,6 +17,7 @@
 from flask import request
 from flask_restful_swagger import swagger
 
+from manager_rest import manager_exceptions
 from manager_rest.rest import (
     resources_v1,
     rest_decorators,
@@ -57,6 +58,10 @@ class Nodes(resources_v1.Nodes):
             '_get_all_results',
             request.args.get('_get_all_results', False)
         )
+        instance_counts = rest_utils.verify_and_convert_bool(
+            '_instance_counts',
+            request.args.get('_instance_counts', False)
+        )
         nodes_list = get_storage_manager().list(
             models.Node,
             include=_include,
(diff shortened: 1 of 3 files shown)


glukhman (dismissed):

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -2038,6 +2046,43 @@ def set_node(self, node):
...
+        """Has the last status check for this NI succeeded?
+
+        This examines the result of the most recent check_status call
+        on this node instance, and returns whether the call succeeded.
+
+        If the result is missing, the result is succeeded.
+        """
+        props = self.system_properties or {}
+        status = props.get('status') or {}
+        return bool(status.get('ok', True))
```
geokala, line 2068: This might be one of the "just 'cause" ones, but why is the bool conversion needed here? Isn't the props['status'] either bool or nothing?
I mean, if it's "false" then this'll interpret it as true which seems wrong.

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -2038,6 +2046,43 @@ def set_node(self, node):
...
+        props = self.system_properties or {}
+        instance_drift = props.get('configuration_drift') or {}
+        sources_drift = \
+            props.get('source_relationships_configuration_drift') or {}
+        targets_drift = \
+            props.get('target_relationships_configuration_drift') or {}
+        for drift in itertools.chain(
+            [instance_drift], sources_drift.values(), targets_drift.values()
+        ):
+            if not drift.get('ok', True) or drift.get('result') is not None:
```
geokala, line 2090: When would drift['ok'] be True, but drift['result'] be not null?

rest-service/manager_rest/storage/resource_models.py
tehasdf: it's not needed, but it is just in case indeed. Meh, I'll keep it.

rest-service/manager_rest/storage/resource_models.py
tehasdf: `ok` is whether the operation succeeded, and result is... whatever the operation returns, which is up to the plugin author

tehasdf: So there's the question of how do we want to signal drift, either by raising an exception, or by returning something, and here we allow both...

rest-service/manager_rest/rest/resources_v2/nodes.py
```diff
@@ -80,8 +85,33 @@ def get(self, _include=None, filters=None, pagination=None,
                     scale_by *= group['properties']['planned_instances']
             node.set_actual_planned_node_instances(
                 scale_by * node.planned_number_of_instances)
+        if instance_counts:
+            if not filters.get('deployment_id'):
```
geokala, line 89: Why does this need to have the dep id requirement? It looks like the nodes_list doesn't.

tehasdf: just to limit it. I don't want to pull in too many node-instances. Like, this is a pretty expensive operation. In RD-4408 we can consider doing something for >1 deployments.

rest-service/manager_rest/storage/resource_models.py
geokala: Hmm... I suppose the operation checking the drift could set informational stuff in its logs rather than output. I'm not sure whether this feels right or not, but I guess we can just wait and see how this ends up being used as it could go either way.

rest-service/manager_rest/rest/resources_v2/nodes.py
geokala: <thumbs up>

geokala (approved):

PR #3678, opened 2022-05-18
RD-4923 Add manager_name/agent_name to logs/events

tehasdf:
And also to operations!

Now, it'll be possible to know which manager (or which agent) emitted
an event, or was the one handling an operation.


tehasdf:
needs cloudify-cosmo/cloudify-common#1055

PR #3679, opened 2022-05-18
RD-3813 Add log fetcher

geokala:


packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
```diff
@@ -0,0 +1,117 @@
...
+        cred_value = os.environ.get(env_var)
+        if not cred_value:
+            sys.stderr.write(
+                f'Expected {cred} to be provided in env var {env_var}.\n'
+            )
+        creds[cred] = cred_value
+
+    if not all(val for val in creds.values()):
+        sys.stderr.write('Credentials not provided correctly.\n')
+        sys.exit(1)
```
tehasdf, line 114: ha, you know how much simpler would this be, if you just did `creds = {'username': os.environ['MONITORING_USERNAME']}`

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
```diff
@@ -0,0 +1,117 @@
...
+            )
+
+    for dir_name in sorted(dirs):
+        get_logs_dir(server, path + dir_name + '/',
+                     data_dir, log_handle, session)
+
+
+def get_logs(servers, ca_cert, credentials):
+    try:
+        temp_path = mkdtemp()
```
tehasdf, line 58: move this to before the try, because if the mkdtemp call fails, the NameError in finally will obscure the reason

tehasdf, line 58: but actually, just use `with tempfile.TemporaryDirectory() as temp_path`

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
```diff
@@ -0,0 +1,117 @@
...
+BASE_URL = 'https://{server}:8009/cfylogs/'
+
+
+def get_logs_dir(server, path, data_dir, log_handle, session):
+    save_path = os.path.join(data_dir, server, path)
+    url = BASE_URL.format(server=server) + path
+
+    contents = session.get(url)
+
+    os.makedirs(save_path, exist_ok=True)
```
tehasdf, line 21: move this to after the `if`, no need to create the dir at all if we fail to grab the index

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
```diff
@@ -0,0 +1,117 @@
...
+
+    for file_name in sorted(files):
+        downloaded_file = session.get(url + file_name)
+        file_path = path + file_name
+
+        if downloaded_file.status_code == 200:
+            log_handle.write(f'Retrieved file {file_path}\n')
+            with open(os.path.join(save_path, file_name), 'w',
+                      encoding='utf8') as file_handle:
+                file_handle.write(downloaded_file.text)
```
tehasdf, line 44: I think there's no reason to go through the decode/encode dance here. We can just dump the bytes as they come:
```suggestion
            with open(os.path.join(save_path, file_name), 'wb') as f:
                f.write(downloaded_file.content)
```

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
```diff
@@ -0,0 +1,117 @@
...
+        else:
+            log_handle.write(
+                f'Failed to retrieve file {file_path}: '
+                f'{downloaded_file.status_code}- {downloaded_file.reason}\n'
+            )
+
+    for dir_name in sorted(dirs):
+        get_logs_dir(server, path + dir_name + '/',
+                     data_dir, log_handle, session)
```
tehasdf, line 53: I vote for writing this function with a while loop rather than recursively

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
geokala: Right, but I wanted it to actually log all the missing env vars rather than complaining one thing at a time because that's really annoying.

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
geokala: Oh yeah, good point.
Updated.

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
```diff
@@ -0,0 +1,117 @@
...
+        zip_file.close()
+        zip_path = zip_file.name
+
+        username = credentials['username']
+        password = credentials['password']
+
+        for server in servers:
+            session = requests.Session()
+            session.auth = (username, password)
+            session.verify = ca_cert
```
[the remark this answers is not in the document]
geokala: Oh yes, good point- for some reason I had got it into my head that the session contained the URL which clearly makes no sense. Updating.

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
geokala: Good catch.

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
geokala: I had it throwing errors at me without the encoding being set, which is why I shoved it in there in the first place.

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
geokala: That feels like it'll come out more complicated, but I'll have a play with it and see how it looks.

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
geokala: I was trying to remember why I chose this route, but yeah- I should do that. Will update.

packaging/mgmtworker/files/opt/mgmtworker/scripts/fetch-logs
tehasdf: but .content is bytes, .text is text

geokala: Ah, that'll be it- will give it a check

tehasdf (approved):

PR #3688, opened 2022-05-23
RD-5007 Admin-only config values

tehasdf:
Implement the concept of an "admin-only" setting, whose values
are only viewable by admins, and only changeable by admins.


geokala:
log_fetch credentials (log_fetch_{username,password}) should also be admin only.

geokala (approved):

PR #3694, opened 2022-05-27
RD-4619 Implement `node_instances_active` availability rule

mateumann:


rest-service/manager_rest/storage/resource_models.py
```diff
@@ -590,9 +590,22 @@ def _is_workflow_available(self, workflow):
...
 
+    def _node_instances_active_states_match(self, node_instance_active_rule):
+        ni_states = set(ni.state for n in self.nodes for ni in n.instances)
+        if node_instance_active_rule == 'all':
+            return all(state == 'started' for state in ni_states)
+        if node_instance_active_rule == 'some':
+            return any(state == 'started' for state in ni_states)
+        if node_instance_active_rule == 'none':
+            return not any(state == 'started' for state in ni_states)
+        return False
```
mateumann, line 607: maybe `return True`?

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -590,9 +590,22 @@ def _is_workflow_available(self, workflow):
             return True
         if not rules.get('available', True):
             return False
+        ni_active_rule = rules.get('node_instances_active')
+        if ni_active_rule:
+            return self._node_instances_active_states_match(ni_active_rule)
         # TODO add other availability rules checking here
```
geokala, line 596: We probably shouldn't have a TODO in here

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -590,9 +590,22 @@ def _is_workflow_available(self, workflow):
             return True
         if not rules.get('available', True):
```
geokala, line 591: Perhaps the entire rules thing should be done by matching each specified key against a dict that defines the function used to validate them?

e.g. return all(MAPPING[rules[rule]] for rule in rules if rule in MAPPING)

rest-service/manager_rest/storage/resource_models.py
geokala: Seems like an exceptional condition to me?

rest-service/manager_rest/storage/resource_models.py
mateumann: why not?  there will be a few more, which I'm going to implement soon.  I'm keeping that as a reminder/placeholder

rest-service/manager_rest/storage/resource_models.py
mateumann: Yeah, maybe.  I'd like to work on it once I implement some other rules and we'll see some pattern(s)

rest-service/manager_rest/storage/resource_models.py
mateumann: probably, yeah, I'll just throw an error then

rest-service/manager_rest/storage/resource_models.py
geokala: As long as it's gone by release.

I note though that as it stands (without the mapping based change I suggested above, or similar), if there is a ni_active_rule and it states that the workflow is available then the workflow will be available even if any following rules want to return False.

rest-service/manager_rest/storage/resource_models.py
geokala: It might be worth doing that now (see my response to your other comment below) to make that be the pattern, and only break it out into something more bespoke if we discover that we really can't fit the pattern- though presumably all of the rules will have an accompanying value from which can be determined whether that rule considers the workflow to be available or not, so this pattern should hold.

tests/integration_tests/tests/agentless_tests/deployment_update/test_deployment_update_modification.py
```diff
@@ -540,3 +540,54 @@ def test_enable_workflow(self):
         assert cm.exception.error_code == 'unavailable_workflow_error'
         self._do_update(deployment.id, BLUEPRINT_ID)
         self.execute_workflow('wf1', deployment.id)  # doesn't throw
+
+    def test_workflows_availability_node_instances_state(self):
```
tehasdf, line 544: I think I'd prefer to see this test not tied to update, sorry.

The other test is in here, because the `available` rule is strictly tied to update - updating the deployment is the only way to change that flag.

But for `node_instances_active`, I don't see a reason to make it related to update - I suggest instead of updating, just have wf1, wf2, and wf3 in the blueprint, and check them all directly.

It's also going to be a way faster test then

rest-service/manager_rest/storage/resource_models.py
mateumann: I think it's not going to be that simple.  See @tehasdf 's comment on "AND-ing" `node_instances_active` rule

tests/integration_tests/tests/agentless_tests/deployment_update/test_deployment_update_modification.py
mateumann: :+1:

rest-service/manager_rest/storage/resource_models.py
geokala: This will AND them- using all(<conditions>)

geokala (approved):

tehasdf (approved):

PR #3695, opened 2022-05-27
RD-4621 State workflow availability for default workflows

mateumann:


mateumann:
I'm not sure about these default workflows.  I didn't define any `node_instances_active` availability rules for them, but maybe I should?
* check_status,
* execute_operation,
* install_new_agents,
* validate_agents,
* pull

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -598,19 +598,26 @@ def _is_workflow_available(self, workflow):
...
+                          and any(state != 'started' for state in ni_states)
+            elif rule == 'none':
+                result |= not any(state == 'started' for state in ni_states)
+            else:
+                raise manager_exceptions.InvalidWorkflowAvailabilityRule(
+                    "Invalid value for 'node_instances_active' availability "
+                    f"rule: '{node_instance_active_rules}'")
+            if result is True:
+                return True
+        return False
```
tehasdf, line 620: hmm, since you made the result into a set, this can be written simpler
```suggestion
        result = False
        for rule in node_instance_active_rules:
            if rule == 'all':
                result |= (ni_states == {'started'})
            elif rule == 'partial':
                result |= (len(ni_states) > 1 and 'started' in ni_states)
            elif rule == 'none':
                result |= ('started' not in ni_states)
            else:
                raise manager_exceptions.InvalidWorkflowAvailabilityRule(
                    "Invalid value for 'node_instances_active' availability "
                    f"rule: '{node_instance_active_rules}'")
        return result
```
or even:
```suggestion
        for rule in node_instance_active_rules:
            if rule == 'all':
                if ni_states != {'started'}:
                    return False
            elif rule == 'partial':
                if 'started' not in ni_states or ni_states == {'started'}:
                    return False
            elif rule == 'none':
                if 'started' in ni_states:
                    return False
            else:
                raise manager_exceptions.InvalidWorkflowAvailabilityRule(
                    "Invalid value for 'node_instances_active' availability "
                    f"rule: '{node_instance_active_rules}'")
        return True
```

tehasdf:
- check_status might well be partial+all - there's no need to run status checks on a non-installed deployment, it can't possibly pass, can it? Only if the checks aren't defined, but that's not an interesting case to run check_status anyway. Also, in the future, maybe we'll want something like, check_status is only available if the operations is defined anywhere
- `pull` is for pulling new state/config, it's something like check_drift but actually doing the changes, so I think it should be available in the same states that check_drift (so, all+partial is OK)
- `validate_agents` - same as check_status, it cannot possibly pass for `none`
- `install_new_agents` - agent upgrade, it can only run if we already have the old agent, to run the upgrade; so, all+partial. Could also be JUST all, but... hmm. Safer to allow partial too, in case some software on the agent's vm got uninstalled..?
- `execute_operation` - I don't see the reason to define availability there

tehasdf (approved):

tehasdf (approved):

tehasdf (approved):

PR #3698, opened 2022-05-31
Workflow availability fixes

mateumann:


tests/integration_tests/resources/dsl/resumable_mgmtworker.yaml
```diff
@@ -30,3 +30,6 @@ node_templates:
     relationships:
       - type: cloudify.relationships.depends_on
         target: node1
+
+  node3:
+    type: cloudify.nodes.Root
```
tehasdf, line 35: why this?

tests/integration_tests/resources/dsl/blueprint_parent.yaml
```diff
@@ -8,6 +8,10 @@ inputs:
     type: string
     default: "i"
 
+node_templates:
+  x:
+    type: cloudify.nodes.Root
+
```
tehasdf, line 14: do you mean we require some instances to exist, to pass the checks at all? I thought no instances at all, would mean that all instance availability checks pass, because of the `all()` calls we use there?

resources/rest-service/cloudify/types/types.yaml
```diff
@@ -856,8 +856,6 @@ workflows:
   scale:
     mapping: default_workflows.cloudify.plugins.workflows.scale_entity
     is_cascading: false
-    availability_rules:
-      node_instances_active: [ 'all', 'partial' ]
```
tehasdf, line 860: hmm, I think this was correct after all. Why would we want to scale something that isn't installed at all?

tests/integration_tests/resources/dsl/resumable_mgmtworker.yaml
mateumann: to have at least one node `started` for workflow to be available and not mess with the tests

tests/integration_tests/resources/dsl/blueprint_parent.yaml
mateumann: That we do, yeah.  It's similar thing in many other places.

resources/rest-service/cloudify/types/types.yaml
mateumann: because "Don't forbid scaling from zero instances"

tests/integration_tests/resources/dsl/resumable_mgmtworker.yaml
tehasdf: but there already are node templates in this bp, we can just use those, right? I suppose they should be just installed (`self.deploy_application`), that's the primary interface of making nodes into the 'started' state

tehasdf (approved):

PR #3708, opened 2022-06-06
Rd 4569 password node type

glukhman:


cloudify_types/cloudify_types/password/operations.py
```diff
@@ -0,0 +1,107 @@
+from random import choice, shuffle
```
tehasdf, line 1: for security purposes, `random.choice` and `random.shuffle` aren't enough, because their output is predictable. `random.SystemRandom` would be better.

Python3 also exposes `choice` in the `secrets` module, usefully - but alas, not `shuffle`.

cloudify_types/cloudify_types/password/constants.py
```diff
@@ -0,0 +1 @@
+PASSWORD_TYPE = 'cloudify.nodes.Password'
```
tehasdf, line 1: I don't think we actually need this!

cloudify_types/cloudify_types/password/operations.py
```diff
@@ -0,0 +1,107 @@
...
+            'group should be used')
+
+    if min_uppercase + min_lowercase + min_digits + min_symbols > length:
+        raise NonRecoverableError(
+            'Could not satisfy password requirements: lengths of required '
+            'character groups is larger than the password length')
+    return True
+
+
+def _generate_password(len, alphabet):
```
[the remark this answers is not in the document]
glukhman: ah, right

cloudify_types/cloudify_types/password/constants.py
glukhman: also right

cloudify_types/cloudify_types/password/operations.py
```diff
@@ -0,0 +1,103 @@
...
+        for _ in range(min_symbols):
+            password.append(choice(symbols))
+    for _ in range(length - required_length):
+        password.append(choice(alphabet))
+
+    SystemRandom().shuffle(password)
+    password = ''.join(password)
+
+    _set_secrets(manager.get_rest_client(), {secret_name: password})
+    ctx.instance.runtime_properties['secret_name'] = secret_name
```
tehasdf, line 59: now, you say `resumable=True`, but if this function is rerun, `_set_secrets` will abort that the secret already exists. For it to really be resumable, the operation would have to be idempotent

cloudify_types/cloudify_types/password/operations.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@operation(resumable=True)
+@errors_nonrecoverable
+def delete(**kwargs):
+    secret_name = ctx.instance.runtime_properties.get('secret_name')
+    remove_secret = \
+        get_desired_operation_input('uninstall_removes_secret', kwargs)
+    if remove_secret:
+        _delete_secrets(manager.get_rest_client(), [secret_name])
```
tehasdf, line 71: for this to be resumable, this func needs to not break if the secret is already gone

cloudify_types/cloudify_types/password/operations.py
```diff
@@ -0,0 +1,103 @@
...
+    _set_secrets(manager.get_rest_client(), {secret_name: password})
+    ctx.instance.runtime_properties['secret_name'] = secret_name
+    ctx.logger.info('Created password. Secret name: %s', secret_name)
+    return True
+
+
+@operation(resumable=True)
+@errors_nonrecoverable
+def delete(**kwargs):
+    secret_name = ctx.instance.runtime_properties.get('secret_name')
```
tehasdf, line 67: should we also clean the runtime property?

tehasdf (approved):

PR #3711, opened 2022-06-07
RD-4429 REST tests: clean up .put_deployment_with_labels

tehasdf:
You know I hate .put_deployment for how slow it is, and I like to replace
it with just models.Deployment whenever possible.

So here, I'm dropping all uses of .put_deployment_with_labels

Also, some warnings cleanup


rest-service/manager_rest/test/endpoints/test_labels.py
```diff
@@ -85,7 +85,7 @@ def test_remove_resource_labels(self):
         resource = self.put_resource_with_labels(self.LABELS)
         self.assert_resource_labels(resource.labels, self.LABELS)
         updated_res = self.update_resource_labels(resource.id, [])
-        self.assertEmpty(updated_res.labels)
+        assert not updated_res.labels
```
geokala, line 89: assert updated_res.labels == [] might be better given that we were previously making sure it wasn't None?

geokala: In fact, same with the others of this type- that maintains the assertions but shorter and clearer.

rest-service/manager_rest/test/endpoints/test_labels.py
```diff
@@ -86,7 +86,7 @@ def test_remove_resource_labels(self):
         resource = self.put_resource_with_labels(self.LABELS)
         self.assert_resource_labels(resource.labels, self.LABELS)
         updated_res = self.update_resource_labels(resource.id, [])
-        assert not updated_res.labels
+        assert updated_res.labels == []
```
geokala, line 89: Do you want to do this on the other ones from the first commit too?

tehasdf: OK

geokala (approved):

PR #3714, opened 2022-06-08
RD-5061 Move blueprint icons from stage to backend

geokala:


workflows/cloudify_system_workflows/snapshots/copy_icons.py
```diff
@@ -0,0 +1,55 @@
...
+        return
+
+    with setup_flask_app().app_context():
+        config.instance.load_configuration()
+        fs_blueprints_path = os.path.join(config.instance.file_server_root,
+                                          FILE_SERVER_BLUEPRINTS_FOLDER)
+    existing_blueprints = {}
+    for tenant in os.listdir(fs_blueprints_path):
+        tenant_path = os.path.join(fs_blueprints_path, tenant)
+        for blueprint in os.listdir(tenant_path):
```
tehasdf, line 25: maybe an `os.isdir` somewhere here too?

geokala: Hmm... I'm not sure that's required?

This'll only work on a node with stage on it where blueprint icon dump has occurred, which means we know it's a manager node (because at the moment they must be colocated).

Therefore, we know that the blueprints path must exist (I think it's created during bootstrap or thereabouts?).
If that's the case then because we create this path by listing the tenant directories in that dir, this dir must exist.

Any faults in that logic?

tehasdf (approved):

PR #3721, opened 2022-06-15
RD-4624 Make `node_instance_id` param not required for `heal` workflow

mateumann:
In case `check_status=true` we are going to find unhealthy node
instances on our own using check_status workflow.


mateumann:
A different approach will be implemented

PR #3723, opened 2022-06-17
RD-4429 Don't allow manually overwriting reserved labels

mateumann:


rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -27,6 +27,7 @@
 from cloudify.zip_utils import make_zip64_archive
 
 from manager_rest import utils, manager_exceptions, workflow_executor
+from manager_rest.constants import RESERVED_LABELS
```
tehasdf, line 30: can't do all of these. Only `csys-consumer-id` is managed by the manager itself.

tehasdf (changes requested):

tehasdf:
obsoleted by #3728

PR #3725, opened 2022-06-24
RD-5217 Automatically generate plugin_1_3.yaml

mateumann:
... but only in case it's required because user did not upload any
backward compatible `*_1_3.yaml` file with the plugin.


mateumann:
Needs https://github.com/cloudify-cosmo/cloudify-common/pull/1095

rest-service/manager_rest/upload_manager.py
```diff
@@ -1011,3 +1020,128 @@ def _flatten_labels_dict(labels):
...
+                               f"({yamls[0]}) can not be read.\n{e}")
+    incompatible_paths = {'types': [], 'availability_rules': []}
+    for selector in [
+        'data_types::*::properties::*',
+        'inputs::*',
+        'node_types::*::properties::*',
+        'relationships::*::source_interfaces::*::*::inputs::*',
+        'relationships::*::target_interfaces::*::*::inputs::*',
+        'workflows::*::parameters::*',
+    ]:
```
tehasdf, line 1048: I would prefer not to use `::` because eventually we'll again see an issue where someone uses `xxx::` as the name of their node. Perhaps those could just be tuples

rest-service/manager_rest/upload_manager.py
```diff
@@ -697,8 +698,16 @@ def _prepare_and_process_doc(self,
...
+                wagon_target_path, yamls_target_path = \
                     self._verify_archive(archive_target_path)
+                plugin_yaml, changes = _required_bc_changes(yamls_target_path)
+                if changes:
+                    bc_plugin_yaml = _make_bc_plugin_yaml(plugin_yaml, changes)
+                    _save_bc_plugin_yaml(
+                        os.path.join(archive_target_path, 'plugin_1_3.yaml'),
+                        bc_plugin_yaml
+                    )
+
```
tehasdf, line 710: It seems the split of "find changes first, then pass them to the `make` function" just makes things harder, because now the "find" function needs to structure the paths...

Instead, I think it'd be easier if the interface was just a single function, that would update the yaml as soon as it found any incompatible items.  (and it would presumably have a flag `found = False` or something, so that it could do nothing if there's nothing to do)

rest-service/manager_rest/upload_manager.py
```diff
@@ -1011,3 +1020,128 @@ def _flatten_labels_dict(labels):
     for k, v in labels.items():
         flattened[k] = v.get('values')
     return flattened
+
+
+def _required_bc_changes(yamls):
```
tehasdf, line 1025: I think it makes sense to put all of these in a separate module

rest-service/manager_rest/upload_manager.py
```diff
@@ -1011,3 +1020,128 @@ def _flatten_labels_dict(labels):
...
+        for path, rules in nodes.items():
+            if rules:
+                incompatible_paths['availability_rules'].append(path)
+    return plugin_yaml, incompatible_paths
+
+
+def _make_bc_plugin_yaml(plugin_yaml, incompatible_paths):
+    for data_type_path in incompatible_paths['types']:
+        _substitute_tree_node(
+            plugin_yaml, f'{data_type_path}::derived_from', 'string')
```
tehasdf, line 1071: why `derived_from`? Did we do anything 1.4-specific with it?

rest-service/manager_rest/upload_manager.py
```diff
@@ -1011,3 +1020,128 @@ def _flatten_labels_dict(labels):
...
+        'saving automatically generated backward compatible plugin yaml')
+    with open(file_path, 'w') as fh:
+        yaml.safe_dump(plugin_yaml, fh)
+
+
+def _nodes_in_tree(data, path):
+    """Find leafs of `data` pointed by `path`.
+
+    data={'a': {'b': {'q': 1, 'w': 2}, 'c': {'q': 3, 'w': 4}}}
+    _nodes_in_tree(data, 'a::*::w') == {'a::b::w': 2, 'a::c::w': 4}
```
tehasdf, line 1093: If you do this, might as well use the actual doctest syntax :)

tehasdf (commented):
all the tree/node functions are prime targets for unittesting!

tehasdf (approved):

PR #3741, opened 2022-07-05
RD-4466 The new update-or-reinstall flow

tehasdf:
Now, node-instances, instead of just always being reinstalled, will
be either reinstalled, or have the update interfaces called instead.


tehasdf:
needs cloudify-cosmo/cloudify-common#1102

geokala (approved):

PR #3746, opened 2022-07-07
Rd 5334 pagination metadata total fix

geokala:


geokala:
For the obvious comment of: Do we really need get_all_results on manager/db/broker node listings?
Currently: No, unless someone lowers the pagination count to 2.
However, given discussions around microservices we might want this in future and in any case it also skips an extra query.

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -463,14 +464,21 @@ def _paginate(query, pagination, get_all_results=False, locking=False):
...
             results = query.all()
         else:
             results = query.limit(size).offset(offset).all()
 
+        if get_all_results:
+            total = len(results)
+        else:
+            # distinct is required to cope with many-many relationships
+            # (e.g., otherwise you get all 7 results with a total of 460)
+            total = query.order_by(None).distinct(
```
tehasdf, line 479: adding a distinct is worrying from a performance perspective. Let's check with the benchmarks.

tehasdf: baseline:
```
BENCH test_deployments.create: 0:00:53.176378
BENCH test_deployments.list: 0:00:35.887694
```
with the change:
```
BENCH test_deployments.create: 0:00:56.260330
BENCH test_deployments.list: 0:00:36.027018
```
identical to within measurement error, I'd say

tehasdf (approved):

PR #3751, opened 2022-07-07
Rd 3556 build rhel8

glukhman:


yuvlar (dismissed):

packaging/cloudify-agents.spec
```diff
@@ -7,6 +7,12 @@
 %define _source_payload w0.gzdio
 %define _binary_payload w0.gzdio
 
+# this prevents networkx<2 failure in RH8
```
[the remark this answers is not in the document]
glukhman: RH8 build by default mangles shebangs, which fails with the test files of networkx<2 due to RH8 having python3 rather than python as its python command (e.g. https://github.com/networkx/networkx/tree/v1.11/networkx/tests). 
Since we can't change files of an external library, we just prevent this mangling in the first place

packaging/cloudify-mgmtworker.spec
```diff
@@ -2,6 +2,12 @@
 %define __jar_repack %{nil}
 %define PIP_INSTALL /opt/mgmtworker/env/bin/pip install -c "${RPM_SOURCE_DIR}/packaging/mgmtworker/constraints.txt"
 
+# this prevents networkx<2 failure in RH8
```
geokala, line 5: Do we need to make it conditional? If it'll work with RH/C 7 as well then maybe let's just do it.

Also, like @tehasdf said above, it'd be good to put a comment in saying what failures it is sidestepping (in the spec itself).

glukhman: I'm not positive that RH7 build even has `__brp_mangle_shebangs`, but I can't confirm it

packaging/cloudify-agents.spec
```diff
@@ -7,7 +7,10 @@
 %define _source_payload w0.gzdio
 %define _binary_payload w0.gzdio
 
-# this prevents networkx<2 failure in RH8
+# Prevent mangling shebangs (RH8 build default), which fails
+#  with the test files of networkx<2 due to RH8 not haveing python2.
```
geokala, line 11: *having

packaging/cloudify-agents.spec
```diff
@@ -7,7 +7,10 @@
...
+# Prevent creation of the build ids in /usr/lib, so we can still keep our RPM
```
geokala, line 12: I think this is supposed to be next to the other define?

geokala (approved):

PR #3753, opened 2022-07-07
dep-update endpoint: return a dict from parse args

tehasdf:
I'm kinda tired of those 30-element tuples. I'll be adding more flags,
so this is going to be more readable.


geokala:
Can we maybe make this some sort of data class so that it's immediately obvious what it contains wherever it appears?

tehasdf:
@geokala I'll just inline the whole method, instead of thinking what to use as a return value of that method

geokala (approved):

PR #3757, opened 2022-07-08
RD-4466 dep-up: add force_reinstall

tehasdf:
This will force all node-instances that were changed, to be
reinstalled, instead of running their update interfaces.


tehasdf:
Related to cloudify-cosmo/cloudify-common#1109

workflows/cloudify_system_workflows/deployment_update/update_instances.py
```diff
@@ -210,6 +210,10 @@ def update_or_reinstall_instances(ctx, graph, dep_up, install_params):
             must_reinstall |= instance.get_contained_subgraph()
     instances_to_update = instances_with_drift - must_reinstall
 
+    if install_params.force_reinstall:
```
tehasdf, line 213: this is the logic of it

geokala (approved):

PR #3760, opened 2022-07-11
RD-5229 Snapshot create improvements

glukhman:


rest-service/manager_rest/rest/resources_v2/snapshots.py
```diff
@@ -112,13 +112,21 @@ def put(self, snapshot_id):
             'queue',
             request_dict.get('queue', False)
         )
+        tempdir_path = request_dict.get('tempdir_path')
+        if tempdir_path and not os.access(tempdir_path, os.W_OK):
+            raise manager_exceptions.ForbiddenError(
+                f'Temp dir cannot be created inside unwriteable location '
+                f'{tempdir_path}'
+            )
```
tehasdf, line 120: I note that this won't work in the case of a future splitting the mgmtworker/restservice into separate containers, but I assume this snapshot code isn't planning to live as long

geokala: Yeah, this code needs to be gone soon.

geokala (approved):

PR #3763, opened 2022-07-11
RD-5334 Fix exc/dep group pagination totals

geokala:


tehasdf:
This unfortunately ends up doing 1+N queries :/

But if we mangled `_include` in the endpoint, by doing eg this:
```
        if 'execution_ids' in _include:
            _include.remove('execution_ids')
            _include.append('executions')
```
then it seems it would've been fine. Then, sm.list gets the relationship (not property or assoc-proxy) include, and that one seems to be handled correctly, and `marshal_with` gets the original include, and uses the property (and doesn't fire a query because the executions are already loaded).

Ditto deployments.

Also, I suggest adding an assert to the test as well, for example:
```
        assert {gr.id: set(gr.execution_ids) for gr in groups} == {
            'group1': {'exc1', 'exc2'},
            'group2': {'exc3', 'exc4'},
        }
```

tehasdf (dismissed):

tehasdf (dismissed):

rest-service/manager_rest/test/endpoints/test_deployment_groups.py
```diff
@@ -1022,6 +1023,76 @@ def test_create_from_group(self):
...
+            _include=['id', 'execution_ids'],
+            _get_data=True,
+        )
+        # before RD-5334, this used to return 4 (because there's 4 executions)
+        assert groups.metadata.pagination.total == 3
+        assert groups.metadata.pagination.size == 3
+        assert len(groups) == 3
+        assert {group.id: set(group.execution_ids) for group in groups} == {
+            'group1': {'exc1', 'exc2'},
+            'group2': {f'exc_group2 {i}' for i in range(6)},
```
tehasdf, line 1092:
```suggestion
            'group2': {f'exc_group2_{i}' for i in range(6)},
```

tehasdf (approved):

PR #3768, opened 2022-07-15
RD-4680 Support event restore with correct tenant

geokala:


rest-service/manager_rest/rest/resources_v3/events.py
```diff
@@ -50,22 +51,37 @@ def post(self):
...
-        if exc is None:
-            exc_id = request_dict.get('execution_id')
-            if exc_id is None:
+
+        exc_id = request_dict.get('execution_id')
+        eg_id = request_dict.get('execution_group_id')
+
+        if exc_id and eg_id:
+            raise manager_exceptions.ConflictError(
+                'Both execution_id and execution_group_id were provided.')
```
tehasdf, line 60:
```suggestion
                'Cannot provide both execution_id and execution_group_id.')
```
In general, error messages should talk about an error, not just make a statement of fact

rest-service/manager_rest/security/authorization.py
```diff
@@ -17,31 +16,27 @@
...
-            else:
-                tenant_name = tenant_for_auth
 
-            # finding tenant to add to the app config
+            execution = execution_token.current_execution
+            if execution and allow_if_execution:
+                if execution.tenant.name == tenant_name or not tenant_name:
+                    utils.set_current_tenant(execution.tenant)
+                    return func(*args, **kwargs)
+
```
tehasdf, line 39:
```suggestion
            if allow_if_execution:
                if current_execution and (
                    current_execution.tenant.name == tenant_name
                    or not tenant_name
                ):
                    utils.set_current_tenant(current_execution.tenant)
                    return func(*args, **kwargs)
```
(plus change the import to `from manager_rest.execution_token import current_execution`)
I suggest this way:
1. The whole special-cased code is now only under the `allow_if_execution`, which is the rare case
2. Tbh naming `execution = current_execution` is ugly: `current_execution` is a proxy and keeping that name helps reminds us of that

rest-service/manager_rest/rest/resources_v3/events.py
```diff
@@ -50,22 +51,37 @@ def post(self):
...
-                    'an execution token')
-            exc = sm.get(models.Execution, request_dict.get('execution_id'))
-        exc_params = {
-            '_execution_fk': exc._storage_id,
+                     'No execution passed, and not authenticated by '
+                     'an execution token')
+
+            exc_params['_execution_fk'] = exc._storage_id
+        exc_params.update({
             '_tenant_id': exc._tenant_id,
```
tehasdf, line 79: if `eg_id` is set, then `exc` is just a NameError?

rest-service/manager_rest/security/authorization.py
geokala: OK, reworking this into a separate commit to include the updated import.

rest-service/manager_rest/rest/resources_v3/events.py
geokala: if eg_id is set you can't reach this path.

tehasdf: seems to me you can!

geokala: Good catch, updated.

tehasdf (approved):

PR #3775, opened 2022-07-20
RD-5353 dep-up: skip updating uninitialized instances

tehasdf:
Things that were never installed, must not be touched by update at all.

### DIFF
--- a/workflows/cloudify_system_workflows/deployment_update/update_instances.py
+++ b/workflows/cloudify_system_workflows/deployment_update/update_instances.py
@@ -177,6 +177,11 @@ def _clean_drift(ctx, instance):
 def update_or_reinstall_instances(ctx, graph, dep_up, install_params):
     to_skip = set(install_params.added_instances) \
               | set(install_params.removed_instances)
+    # update must not touch instances that weren't installed previously
+    to_skip |= {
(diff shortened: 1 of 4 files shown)


workflows/cloudify_system_workflows/deployment_update/update_instances.py
```diff
@@ -173,8 +177,13 @@ def _clean_drift(ctx, instance):
 def update_or_reinstall_instances(ctx, graph, dep_up, install_params):
     to_skip = set(install_params.added_instances) \
               | set(install_params.removed_instances)
+    # update must not touch instances that weren't installed previously
+    to_skip |= {
+        ni for ni in workflow_ctx.node_instances
+        if ni.state != 'started'
```
geokala, line 183: Should this maybe just check for ones that == 'uninitialised'?

tehasdf: and then `deleted`? :)
and deleting? stopped? Would it make sense to try and update a `configuring` instance?

All pretty hard questions, and so, I chose to go with the easy answer

geokala: Fair enough, will green then.

geokala (approved):

PR #3790, opened 2022-08-10
RD-5511 Don't pack CentOS 7 agent (Python2-based)

mateumann:


mateumann:
Actually, let's use centos7-py3 agent as a drop-in replacement for the centos7 agent: https://github.com/cloudify-cosmo/cloudify-build-system/pull/702.

PR #3792, opened 2022-08-12
RD-5599 Drop test for plugin.yaml import with unknown dsl_version

mateumann:


tehasdf (approved):

mateumann:
Let's keep testing it.  Maybe someone somewhere will try importing `plugin.yaml` without `dsl_version` context.

PR #3795, opened 2022-08-15
RD-4684 Support restoring execution schedules

geokala:
Deployment workflows weren't being restored and that broke this.

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/deployments.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/deployments.py
@@ -236,20 +236,18 @@ class DeploymentsId(resources_v1.DeploymentsId):
         with sm.transaction():
             if not skip_plugins_validation:
                 rm.check_blueprint_plugins_installed(blueprint.plan)
-            dep_args = {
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -236,17 +236,20 @@ def put(self, deployment_id, **kwargs):
...
-                created_by=owner,
-            )
+                "created_at": created_at,
+                "created_by": owner,
+            }
+            workflows = request_dict.get('workflows')
+            if workflows:
+                dep_args['workflows'] = workflows
+            deployment = rm.create_deployment(blueprint, deployment_id,
+                                              **dep_args)
```
tehasdf, line 252: you don't need to rewrite all this, `rm.create_deployment` has a `if workflows`, so just a `rm.create_deployment(..., workflows=request_dict.get('workflows')` here works

geokala: Good point, will update.

tehasdf (approved):

PR #3801, opened 2022-08-22
RD-5653 inte-tests: show wait-for-starter output

tehasdf:
Showing the output of the manager bootup in the logs is going to make
it way easier to debug some jobs on jenkins (eg. the current aarch tests
failure)

### DIFF
--- a/tests/integration_tests/tests/test_cases.py
+++ b/tests/integration_tests/tests/test_cases.py
@@ -27,7 +27,6 @@ from integration_tests.tests.utils import (
     verify_deployment_env_created,
     run_postgresql_command,
     do_retries,
-    do_retries_boolean
 )
 
 from cloudify_rest_client.executions import Execution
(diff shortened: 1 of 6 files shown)


tests/integration_tests/tests/test_cases.py
```diff
@@ -91,9 +90,6 @@ def copy_file_to_manager(self, source, target, owner=None):
                 self.env.container_id,
                 ['chown', owner, target]
             )
-        do_retries_boolean(docker.file_exists,
-                           container_id=self.env.container_id,
-                           file_path=target)
```
geokala, line 96: I've been trying to figure out why this was even introduced, thought there might be some sort of race condition during the copy, but it doesn't look like that was ever commented, so... I guess we'll find out.

tehasdf: Perhaps there was, back when this was using ssh

geokala (approved):

PR #3802, opened 2022-08-24
restservice: update flask dependencies

tehasdf:
* inte-tests: drop --service-management

We're never testing on systemd. It's time to stop parametrizing it.
System-tests can keep testing it, it's a way more system-related
thing anyway.

* Update dependencies of the restservice

RD-5674 updated jinja2 in c-common, and now we need to do all kinds
of updates.
Soon, we'll have to update some of them again, after bumping python
to 3.7+, but that should then hopefully be just regenerating reqs.txt


glukhman (approved):

geokala:
System tests don't test it either, please make it be gone entirely.

tehasdf:
@geokala not yet. For that, let's make a ticket. But this PR is now mostly about having the build work :)

glukhman (approved):

geokala (approved):

PR #3807, opened 2022-08-30
inte-tests: stop testing cloudify 3.3.1

tehasdf:
No point anymore.


tehasdf:
these tests fail now, because supervisord is now the only way we run the tests

geokala (approved):

PR #3815, opened 2022-09-05
RD-5806 Replace Voluptuous with Pydantic

adrian-polanczyk-codilime:


tehasdf (commented):
In general, this is pretty great. I made sure to nitpick a lot, so that your first ticket is 100% sparkling perfect ;)

Additionally, I plan to use pydantic more in the restservice, so some of my ideas have that in mind.

1. Love it or hate it, this repo does 80 line length in flake. Many people hate it, I do like it, it's just what it is :P  (cf. flake8 in CI)
2. I've noted several comments about moving things from the decorator side, to the pydantic side. I think it's better.
3. Hilariously, with voluptuous, an invalid param does lead to a 500! And so, with this change, it is a 500 still. You can check it with (on manager):
```
curl -k -u admin:admin -H 'tenant: default_tenant' 'https://127.0.0.1/api/v3.1/blueprints?_sort=++a'
```
We can improve it and make it a 400 with a readable error, by putting something like this in `server.py`:
```python
from pydantic import ValidationError

@app_errors.app_errorhandler(ValidationError)
def _validation_error(e):
    return jsonify(
        message=str(e),
        error_code=manager_exceptions.BadParametersError.error_code,
    ), manager_exceptions.BadParametersError.status_code,

```
4. We now can note that the `@rangeable` (and `Range` etc) now are only related to datetime-ranges. Also, they're only used in https://github.com/cloudify-cosmo/cloudify-manager/blob/7cc413892c11f2321e77b6312437700cb19bbb6b/rest-service/manager_rest/rest/resources_v2/events.py - if you want, feel free to rename the decorator to something that says it's about datetimes - `@datetime_range` or something.

tehasdf (approved):

rest-service/setup.py
```diff
@@ -39,6 +38,7 @@
     'cachetools>=3,<4',
     'email-validator>1,<2',
     'retrying',
+    'pydantic',
```
geokala, line 41: We maybe want to pin this to make sure it doesn't jump to the next major version arbitrarily?

tehasdf: I think it's fine, we do have a concrete version in reqs.txt, so this only makes for "how hard is it to upgrade". Things like pydantic should be... very easy, on major versions

geokala: Fair point, request withdrawn!

mateumann (approved):

PR #3818, opened 2022-09-07
REST tests: remember to clean up test request context

tehasdf:
We can't just push and push, gotta pop sometimes too :)

Otherwise, the request context sticks around, and in each next
request, we still have the same current_user.

This seemed to work fine in 3.6, but started erroring out in 3.10,
I'm not entirely sure why

### DIFF
--- a/rest-service/manager_rest/test/base_test.py
+++ b/rest-service/manager_rest/test/base_test.py
@@ -303,6 +303,10 @@ class BaseServerTestCase(unittest.TestCase):
         cls.client = cls.create_client()
 
     def setUp(self):
+        test_context = server.app.test_request_context()
+        test_context.push()  # type: ignore
+        self.addCleanup(test_context.pop)
+
         self._handle_default_db_config()
         default_tenant = models.Tenant.query.get(0)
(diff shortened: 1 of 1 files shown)


glukhman (approved):

rest-service/manager_rest/test/base_test.py
```diff
@@ -303,6 +303,10 @@ def setUpClass(cls):
         cls.client = cls.create_client()
 
     def setUp(self):
+        test_context = server.app.test_request_context()
+        test_context.push()  # type: ignore
```
geokala, line 307: What makes us need ti not check type here?

tehasdf: hmm:
```
manager_rest/test/base_test.py:307: error: "ContextManager[RequestContext]" has no attribute "push"
manager_rest/test/base_test.py:308: error: "ContextManager[RequestContext]" has no attribute "pop"
``` I'm not sure why. It does have those.

tehasdf: Ah, flask 2.x fixes this

PR #3826, opened 2022-09-12
RD-2873 Accept blueprints as multipart/form

geokala:
This means we can stop messing around with dict encoding in query params.

### DIFF
--- a/rest-service/manager_rest/test/endpoints/test_utils.py
+++ b/rest-service/manager_rest/test/endpoints/test_utils.py
@@ -1,19 +1,3 @@
-#########
-# Copyright (c) 2016 GigaSpaces Technologies Ltd. All rights reserved
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#       http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-#  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#  * See the License for the specific language governing permissions and
-#  * limitations under the License.
-
-
 def generate_progress_func(total_size, buffer_size=8192):
     """Generate a function that helps test upload/download progress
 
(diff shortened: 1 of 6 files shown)


geokala:
This is part of the following batch of PRs:
https://github.com/cloudify-cosmo/cloudify-common/pull/1144
https://github.com/cloudify-cosmo/cloudify-manager/pull/3826
https://github.com/cloudify-cosmo/cloudify-cli/pull/1439
https://github.com/cloudify-cosmo/cloudify-stage/pull/2171

As noted in several places, it's to avoid us having to do silly things with encoding of query params (and thus make the new snapshots not require more ugliness).

tehasdf (dismissed):

rest-service/manager_rest/test/mocks.py
```diff
@@ -117,6 +103,23 @@ def _do_request(self, requests_method, request_url, body, params, headers,
...
+                        and body[entry][0] == 'filename'
+                    ):
+                        body[entry] = FileStorage(
+                            stream=open(body[entry][1].name, 'rb'))
+
+                # Remove application/json header to not override multipart
+                for header in headers:
+                    if headers[header] == 'application/json':
+                        break
+                headers.pop(header)
```
tehasdf, line 121: this will always remove the last header. Write this differently, eg. using a dictcomp

geokala: It won't, it'll remove the one that has the value of application/json- but yes, dictcomp would be better.

tehasdf: and if there isn't one that has the value of application/json, then it'll remove an unrelated header that just happens to iterate last

tehasdf: Also, that header's name is always content-type

geokala: Ah, good point, yes- anyway, dictcomp version incoming.

geokala: It's actually Content-type for the application/json. The one that gets added by the multipartencoder is Content-Type.

rest-service/manager_rest/test/endpoints/test_utils.py
```diff
@@ -26,15 +26,17 @@ def generate_progress_func(total_size, buffer_size=8192):
     iteration = [0]
     max_iterations = total_size // buffer_size
 
-    def print_progress(read, total):
+    def print_progress(watcher):
+        read_bytes, total_bytes = watcher.bytes_read, watcher.len
+
         i = iteration[0]
-        assert total == total_size
+        assert read_bytes == total_bytes
```
tehasdf, line 33: this one should be `total_size == total_bytes`

tehasdf (approved):
OK, because CI doesn't seem to work, but if it did, I'd have suggested the changes from #3831 first

PR #3830, opened 2022-09-13
jenkins: use the local version_info file for getting version

tehasdf:
no need to curl out to github (via getVersion). Just load the local packaging/version_info file.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -10,6 +10,21 @@ def vEnv(){
       virtualenv .venv
     ''', label: "creating virtualenv"
 }
+
+def doGetVersion(){
+  sh(script: '''#!/bin/sh -e
+    . cloudify-manager/packaging/version_info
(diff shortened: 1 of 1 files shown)


jenkins/Jenkinsfile
```diff
@@ -10,6 +10,21 @@ def vEnv(){
...
 }
+
+def doGetVersion(){
+  sh(script: '''#!/bin/sh -e
+    . cloudify-manager/packaging/version_info
+    echo ${CLOUDIFY_VERSION}
+  ''', label: 'get package version', returnStdout: true).trim()
+}
+
+def doGetPreRelease(){
```
geokala, line 21: Why not just get${thing}?

tehasdf: I'd like to avoid name conflicts between this and the pipeline-shared-library

geokala: After discussion externally: Namespacing fun.

geokala (approved):

PR #3831, opened 2022-09-14
Multipart blueprint upload: PR review changes

tehasdf:
See commits separately

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/blueprints.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/blueprints.py
@@ -91,10 +91,12 @@ class BlueprintsId(resources_v2.BlueprintsId):
         """
         rest_utils.validate_inputs({'blueprint_id': blueprint_id})
 
-        args = (
-            json.loads(request.form.get('params', '{}'))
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/rest/resources_v3_1/blueprints.py
```diff
@@ -91,10 +91,12 @@ def put(self, blueprint_id, **kwargs):
...
 
-        args = (
-            json.loads(request.form.get('params', '{}'))
-            or dict(request.args)
-        )
+        args = None
+        form_params = request.form.get('params')
+        if form_params:
+            args = json.loads(form_params)
+        if not args:
```
geokala, line 98: This should maybe be 'if args is not None', given that args could be {} if form params were passed in, but empty?
Are we likely to care about that condition?

geokala (approved):

rest-service/manager_rest/rest/resources_v3_1/blueprints.py
tehasdf: I note that it was the same before this change! But also, well, why not. Might as well check args.

PR #3836, opened 2022-09-21
RD-5930 Render DB schema to .sql

tehasdf:
This is the DB schema rendered by:

    cd /opt/manager/resources/cloudify/migrations/
    /opt/manager/env/bin/alembic upgrade --sql 272e61bf5f4a

For now, I'm just exposing this file right here. In the future, we'll probably replace this with something automatically rendering the newest version in CI

### DIFF
--- a/packaging/db-schema.sql
+++ b/packaging/db-schema.sql
@@ -1,0 +1,6958 @@
+BEGIN;
+
+CREATE TABLE alembic_version (
+    version_num VARCHAR(32) NOT NULL, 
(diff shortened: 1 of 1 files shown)


tehasdf:
You can easily start a container with this db, using eg.:
```
docker run -v /path/to/db-schema.sql:/docker-entrypoint-initdb.d/schema.sql -e POSTGRES_PASSWORD=cloudify -e POSTGRES_DB=cloudify_db -e POSTGRES_USER=cloudify postgres
```

geokala (approved):

PR #3838, opened 2022-09-21
RD-5928 Create Dockerfile for rest-service

adrian-polanczyk-codilime:


rest-service/setup.py
```diff
@@ -40,6 +40,7 @@
     'retrying',
     'pydantic',
     'distro',
+    'gunicorn',
```
tehasdf, line 43: no no no, this isn't a dependency of the restservice itself, that's just how it happens to be deployed (but it doesn't need to be gunicorn)

more practically: we don't want to have to install gunicorn to run tests

rest-service/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+ENTRYPOINT ["gunicorn"]
+CMD [\
+    "-k", "uvicorn.workers.UvicornWorker", \
+    "-u", "cfyuser", \
+    "-g", "cfyuser", \
+    "--pid", "/run/cloudify-api/pid", \
+    "--chdir", "/", \
+    "--workers", "0", \
+    "--max-requests", "4", \
+    "--bind", "127.0.0.1:8101", \
```
tehasdf, line 59: obviously this can't bind to 127.0.0.1

rest-service/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+RUN --mount=type=cache,target=/root/.cache apt-get install -y libpq-dev # Required for psycopg2
+RUN --mount=type=cache,target=/root/.cache apt-get install -y gcc # Required for psycopg2
+RUN --mount=type=cache,target=/root/.cache apt-get install -y libffi-dev # Required for cffi
+
+
+WORKDIR /opt
+
+# TODO: Change the way dependencies are installed
+COPY cloudify-common cloudify-common
```
tehasdf, line 27: should probably just use reqs.txt tbh

rest-service/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+ENTRYPOINT ["gunicorn"]
+CMD [\
+    "-k", "uvicorn.workers.UvicornWorker", \
+    "-u", "cfyuser", \
+    "-g", "cfyuser", \
+    "--pid", "/run/cloudify-api/pid", \
+    "--chdir", "/", \
+    "--workers", "0", \
+    "--max-requests", "4", \
```
tehasdf, line 58: wow a restart every 4 requests :)

rest-service/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+ENTRYPOINT gunicorn \
+  -u cfyuser \
+  -g cfyuser \
+  --pid /run/cloudify-restservice/pid \
+  --chdir / \
+  --workers $WORKER_COUNT \
+  --max-requests $MAX_REQUESTS \
+  --bind 0.0.0.0:$PORT \
+  --timeout 300 rest-service.manager_rest.wsgi:app \
```
tehasdf, line 57: `rest-service`? but this should be an import path...

tehasdf (commented):
OK I think this is good enough for now, we'll need to do something about the installation/dependencies in a way that allows for both fast building (none of this "directory has to contain" nonsense), but also for easy development too.

Just plz clean-up the dockerfile a bit:
 - CPU_RATIO is unused?
 - maybe we can just drop the `adduser` & `-u cfyuser -g cfyuser`. I'm not sure why would it be needed in docker. It was in the original script because that was aimed at a full system
 - consider putting the big huge `gunicorn` command into a separate .sh, like in the original. Maybe it'd be nicer. But maybe not :)
 - how about only a single `apt-get install` call for all 3 deps?
 - WORKER_COUNT=0 is a bad default, with 0 workers, well, it doesn't start at all. 0 doesn't mean "auto", it means zero :)

rest-service/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+RUN --mount=type=cache,target=/root/.cache apt-get update
+
+RUN --mount=type=cache,target=/root/.cache apt-get install -y libpq-dev # Required for psycopg2
+RUN --mount=type=cache,target=/root/.cache apt-get install -y gcc # Required for psycopg2
+RUN --mount=type=cache,target=/root/.cache apt-get install -y libffi-dev # Required for cffi
+
+
+WORKDIR /opt
+
+# TODO: Change the way dependencies are installed
```
tehasdf, line 26: see #3839 for some ideas of how I think this could look

docker-compose.yml
```diff
@@ -1,35 +1,36 @@
...
-    volumes:
-      - ../config.yaml:/etc/cloudify/config.yaml
-      - ../cloudify_internal_ca_cert.pem:/tmp/cloudify_internal_ca_cert.pem
-      - ../cloudify_internal_ca_key.pem:/tmp/cloudify_internal_ca_key.pem
+#  mgmtworker:
+#    container_name: mgmtworker
+#    build: mgmtworker
+#    env_file:
+#      - .env
+#    volumes:
```
adrian-polanczyk-codilime, line 9: I would like to stay `mgmtworker` as a "workable", so I decided to mount types and cert for now.

rest-service/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM python:3.10-slim
+
+ENV WORKER_COUNT=1
+ENV MAX_WORKER_COUNT=4
+ENV MAX_REQUESTS=1000
+ENV PORT=8101
```
mateumann, line 6: we use 8100 by default for rest-service

docker-compose.yml
mateumann: Actually it's entirely commented-out.  Maybe you could just drop these lines (for clarity if nothing else)

rest-service/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+      libffi-dev
+
+WORKDIR /opt
+
+COPY . rest-service
+
+RUN --mount=type=cache,target=/root/.cache pip install --upgrade pip
+RUN --mount=type=cache,target=/root/.cache pip install -r rest-service/requirements.txt
+RUN --mount=type=cache,target=/root/.cache pip install rest-service/
+RUN --mount=type=cache,target=/root/.cache pip install gunicorn
```
mateumann, line 28: maybe you could combine them in one `RUN` clause?

rest-service/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN mkdir /src
+COPY docker/cloudify.pth /usr/local/lib/python3.10/site-packages/cloudify.pth
+
+RUN mkdir -p /var/log/cloudify/rest
+RUN mkdir -p /opt/manager
+RUN mkdir -p /run/cloudify-restservice
```
mateumann, line 35: ditto

docker-compose.yml
adrian-polanczyk-codilime: Probably we will run that service in the future, so I decided to mount required volumes for that.

mateumann (approved):

PR #3839, opened 2022-09-22
dockerfiles: an idea for installing dependencies

tehasdf:
Here's my idea for installing dependencies in dockerfiles:
1. Just copy the local code (`COPY . x`), then `pip install -r reqs.txt && pip install .` - this is now very standard, no need for any additional `COPY`, no directory setup
2. Put a .pth file that will allow overriding that code we've installed

And so, to "just build an image", just a `docker build .` is enough. 
And if you're working on the service and want to use your local code, mount it while running a container, not while building the image: like
```
docker run -v /my/code/cloudify-common/cloudify:/src/cloudify -- restservice
```
This way, you don't need to rebuild after changing your code, just restart. Of course, if you want more dependencies, then you will need to rebuild.

**Commit messages have additional details** - there's a few more changes here:
- put the restservice entrypoint into a .sh
- instead of copying restservice config files, render them on startup from envvars
- no need for the separate user
- tune down the compose file: let's just run what we have working right now: DB & restservice. No "full manager", no mgmtworker (mgmtworker to be restored after nginx is here)

**Usage**
Just `docker compose up`, that's it. Then, you can use curl to check that it works:
```bash
curl  -u admin:admin ${RESTSERVICE_IP}:8101/api/v3.1/blueprints
curl ${RESTSERVICE_IP}:8101/api/v3.1/ok
```
The blueprint call won't work - the database is empty, there's no admin/admin. But you'll get a response.
The `ok` call will return "FAIL" because of course this isn't fully configured (no rabbitmq, no ssl). But you'll also get a response :)


tehasdf:
most of this was pulled into #3838

PR #3842, opened 2022-09-28
Fix aarch64 rpm build 1

glukhman:


packaging/cloudify-agents.spec
```diff
@@ -28,6 +28,7 @@ Vendor:         Cloudify Platform Ltd.
 Packager:       Cloudify Platform Ltd.
 
 Requires(pre):  shadow-utils
+Requires:       python3 >= 3.6
```
geokala, line 31: I think we're targeting minimum 3.10?

packaging/cloudify-mgmtworker.spec
```diff
@@ -29,14 +29,13 @@ BuildRequires:  postgresql-devel
...
+
 %description
 Cloudify's Management worker
 
 %prep
-
-# Download and untar our python3.10 package
-curl https://cloudify-cicd.s3.amazonaws.com/python-build-packages/cfy-python3.10.tgz -o cfy-python3.10.tgz
-sudo tar zxvf cfy-python3.10.tgz -C /
+sudo tar xf %{S:0} -C /
```
geokala, line 38: Does the Sources# auto download?

geokala: Looks like we do this already in premium, so I assume it's fine here too.

packaging/cloudify-agents.spec
geokala: In fact, this isn't a build-requires, it's install-time, but we bring our own python so I think this line is just wrong?

geokala: Needed for copy-packages at the moment. We probably want to nix this at some point, but let's leave it for now.

geokala (approved):

PR #3843, opened 2022-09-28
RD-5988 Fix agent package renaming

glukhman:
The previous code turned e.g. `manylinux-x86_64-agent_7.0.0-.dev1.tar.gz` into `manylinux-x86.tar.gz`

### DIFF
--- a/packaging/agents/copy_packages.py
+++ b/packaging/agents/copy_packages.py
@@ -19,7 +19,7 @@ def splitext(filename):
 
 
 def normalize_agent_name(filename):
-    return filename.split("_", 1)[0].lower()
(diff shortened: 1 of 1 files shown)


packaging/agents/copy_packages.py
```diff
@@ -19,7 +19,7 @@ def splitext(filename):
 
 
 def normalize_agent_name(filename):
-    return filename.split("_", 1)[0].lower()
+    return filename.rpartition("_")[0].lower()
```
geokala, line 22: rsplit('_', 1) would also work- any reason you preferred to shift to partition? (longer returned value though negligibly for what we're doing)

geokala (approved):

PR #3847, opened 2022-09-29
RD-6039 Allow sanity on community

geokala:


glukhman (dismissed):

geokala:
Removed the swagger pin- looks like it was a CI issue which probably wants a same-named branch in premium

mateumann (approved):

PR #3848, opened 2022-09-30
RD-5931 Docker image for NGINX

adrian-polanczyk-codilime:


tehasdf (commented):
this whole `$DOLLAR` business is a bit ugly. It'll make it hard to work on the config files later.

But we don't have to do it!

1. I note that the nginx image has templating built-in, we don't need to do it ourselves, we just need to put templates `/etc/nginx/templates` (see [the docs](https://hub.docker.com/_/nginx), ctrl+f "templates")
2. So I suggest taking the smallest possible subset of things that need templating, putting them in templates, and including those templated files.

From what I see, this is only going to be the `worker_processes`, `worker_connections`, and `REST_SERVICE_PORT`? In which case, we can have just two templates:
- `workers.conf.template` containing `worker_processes` + `events { ...` - and an `include workers.conf;` in the main file
- `rest_upstream.conf.template` with just that `upstream`, included where it currently is

nginx/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /opt
+
+RUN rm -rf /etc/nginx/conf.d/default.conf
+
+RUN mkdir -p /var/log/cloudify/nginx
+
+COPY ./config/nginx.conf /etc/nginx/nginx.conf
+COPY ./config/conf.d /etc/nginx/conf.d
+COPY ./config/templates /etc/nginx/templates
```
mateumann, line 15: Is `nginx/templates/` used anywhere?

nginx/config/conf.d/cloudify.conf
```diff
@@ -0,0 +1,7 @@
+# Declares the size of a request's body. This is mainly used to allow large blueprints to be uploaded.
+client_max_body_size 0m;
+
+# Overcoming the renowned http://stackoverflow.com/questions/19238122/nginx-with-ubuntu-and-rails-on-digital-ocean
+types_hash_max_size 4096;
+
+include "/etc/nginx/conf.d/*.upstream";
```
mateumann, line 7: Hmm… I don't see any `*.upstream` files here.  Where are they?

nginx/config/nginx.conf
```diff
@@ -0,0 +1,37 @@
+user  nginx;
+include /etc/nginx/conf.d/workers.vars;
```
mateumann, line 2: No such file or directory ;)

mateumann (approved):

mateumann (changes requested):
please see the questions

mateumann (approved):

nginx/Dockerfile
adrian-polanczyk-codilime: files inside `templates` will be translated to ` conf.d` directory automatically

PR #3853, opened 2022-10-11
RD-6086 inte-tests follow_events: use the non-deprecated method

tehasdf:
as of 3.10, this is the new way. Could also maybe rewrite this script to be asyncio.run, but this is fine too.


tehasdf:
cf https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop

geokala (approved):

PR #3866, opened 2022-10-18
RD-6073 Register RabbitMQ brokers in DB

adrian-polanczyk-codilime:


rest-service/manager_rest/configure_manager.py
```diff
@@ -25,6 +31,75 @@ def _generate_password(length=12):
...
+    return ''
+
+
+def _insert_rabbitmq_ca_cert(cert, name):
+    sm = get_storage_manager()
+    inst = models.Certificate(
+        name=name,
+        value=cert,
+        updated_at=datetime.datetime.now(),
+        _updater_id=0,
```
tehasdf, line 80: as we found recently, `_updater_id` doesn't actually need to be set here, so we can omit this line (I think at some point it was not null?)

rest-service/manager_rest/configure_manager.py
```diff
@@ -25,6 +31,75 @@ def _generate_password(length=12):
...
+    )
+    sm.put(inst)
+
+    return inst.id
```
tehasdf, line 84: hmm, if we returned `inst` directly rather than `inst.id`, then creating `models.RabbitMQBroker(` could take `ca_cert=this_inst` rather than `_ca_cert_id`

The difference is that `.id` is not necessarily assigned immediately, only on commit, so we could have this not require a commit
(currently `sm.put()` commits)

PR #3869, opened 2022-10-19
configure_manager: load rabbitmqs before inserting tenants

tehasdf:
`configure_manager.py` tries to insert tenants, while also adding them to rabbitmq - that last part, only if a rabbitmq broker is configured.

Well, so we had better configure rabbitmq brokers FIRST :)

### DIFF
--- a/rest-service/manager_rest/configure_manager.py
+++ b/rest-service/manager_rest/configure_manager.py
@@ -205,6 +205,8 @@ def _setup_user_tenant_assoc(admin_user, default_tenant):
 
 def configure(user_config):
     """Configure the manager based on the provided config"""
+    _register_rabbitmq_brokers(user_config)
+
     default_tenant = _get_default_tenant()
     need_assoc = False
     if not default_tenant:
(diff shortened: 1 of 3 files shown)


rest-service/manager_rest/configure_manager.py
```diff
@@ -306,8 +306,7 @@ def _register_rabbitmq_brokers(user_config):
 
         # reload config after inserting rabbitmqs, so that .amqp_host
         # and others are set
-        db.session.commit()
```
geokala: Why don't you need to commit any more here?

tehasdf: We used to need to commit, because `.load_from_db` created a new connection to the db, and so it of course didn't see uncommitted entries.
Now, we pass through the session, so `.load_from_db` runs its queries in the context of the currently-open transaction anyway, so it does see the rows that aren't committed yet.

geokala (approved):

adrian-polanczyk-codilime (approved):

PR #3879, opened 2022-10-24
RD-6131 Align DSL versions in test blueprints

mateumann:
Mark `test_source_plugin_requires_old_package` as expected to fail (until we find a way to deal with plugins based on Python 3.6)


tests/integration_tests/tests/agent_tests/test_existing_vm.py
```diff
@@ -47,6 +47,8 @@ def test_existing_vm(self):
 class HostPluginTest(BaseExistingVMTest):
     BLUEPRINTS = 'dsl/agent_tests/plugin-requires-old-package-blueprint'
 
+    # We should find a way to deal with plugins based on Python 3.6
```
tehasdf, line 50: this will be un-xfailed after RD-6168

tehasdf (approved):

PR #3889, opened 2022-10-26
RD-6168 Don't test py36-based source plugin

mateumann:
... it is not compatible with Python 3.10


geokala:
We do still need to be able to support 3.6 based source plugins though, somehow.

tehasdf:
@geokala I don't easily see how. Source plugins can't declare what's their required python version (unlike managed plugins)

geokala:
Source plugins are somewhat legacy. We should perhaps default to it py3.6 unless $something?

tehasdf:
Eh, that's kinda silly.

I mean, python packages CAN say in setup.py what version do they want, so perhaps we'll want some parsing of it eventually. But source plugins are so rarely used that I thought it's not worth the effort (yet)

geokala:
In that case let's just default source plugins to 3.6 so that they keep working.

tehasdf:
That makes new source plugins not work, then. (3.10-only ones)

tehasdf (approved):
OK, this is to some degree removing a "feature", but let's do this for now. If we want to, we can revert this after RD-6196

PR #3897, opened 2022-10-28
RD-6091 Intrinsic functions in workflow parameters

mateumann:
Allow (the default) values for the custom workflow parameters to be declared using intrinsic functions, e.g.

```yaml
tosca_definitions_version: cloudify_dsl_1_5

inputs:
  a_blueprint_id:
    type: blueprint_id
  b_blueprint_id:
    type: blueprint_id

workflows:
  test_parameters:
    mapping: scripts/workflows/test_deployment_id_parameters.py
    parameters:
      blueprint_id:
        type: blueprint_id
        default: {get_input: a_blueprint_id}
```

or

```python
client.executions.start(
    deployment_id='d1',
    workflow_id='test_parameters',
    parameters={'blueprint_id': {'get_input': 'b_blueprint_id'}},
)
```

### DIFF
--- a/rest-service/manager_rest/storage/resource_models.py
+++ b/rest-service/manager_rest/storage/resource_models.py
@@ -1083,7 +1083,7 @@ def evaluate_workflow_parameters(deployment_id: str, parameters: dict):
     from manager_rest.dsl_functions import evaluate_intrinsic_functions
 
     to_evaluate = {k: v for k, v in parameters.items()
-                   if isinstance(v, dict)}
+                   if isinstance(v, (list, dict))}
     evaluated = evaluate_intrinsic_functions(to_evaluate, deployment_id)
     parameters.update(evaluated)
 
--- a/rest-service/manager_rest/storage/resource_models.py
+++ b/rest-service/manager_rest/storage/resource_models.py
@@ -1078,6 +1078,16 @@ class BlueprintsFilter(FilterBase):
     )
 
 
+def evaluate_workflow_parameters(deployment_id: str, parameters: dict):
+    # Keep this import line here because of circular dependencies
+    from manager_rest.dsl_functions import evaluate_intrinsic_functions
+
+    to_evaluate = {k: v for k, v in parameters.items()
+                   if isinstance(v, dict)}
+    evaluated = evaluate_intrinsic_functions(to_evaluate, deployment_id)
+    parameters.update(evaluated)
+
+
 class Execution(CreatedAtMixin, SQLResourceBase):
     def __init__(self, **kwargs):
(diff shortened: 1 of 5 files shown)


rest-service/manager_rest/storage/resource_models.py
```diff
@@ -1078,6 +1078,16 @@ class BlueprintsFilter(FilterBase):
     )
 
 
+def evaluate_workflow_parameters(deployment_id: str, parameters: dict):
+    # Keep this import line here because of circular dependencies
+    from manager_rest.dsl_functions import evaluate_intrinsic_functions
+
+    to_evaluate = {k: v for k, v in parameters.items()
+                   if isinstance(v, dict)}
```
tehasdf, line 1086: the `isinstance` like this would omit eg `parameter: [{'get_secret': 'x'}, {'get_secret': 'y'}]`, which is otherwise valid

tests/integration_tests/resources/dsl/scripts/workflows/test_deployment_id_parameters.py
```diff
@@ -1,11 +1,28 @@
+import json
 import sys
 
-import json
+from cloudify.decorators import workflow
+from cloudify.manager import get_rest_client
+from cloudify.workflows import ctx
 
 
+@workflow
```
tehasdf, line 9: this isn't actually a workflow function though. The whole script is run (so the workflow function is `script_runner.tasks.execute_workflow`)

tests/integration_tests/resources/dsl/scripts/workflows/test_deployment_id_parameters.py
```diff
@@ -1,11 +1,28 @@
...
+    if not node:
+        return
+    for ni in client.node_instances.list(node_id=node.id):
+        rp = ni.runtime_properties
+        rp.update({'tested': True, name: value})
+        client.node_instances.update(
+            ni.id,
+            version=ni.version,
+            runtime_properties=rp
+        )
```
tehasdf, line 25:
```suggestion
    node = ctx.get_node('test_node')
    if not node:
        return
    for ni in node.instances:
        rp = ni.runtime_properties
        rp.update({'tested': True, name: value})
        ctx.update_node_instance(
            ni.id,
            runtime_properties=rp,
            version=ni.version,
        )
```
In principle, let's avoid using the rest-client directly. That stuff is exposed on the workflow ctx.

(the real underlying reason is cfy local compat, which is irrelevant here, but... well, it also makes it shorter)

tehasdf (approved):

PR #3898, opened 2022-10-29
Python 3.11 compat

tehasdf:
Turns out, there's only very few (really - one) changes required to make the restservice compatible with Python 3.11 - and it makes the tests faster indeed!

The cloudify-premium thing is to some degree unrelated, but that's how I ended up running the tests, so I'm cleaning that up too.

### DIFF
--- a/rest-service/manager_rest/constants.py
+++ b/rest-service/manager_rest/constants.py
@@ -14,8 +14,18 @@
 #  * limitations under the License.
 
 from collections import namedtuple
-from enum import Enum
 from typing import List, Union, Tuple
+
+
+try:
+    from enum import StrEnum
+except ImportError:
+    from enum import Enum
+    # before python 3.11, StrEnum didn't exist. Trivial reimplementation:
+
+    class StrEnum(str, Enum):
+        def __str__(self):
+            return self.value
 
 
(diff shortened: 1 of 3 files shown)


tehasdf:
Running the tests on my machine,
With `-n 3`:
  - python 3.10 - 1:59
  - python 3.11 - 1:35
With `-n 4`:
  - python 3.10 - 1:35
  - python 3.11 - 1:18

It's free real est- I mean, free performance.

rest-service/manager_rest/constants.py
```diff
@@ -14,10 +14,20 @@
 #  * limitations under the License.
 
 from collections import namedtuple
-from enum import Enum
 from typing import List, Union, Tuple
 
 
+try:
+    from enum import StrEnum
```
geokala, line 21: It might make more sense to just support py3.11 and drop 3.10?

tehasdf: That's not gonna be as direct... we'd have to build it first, etc.

geokala (approved):

rest-service/manager_rest/constants.py
geokala: Fair enough, green.

PR #3903, opened 2022-11-03
RD-6275 Support restoring blueprint labels

geokala:


rest-service/manager_rest/rest/responses.py
```diff
@@ -152,18 +154,30 @@ def __init__(self, **kwargs):
...
+    """A blueprint or deployment label."""
+    key: str
+    value: str
+    created_at: Optional[str] = field(default=None, compare=False)
+    created_by: Optional[str] = field(default=None, compare=False)
+
+    resource_fields: dict = field(
+        default_factory=lambda: LABEL_FIELDS,
+        compare=False, repr=False, init=False,
+    )
```
tehasdf, line 177: this is a bad idea and let's not do this - adds some runtime overhead for no good reason.

Instead, class attributes are to be type-declared as `ClassVar` (or we can be even more specific here with a `ClassVar[Dict[str, Any]]` - but the values are indeed going to need to be Any...)

geokala: Does that withold them from the compare? Not that it matters, I guess... but it will still put them in the repr, which sucks.

geokala: Then again, I'm not sure we actually need resource_fields to be on the Label at all, checking at the mo.

geokala: OK, it is ClassVar[Mapping]ed up.

rest-service/manager_rest/resource_manager.py
```diff
@@ -2703,40 +2707,42 @@ def is_computed_label(self, resource, key):
...
-                               resource,
-                               labels_list,
-                               creator=None,
-                               created_at=None):
+                               labels_resource_model: Type[LabelBase],
+                               resource: SQLResourceBase,
+                               labels_list: list[Label],
+                               creator: str = None,
+                               created_at: Union[Optional[str],
+                                                 datetime] = None):
```
tehasdf, line 2739:
```suggestion
                               created_at: Optional[str | datetime]=None):
```

rest-service/manager_rest/resource_manager.py
```diff
@@ -2665,11 +2668,12 @@ def _send_hook(self, execution, new_status):
...
-                               labels_resource_model,
-                               resource,
-                               new_labels,
-                               creator=None,
-                               created_at=None):
+                               labels_resource_model: Type[LabelBase],
+                               resource: SQLResourceBase,
+                               new_labels: list[Label],
+                               creator: str = None,
+                               created_at: Union[Optional[str],
```
tehasdf, line 2675: ditto `|`

rest-service/manager_rest/resource_manager.py
```diff
@@ -2665,11 +2668,12 @@ def _send_hook(self, execution, new_status):
         workflow_executor.send_hook(event)
 
     def update_resource_labels(self,
-                               labels_resource_model,
-                               resource,
-                               new_labels,
-                               creator=None,
-                               created_at=None):
+                               labels_resource_model: Type[LabelBase],
```
tehasdf, line 2671:
```suggestion
                               labels_resource_model: type[LabelBase],
```

rest-service/manager_rest/resource_manager.py
```diff
@@ -2703,40 +2707,42 @@ def is_computed_label(self, resource, key):
...
                 labels_to_delete.add(label)
         return labels_to_delete
 
     def create_resource_labels(self,
-                               labels_resource_model,
-                               resource,
-                               labels_list,
-                               creator=None,
-                               created_at=None):
+                               labels_resource_model: Type[LabelBase],
```
tehasdf, line 2734:
```suggestion
                               labels_resource_model: type[LabelBase],
```

rest-service/manager_rest/rest/resources_v3_1/blueprints.py
```diff
@@ -327,37 +320,34 @@ def patch(self, blueprint_id, **kwargs):
...
-                rm.update_resource_labels(models.BlueprintLabel,
-                                          blueprint,
-                                          labels_list,
-                                          creator=creator,
-                                          created_at=created_at)
+        labels_list = None
+        if state == BlueprintUploadState.UPLOADED:
+            if request_dict.get('labels'):
+                labels_list = [Label(**label)
+                               for label in request_dict.get('labels')]
```
tehasdf, line 328:
```suggestion
                               for label in request_dict['labels']]
```
the .get is redundant here

rest-service/manager_rest/rest/responses.py
```diff
@@ -153,17 +156,24 @@ def __init__(self, **kwargs):
...
 @swagger.model
-class Label(object):
-
-    resource_fields = {
+@dataclass(unsafe_hash=True)
+class Label:
+    """A blueprint or deployment label."""
+    key: str
+    value: str
+    created_at: Union[Optional[str],
```
tehasdf, line 164: ditto `|`

rest-service/manager_rest/rest/responses.py
```diff
@@ -153,17 +156,23 @@ def __init__(self, **kwargs):
...
-    def __init__(self, **kwargs):
-        self.key = kwargs.get('key')
-        self.value = kwargs.get('value')
-        self.created_at = kwargs.get('created_at')
-        self.created_by = kwargs.get('created_by')
+    def to_dict(self) -> dict:
+        return {'key': self.key,
+                'value': self.value,
+                'created_at': self.created_at,
+                'created_by': self.created_by}
```
tehasdf, line 178: `dataclasses.asdict` perhaps

rest-service/manager_rest/rest/resources_v3_1/deployments.py
```diff
@@ -1260,7 +1257,7 @@ def _make_new_group_deployment(self, rm, group, new_dep_spec, count,
         new_id, is_id_unique = self._new_deployment_id(group, new_dep_spec)
         inputs = new_dep_spec.get('inputs', {})
         labels = rest_utils.get_labels_list(new_dep_spec.get('labels') or [])
-        labels += group_labels
+        labels.extend([Label(*label) for label in group_labels])
```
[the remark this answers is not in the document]
tehasdf: also btw, if you use `.extend` rather than `+=` for lists, then no need to make it a listcomp that just wastes memory. Just `.extend(Label() for label in group_labels)` is fine

rest-service/manager_rest/rest/responses.py
geokala: Note: After discussion this should perhaps use asdict within the method (so that it need not be imported elsewhere, but so that we get the behaviour even if we update the dataclass later).

rest-service/manager_rest/rest/resources_v3_1/deployments.py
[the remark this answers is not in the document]
geokala: Updated.

rest-service/manager_rest/rest/responses.py
geokala: Updated

tehasdf (approved):

PR #3905, opened 2022-11-03
RD-6241 yamllint on blueprint_upload workflow

mateumann:


mateumann:
Because of yamllint's GPL license, we won't be using that in the Manager.

PR #3902, opened 2022-11-02
[Snyk] Security upgrade wheel from 0.37.1 to 0.38.0

qooban:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to fix one or more vulnerable packages in the `pip` dependencies of this project.</h3>

#### Changes included in this PR

- Changes to the following files to upgrade the vulnerable dependencies to a fixed version:
    - rest-service/requirements.txt

<details>
<summary>⚠️ <b>Warning</b></summary>

```
jsonschema 3.2.0 requires pyrsistent, which is not installed.
Flask-SQLAlchemy 2.5.1 requires SQLAlchemy, which is not installed.

```
</details>

##### By pinning:
Severity                   |  Issue                   | Upgrade                   | Breaking Change                   | Exploit Maturity
:-------------------------:|:-------------------------|:-------------------------|:-------------------------|:-------------------------
medium severity | Regular Expression Denial of Service (ReDoS) <br/>[SNYK-PYTHON-WHEEL-3092128](https://snyk.io/vuln/SNYK-PYTHON-WHEEL-3092128) |  `wheel:` <br> `0.37.1 -> 0.38.0` <br>  |  No  | No Known Exploit 

Some vulnerabilities couldn't be fully fixed and so Snyk will still find them when the project is tested again. This may be because the vulnerability existed within more than one direct dependency, but not all of the affected dependencies could be upgraded.

Check the changes in this PR to ensure they won't cause issues with your project.

------------

**Note:** *You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs.*

For more information:  
🧐 [View latest project report](https://app.snyk.io/org/cloudify-manager/project/6c769487-d4d1-49a4-8379-ab950a93e571?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)

🛠 [Adjust project settings](https://app.snyk.io/org/cloudify-manager/project/6c769487-d4d1-49a4-8379-ab950a93e571?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)

📚 [Read more about Snyk's upgrade and patch logic](https://support.snyk.io/hc/en-us/articles/360003891078-Snyk-patches-to-fix-vulnerabilities)

[...]

PR #3907, opened 2022-11-04
reqs.txt: bump wheel to 0.38.1

tehasdf:
For some reason, snyk won't leave me be with this pointless suggestion. It keeps suggesting 0.38.0, which is yanked. 0.38.1 exists though.


geokala (approved):

tehasdf:
cf #3901, #3902, #3906

PR #3914, opened 2022-11-09
RD-6264 Create and list Secret Providers

adrian-polanczyk-codilime:


tehasdf (commented):
Yeah I suggest sticking with the plural form for 1) the URL,  2) tablename;  similar to all other resources.

tehasdf (approved):

PR #3918, opened 2022-11-10
RD-2873 New snapshot create and restore

geokala:


tehasdf (approved):
OK, then let's just merge this, and get on rewriting it from scratch again

PR #3920, opened 2022-11-13
RD-6324 Evaluate depl. inputs intrinsic functions

glukhman:


mgmtworker/cloudify_system_workflows/deployment_environment.py
```diff
@@ -98,13 +98,26 @@ def _get_deployment_labels(new_labels, plan_labels):
     return [{k: v} for k, v in labels]
 
 
+def _evaluate_inputs_constraints(ctx, client, plan):
+    for in_key, in_val in plan.get('inputs', {}).items():
+        for constraint in in_val.get('constraints', {}):
```
tehasdf, line 103:
```suggestion
        for constraint in in_val.get('constraints', []):
```

mgmtworker/cloudify_system_workflows/deployment_environment.py
```diff
@@ -98,13 +98,32 @@ def _get_deployment_labels(new_labels, plan_labels):
...
+                ctx.deployment.id, {},
+                plan['inputs'],
+            )['payload']
+        except CloudifyClientError as e:
+            if e.status_code == 500:
+                raise ValueError(
+                    "Error evaluating deployment inputs. This could happen "
+                    "if the inputs contain an intrinsic function which "
+                    "requires the deployment to be parsed, such as `get_input`"
+                    " or `get_attribure`")
```
tehasdf, line 114:
```suggestion
                    " or `get_attribute`")
```

tehasdf, line 114: surely this error should also contain `e`? or at least a `ctx.logger` call?

mgmtworker/cloudify_system_workflows/tests/test_create_deployment.py
```diff
@@ -114,3 +114,52 @@ def test_labels_from_plan_and_kwargs(mock_ctx, mock_client, blueprint_plan):
...
+                            type(constraint['valid_values']) == list:
+                        parsed_values = []
+                        for item in constraint['valid_values']:
+                            if item == {'get_secret': 's1'}:
+                                parsed_values.append('y')
+                            elif item == {'get_secret': 's2'}:
+                                parsed_values.append('z')
+                            else:
+                                parsed_values.append(item)
+                        constraint['valid_values'] = parsed_values
```
tehasdf, line 134: I think this is WAY too much logic for a test util, and not enough assertions. What if obj isn't a dict, or the thing isn't a list? Well we don't really want to then return nothing, rather, we'd prefer to just fail an assert...
All in all, I think this function should looke more like,
```python
assert obj['a']['constraints']['valid_values'] == [...]
obj['a']['constraints']['valid_values'] = [hardcoded]
```
(+- more asserts as you see fit)

Anyway, the point is, I'd prefer test utils to be as straightforward as possible, and as strict as possible

mgmtworker/cloudify_system_workflows/tests/test_create_deployment.py
```diff
@@ -114,3 +114,52 @@ def test_labels_from_plan_and_kwargs(mock_ctx, mock_client, blueprint_plan):
     assert len(call.kwargs['labels']) == 2
     assert {'a': 'b'} in call.kwargs['labels']
     assert {'a': 'c'} in call.kwargs['labels']
+
+
+def test_inputs(mock_ctx, mock_client, blueprint_plan):
+    def parse_inputs(obj):
+        if type(obj) == dict:
```
tehasdf, line 121: btw there's `isinstance` for type checking :P

tehasdf (approved):

PR #3924, opened 2022-11-16
Rd 6368 secrets jsonschema

glukhman:


tehasdf:
OK so the main problem here is indeed the `server_default='{"type": "string"}'` - ie. how do we treat "old" secrets.

First, I note that this default doesn't mean schema is always going to be set. It can still be `None` :) But it's going to be stored as just json-null, not sql-null (you can insert the _text_ `'null'` into a not-null column, and then it json-decodes as none. And indeed, just `models.Secret(id='k3', value='xxx', ..)` is going to insert that...).

Anyway, I think I suggest the other way around, ie:
 - schema is nullable
 - if schema is set, then the value is always json (yes even type=string)
 - if schema is null, then the value is returned directly (ie. it's just a string)

This way, all the handling should be easier, and we won't need all the `if type != 'string'` stuff.

So:
1. Make the column nullable, and no server_default
2. In fact, no default anywhere (ie. drop the `schema = request_dict.get('schema', {'type': 'string'})`)
3. Then, in `_get_secret_params`, the validation, and the `json.dumps`, would go under a `if schema:`
4. Similar in `_update_value` - no `json.loads`, but indeed, `if schema:  validate();  value = json.dumps(value)`
5. In CLI, `if secret_flag_dict or secret_flag_list or schema: value = json.loads(value)`
6. in restclient, in `create`, no need for the `loads` or the `if` at all. Just pass schema & value directly.
7. In secrets GET, `if sec.schema: value = json.loads(value)` (probably no need for validation here...)
8. In restclient, the `get` change can be just dropped

I think this is better, because:
 - in db, we always store schema+JSON value, or no schema+string value (default, and for old secrets). So type=string isn't specialcased
 - I'm very worried about specialcasing type=string, because then we don't know if we store in the db a json-string (ie. with the quotes) or a bare-string (without quotes)
 - specialcasing type=string also means string validation is hard, or entirely missing (wrt. min/max length, regex pattern, or enum)

tehasdf:
BTW I've sent in #3927 to move the tests around a bit

tehasdf (commented):
cool, I think this should work now? remember to revert dev-reqs.txt though

PR #3925, opened 2022-11-17
RD-6395 REST: set sqlalchemy client_encoding to utf-8

tehasdf:
We don't ever want anything but utf-8 here.

If this is not set, it might default to ascii, and then saving non-ascii strings into the db, is going to cause UnicodeEncodeErrors.

### DIFF
--- a/rest-service/manager_rest/server.py
+++ b/rest-service/manager_rest/server.py
@@ -205,7 +205,8 @@ class CloudifyFlaskApp(Flask):
         the tables if necessary
         """
         self.config['SQLALCHEMY_ENGINE_OPTIONS'] = {
-            'pool_size': 1
+            'pool_size': 1,
+            'client_encoding': 'utf-8',
         }
         self.update_db_uri()
         self.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False


tehasdf:
With this,
```
2022-11-17 07:02:38.578  CFY <d2> [t1_qene5f.bar] Task started 'script_runner.tasks.run'
2022-11-17 07:02:38.849  LOG <d2> [t1_qene5f.bar] INFO: Downloaded scripts/script.py to /tmp/ZQ3LH/script.py
2022-11-17 07:02:38.866  LOG <d2> [t1_qene5f.bar] INFO: hello │    ├──  └── 
2022-11-17 07:02:39.018  CFY <d2> [t1_qene5f.bar] Task succeeded 'script_runner.tasks.run'
```

tehasdf:
FWIW, a test could look like this, but I wasn't able to make it fail. It seems my machine defaults to utf-8 indeed, and it doesn't seem to depend on locale. I don't know what it depends on.
This fails with `'client_encoding': 'sqlascii'`
```python
    def test_store_nonascii_event(self):
        exc = self.sm.put(Execution(
            id='exc1',
            workflow_id='wf1',
            creator=self.user,
            tenant=self.tenant,
        ))
        branch = '│   '
        tee = '├── '
        last = '└── '
        self.client.events.create(
            events=[{
                'message': {
                    'text': '{0} - {1} - {2}'.format(branch, tee, last),
                },
                'event_type': 'x',
                'context': {},
                'logger': 'root',
                'level': 'INFO',
            }],
            execution_id=exc.id,
        )
```

qooban (approved):

PR #3927, opened 2022-11-17
Move secrets tests to manager_rest/tests/

tehasdf:
No reason for these tests to be inte-tests! They work just fine as regular restservice tests - I pretty much just copied them verbatim :)

(almost - I replaced `self.assertX` with just pytest `assert`, and also changed assertRaises to the `with` syntax)


tehasdf:
already merged into the parent PR

PR #3930, opened 2022-11-21
RD-6427 create_dep_env: only evaluate input constraints

tehasdf:
Don't evaluate the whole inputs - defaults must stay un-evaluated here.
That means if the default contains other function calls, whenever we use `get_input` (eg. in a property), that function call is essentially copied there, and only evaluated on demand. Then, defaults can contain other get_input calls (or secrets, or attributes)


mgmtworker/cloudify_system_workflows/tests/test_create_deployment.py
```diff
@@ -150,18 +150,14 @@ def parse_inputs(obj):
                 ]
             }]
         },
-        'c': {
-            'type': 'string',
-            'default': {'concat': ['hello ', 'world']}
-        }
```
tehasdf, line 156: this one is removed, because this is the very behaviour we DON'T want after all

glukhman (approved):

PR #3934, opened 2022-11-22
RD-6065 Secret List by Provider name

adrian-polanczyk-codilime:


rest-service/manager_rest/rest/resources_v3/secrets.py
```diff
@@ -145,12 +146,23 @@ def _get_secret_params(key):
...
 
+        provider_name = request_dict.get('provider')
+        provider = models.SecretsProvider.query.filter_by(
+            name=provider_name,
+        ).first()
+
+        if not provider:
+            raise manager_exceptions.BadParametersError(
+                f'The Secrets Provider does not exist: {provider_name}',
+            )
```
tehasdf, line 157: ah this should go through storage_manager, so that the tenant/visibility filter is applied (based on the currently logged in user)

something like-
```python
provider = None
if provider_name := request_dict.get('provider'):
    sm = get_storage_manager()
    provider = sm.get(models.SecretsProvider, None, {'name': provider_name})
```
Then `.get()` automatically will just 404 if it doesn't exist

tehasdf: btw the `.get()` syntax is a bit stupid with the `None` there as the ID + separate name filter.

Hmm.
Secrets were able to make it better, eg. `secret = get_storage_manager().get(models.Secret, key)`
by doing (in the Secrets model)
```python
    @hybrid_property
    def key(self):
        return self.id
```
maybe we can do something similar here for nicer syntax (although same behaviour)

tehasdf (dismissed):

rest-service/manager_rest/storage/resource_models.py
```diff
@@ -303,6 +303,19 @@ def agent(cls):
...
+    __tablename__ = 'secrets_providers'
+
+    name = db.Column(db.Text, nullable=False)
+    type = db.Column(db.Text, nullable=False)
+    connection_parameters = db.Column(JSONString, nullable=True)
+    updated_at = db.Column(UTCDateTime)
+
+    @hybrid_property
+    def key(self):
+        return self.name
```
tehasdf, line 316: ah, i was wrong, sorry - this is entirely unnecessary, because secrets providers have both `id` and `name` already (we set both to the same value) (maybe we'll differentiate them in the future)

but anyway, `key` isn't related here

tehasdf (approved):

PR #3935, opened 2022-11-23
RD-5952 Add a multi-instance IDD test

tehasdf:
Well the test is pretty complex. But just read the docstring :)

Also, add a pre-check for a more readable error. This will happen when you do `{get_capability: {get_attribute: ..}}` in case the attribute returns a non-string. If we allowed the non-string to go into the db select, we'd get a psycopg2 db error, and that one was completely unreadable.

### DIFF
--- a/tests/integration_tests/tests/agentless_tests/test_inter_deployment_dependencies.py
+++ b/tests/integration_tests/tests/agentless_tests/test_inter_deployment_dependencies.py
@@ -228,6 +228,138 @@ node_templates:
         assert nis[0].runtime_properties['rel_operation'] == \
             'capability value'
 
+    def test_idd_with_context_multiple_instances(self):
+        """Check that IDDs can come from functions in scaled instances.
+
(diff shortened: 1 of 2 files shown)


tests/integration_tests/tests/agentless_tests/test_inter_deployment_dependencies.py
```diff
@@ -228,6 +228,138 @@ def test_dependency_func_with_context(self):
...
+            deployment_id='d2',
+            inputs={'value': 'value2'},
+        )
+        dep3 = self.deploy(
+            blueprint_id='bp2',
+            deployment_id='d3',
+            # I wish this wasn't required, but otherwise get_input is angry
+            # when it tries to be evaluated up-front, at the get_attribute
+            # call in its arguments
+            runtime_only_evaluation=True,
```
geokala, line 346: Can we just make that the default-and-only way now?

geokala (approved):

PR #3939, opened 2022-11-28
RD-6476 Rewrite upload_manager.py

tehasdf:
In this, I aim to rewrite `upload_manager.py`, in order to:
- I'm waging class warfare - replace all the UploadXXX classes, with just `upload_blueprint`, `upload_plugin`, etc functions
- for each kind of resource uploaded, there's a very clear place that moves the uploaded file to its final destination
- make the "plugin/caravan" part a bit less stupid. It now just decides whether the input is a caravan, wagon, or a zip, just right there in the function
- avoid copying the uploaded file unnecessarily

### Review note
Before reviewing, I suggest to first examine the [Rewrite UploadedSnapshotsManager](https://github.com/cloudify-cosmo/cloudify-manager/pull/3939/commits/052b08dd9f07f9b1caa2cd16cb5ea632387fb6ca) commit only. It is the smallest one here, and illustrates well the idea of the other changes too.

### .uploading directory

All the `upload_x` functions will store the file in `/<fileserver>/.uploading`, and then move it to the target location. For example, if you upload a snapshot named `abc`, it will be stored into `/opt/manager/resources/snapshots/.uploading/abc/abc.zip`, and after it's all uploaded, that will be renamed to `/opt/manager/resources/snapshots/abc/abc.zip`.
That way, we store in-flight uploads on the same filesystem as the target destination (not tmpdir), so the rename can hopefully be a very cheap operation (good chance that `.uploading` is on the same filesystem as the target location)

### DIFF
--- a/rest-service/manager_rest/dsl_back_compat.py
+++ b/rest-service/manager_rest/dsl_back_compat.py
@@ -5,12 +5,12 @@ import yaml
 from dsl_parser.constants import OBJECT_BASED_TYPES
 
 
-def create_bc_plugin_yaml(yamls, archive_target_path, logger):
+def create_bc_plugin_yaml(yamls, archive_target_path):
     if not yamls:
         raise RuntimeError("At least one yaml file must be provided")
 
     if any(filename.endswith('_1_3.yaml') for filename in yamls):
-        return None, None
+        return []
 
     # take the first yaml file - this is the default behaviour
     with open(yamls[0]) as fh:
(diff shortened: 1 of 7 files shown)


tehasdf:
TODO write actual commit messages

tehasdf:
TODO note this in jira

tehasdf:
TODO make sure the `.uploading` directories are always removed

mateumann (approved):

glukhman (approved):

PR #3943, opened 2022-11-29
Separate services: simplify `docker compose up`

tehasdf:
With this, the `d compose up` experience is much nicer!

I do recommend viewing commits separately for the details.

- download db schema from s3
- keep logs tidy (tidier?)
- add the license public key
- default the admin password to "admin" after all
- allow non-ssl access

With this, you can just `docker compose up`, and after it's spawned, you can:
```bash
cfy prof use localhost -u admin -p admin -t default_tenant
```
Then, you will be connected to this """manager""", and you can `cfy lic upl ...` and then `cfy blue li` and other stuff as needed.

For your development experience, other useful commands (that I use a lot) include:
- `rm ~/.cloudify -rf` - kills all your stored `cfy` profiles (ALL of them)
- `docker images -f 'reference=cloudify-manager-*' -q | xargs docker rmi -f` - nukes all cloudify-manager images (warning, this will kill AIO too, maybe you don't want that?)
- `docker system prune -f` / `docker volumes prune -f` - the usual "docker clean slate" approach. You maybe don't want that, if you have anything else in docker :)
- `docker compose rm -vaf` - for cleanup after your `docker compose up`


tehasdf:
Note that `cfy prof use` will keep throwing an error (but it'll stil work though) until we also merge cloudify-cosmo/cloudify-cli#1456

adrian-polanczyk-codilime (approved):

PR #3944, opened 2022-11-30
RD-6493 Replace no_external_authenticator with check_external_authenticator

tehasdf:
Now, those actions will be possibly allowed by the external auth! External auth handler can now return a "handler" for any of those actions.
If they don't (which is the default, authenticators need to opt-in), then the action is still disallowed.


tehasdf:
The other part of this is cloudify-cosmo/cloudify-premium#915

geokala (approved):

PR #3947, opened 2022-12-01
CYBL-1925 Fix cancel while having default 0 instances for servicecomp…

szpotona:
…onent

Possible if ServiceComponent instance count is 0. Bug prevents to cancel a workflow.


tehasdf (approved):
sure why not, although there's still the case of >1 NIs. But this at least prevents a 500, so we can get that in

tehasdf:
thanks! However, I can't seem to get Jenkins to run CI on external-fork PRs.

Anyway, I've implemented #3950 to handle this "0 instances" case, as well as the "more than 1 instance" case.

PR #3948, opened 2022-12-01
Blueprint upload: on skip_execution, still store the files

tehasdf:
Even when skip_execution is passed, we still want to call the upload_blueprint_archive_to_file_server function.

And then, we won't have any messages to send. And that also implies async_upload, as in, there's nothing to wait _for_.


tehasdf:
The second commit is mostly unrelated, but allows new-snap-res to run to completion (mostly, the test I've been using for this - `test_queued_executions_in_snapshot` - also checks queued executions, and new-snap-res doesn't support that)

mateumann (approved):

PR #3950, opened 2022-12-02
RD-6293 Executions cascade: handle all node-instances

tehasdf:
When starting/cancelling/resuming executions, make sure to look at all node instances that can contain a component deployment id.

This means:
- if there's 0 instances, well, don't throw a 500, but there's no cascade to be done
- if there's 1 instance, this will work the same as before
- if there's >1 instances, pull the deployment id from all of them.

Also, rewrite that whole method using DB-side filtering, rather than python-side.
If we do want this to be ported into 6.4.x, that "DB methods" commit must be skipped in the port, because 6.4.x doesn't have JSON columns.

### DIFF
--- a/rest-service/manager_rest/constants.py
+++ b/rest-service/manager_rest/constants.py
@@ -213,3 +213,4 @@ DEPLOYMENT_UPDATE_ACTION_TYPES = DeploymentUpdateActionTypes(
     REMOVE='remove',
     MODIFY='modify',
 )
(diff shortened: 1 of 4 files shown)


tehasdf:
inte-tests: `= 16 passed, 1 warning in 448.56s (0:07:28) =` for `integration_tests/tests/agentless_tests/component/test_component_cascading.py`

mateumann (approved):

PR #3956, opened 2022-12-05
RD-6401 Add a test for computing blueprint requirements

tehasdf:
Pretty much the same tests as in cloudify-cosmo/cloudify-common#1197 but as integration-tests; I want to check and be able to show that after the blueprint has been parsed, the requirements are indeed in the plan.


glukhman (approved):

tehasdf:
This PR used to be just about adding the tests, but now we store the requirements separately from the plan, and so now we also add a column and all that.

glukhman (dismissed):

PR #3957, opened 2022-12-06
RD-6173 blueprint set-icon fix

mateumann:


tests/integration_tests/tests/agentless_tests/test_blueprint_metadata.py
```diff
@@ -0,0 +1,20 @@
...
+        blueprint_id = 'bp'
+        self.client.blueprints.upload(
+            resource('dsl/empty_blueprint.yaml'),
+            entity_id=blueprint_id
+        )
+        self.client.blueprints.upload_icon(
+            blueprint_id,
+            resource('assets/cloudify.png')
+        )
+        self.client.blueprints.remove_icon(blueprint_id)
```
tehasdf, line 20: maybe we can have some assert between them?

mateumann: There is no good solution to assert icon existence (yet).  I am keeping that in mind though and will get back to it soon.

tehasdf (approved):

PR #3958, opened 2022-12-06
RD-6537 inte-tests: fix agent_tests/test_snapshot.py

tehasdf:
- After cloudify-cosmo/cloudify-common#1198, all agents are going to be listed
- Still, the download/reset/upload dance is going to be necessary, new snapshots don't have a `force=True` that removes everything from the manager
- Also make the deploy/undeploy concurrent in these tests. Saves quite a bit of time.


tests/integration_tests/tests/agent_tests/test_snapshots.py
```diff
@@ -103,23 +126,57 @@ def _restore_snapshot_multitenant(self, snapshot_id):
...
+
+        reset_storage(self.env.container_id)
+        self.client.snapshots.upload(downloaded_snapshot, snapshot_id)
         self._restore_snapshot(states, deployments, snapshot_id)
-        self.assertEqual(len(self.client.agents.list().items), 1)
+
+        # unlike the other tenant tests, in this one we FIRST reset + restore,
+        # and only then undeploy. This checks that agent connectivity is
+        # still available after that reset + restore
+        self.assertEqual(len(self.client.agents.list().items), 2)
```
geokala, line 145: It doesn't assert that with your changes though, because if they're not connected you'll still get two items?

geokala (approved):

tests/integration_tests/tests/agent_tests/test_snapshots.py
tehasdf: the undeploy being successful (not throwing an exception) checks that. Without the other changes, either way, this one was just checking the in-db state (just listing agents never checks connectivity)

PR #3962, opened 2022-12-07
RD-6512 Auto set related executions

geokala:


rest-service/manager_rest/rest/resources_v1/executions.py
```diff
@@ -201,6 +201,58 @@ def post(self, **kwargs):
                 execution.id = execution_id
             sm.put(execution)
             messages = []
+            if (
+                force_status in ExecutionState.STATES
+                and force_status not in ExecutionState.WAITING_STATES
+                and force_status != ExecutionState.PENDING
+            ):
+                if deployment_id:
+                    deployment = sm.get(models.Deployment, deployment_id)
```
tehasdf, line 210: `deployment` is already retrieved here (see 15 lines above), it'd be good if we were able to avoid additional queries

rest-service/manager_rest/rest/resources_v1/executions.py
```diff
@@ -201,6 +201,58 @@ def post(self, **kwargs):
...
+                            deployment.latest_execution.started_at)
+
+                    exec_relations_set = False
+                    if (
+                        deployment_latest_time is None
+                        or (
+                            deployment_latest_time
+                            and deployment_latest_time < started_at
+                        )
+                    ):
```
tehasdf, line 223: we do this stuff multiple times, maybe we can factor out a function here, something like

```python
def _is_execution_newer(started_at, execution):
    if not execution:
        return True
    x = parse(...)

# and then
if _is_execution_newer(started_at, dep.latest_execution):
    dep.latest_execution = execution
```

And this could be applied to latest execution, create execution, and blueprint upload execution

rest-service/manager_rest/rest/resources_v1/executions.py
```diff
@@ -201,6 +201,58 @@ def post(self, **kwargs):
                 execution.id = execution_id
             sm.put(execution)
             messages = []
+            if (
+                force_status in ExecutionState.STATES
+                and force_status not in ExecutionState.WAITING_STATES
+                and force_status != ExecutionState.PENDING
+            ):
+                if deployment_id:
```
tehasdf, line 209: maybe this whole case could be in a separate method, to avoid growing this `def post` even more?
Like
```python
def _on_execution_restored(self, sm, execution, deployment=None):
```

rest-service/manager_rest/rest/resources_v1/executions.py
geokala: Factored out.

rest-service/manager_rest/rest/resources_v1/executions.py
geokala: Agreed-ish- I've factored it out specifically for linked executions. if we have more things to deal with later then maybe we tweak it further.

tehasdf (approved):

PR #3974, opened 2022-12-12
RD-6566 configure_manager.py: insert manager & cert

tehasdf:
Move inserting the manager here, from c-m-i's script.

Also, insert that data in the config container.

### DIFF
--- a/rest-service/docker/entrypoint_config.sh
+++ b/rest-service/docker/entrypoint_config.sh
@@ -2,6 +2,16 @@
 
 /opt/rest-service/docker/prepare_secrets.sh
 
+if [ ! -e "/tmp/config.yaml" ]; then
+    echo "
+manager:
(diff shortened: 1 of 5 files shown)


rest-service/docker/entrypoint_config.sh
```diff
@@ -2,6 +2,16 @@
 
 /opt/rest-service/docker/prepare_secrets.sh
 
+if [ ! -e "/tmp/config.yaml" ]; then
```
geokala, line 5: Having this as a hard-coded path feels wrong and like it could lead to interesting issues. Can we make this accept a passed in arbitrary?

geokala: (issues such as lingering config.yamls or similar)

tehasdf: sure why not

PR #3978, opened 2022-12-13
RD-6064 Provider options for Secret resource

adrian-polanczyk-codilime:


resources/rest-service/cloudify/migrations/versions/edd6d829a209_6_4_to_7_0.py
```diff
@@ -643,10 +645,28 @@ def add_secrets_schema():
...
+
+
 def drop_secrets_schema():
     op.drop_column('secrets', 'schema')
 
 
+def drop_secrets_provider_options():
+    op.drop_column(
+        'secrets',
+        'schema',
```
tehasdf, line 666: :)

rest-service/manager_rest/dsl_functions.py
```diff
@@ -154,10 +154,28 @@ def get_secret_from_provider(secret):
...
+
+        if secret.provider_options:
+            provider_options = json.loads(
+                decrypt(
+                    secret.provider_options,
+                ),
+            )
+            secret_path = provider_options.get('path')
+
+        path = f'{default_path}'
```
tehasdf, line 170: if default path is not provided, then this will become `path = 'None'`, and that's not really what we want...

tehasdf (approved):

PR #3985, opened 2022-12-19
RD-6634 Update: run reinstalls before new installs

tehasdf:
Before, update handled instances in the order of:
1. Uninstall removed instances
1. Install added instances
1. Reinstall/update changed instances

That means if we add an instance, and also change an instance that the added one depends on, we need to run relationship operations twice: once after adding the new instance, and then after reinstalling the dependency.

With this change, we reinstall/update BEFORE installing new instances.
Thanks to this, we only need to run relationship operations once.

### DIFF
--- a/mgmtworker/cloudify_system_workflows/deployment_update/workflow.py
+++ b/mgmtworker/cloudify_system_workflows/deployment_update/workflow.py
@@ -368,7 +368,7 @@ def update_deployment_node_instances(*, update_id):
         ]
         workflow_ctx.create_node_instances(
             dep_up.deployment_id,
-            added_instances
+            added_instances,
         )
     if update_instances.get('extended_and_related'):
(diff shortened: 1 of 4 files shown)


mgmtworker/cloudify_system_workflows/deployment_update/workflow.py
```diff
@@ -368,7 +368,7 @@ def update_deployment_node_instances(*, update_id):
         ]
         workflow_ctx.create_node_instances(
             dep_up.deployment_id,
-            added_instances
+            added_instances,
```
mateumann, line 371: :heart:

tehasdf: I think I've been writing Go too much :P

mateumann (approved):

PR #3986, opened 2022-12-19
RD-6269 Cloudify Secrets Provider

adrian-polanczyk-codilime:


rest-service/manager_rest/dsl_functions.py
```diff
@@ -246,6 +311,41 @@ def check_vault_connection(url, token, path):
...
+    except requests.exceptions.RequestException as e:
+        status['status'] = False
+        status['message'] = e
+
+    return status
+
+
+def _get_cloudify_cert_temp_file(cert):
+    tmp_cert = tempfile.NamedTemporaryFile()
+    tmp_cert.write(bytes(cert, 'utf-8'))
```
tehasdf, line 343:
```suggestion
    tmp_cert = tempfile.NamedTemporaryFile(mode='w+')
    tmp_cert.write(cert)
```
no need to call bytes() ourselves, NamedTemporaryFile can very well work in text mode, with mode='w+' or mode='w+t' (not the default w+b)

rest-service/manager_rest/dsl_functions.py
```diff
@@ -246,6 +311,41 @@ def check_vault_connection(url, token, path):
...
+    status = {
+        'status': True,
+        'message': '',
+    }
+
+    if not connection_parameters.get('protocol'):
+        connection_parameters['protocol'] = 'https'
+
+    if cert := connection_parameters.get('cert'):
+        temp_cert_file = _get_cloudify_cert_temp_file(cert)
```
tehasdf, line 324: remember to remove this file, probably in a `finally` block

rest-service/manager_rest/dsl_functions.py
```diff
@@ -246,6 +311,41 @@ def check_vault_connection(url, token, path):
...
+        connection_parameters['cert'] = temp_cert_file.name
+
+    client = CloudifyClient(
+        **connection_parameters,
+    )
+
+    try:
+        status_response = client.manager.get_status()
+        status_response.raise_for_status()
```
tehasdf, line 333: Nope! `.get_status()` returns a dict. No need for `raise_for_status`, because the restclient will already throw if it receives a non-200 here.

For reference (pardon my trust-all in the example),
```python
>>> c = CloudifyClient(username='admin', password='admin', host='127.0.0.1', protocol='https', trust_all=True)
>>> c.manager.get_status()
/opt/manager/env/lib/python3.10/site-packages/urllib3/connectionpool.py:1045: InsecureRequestWarning: Unverified HTTPS request is being made to host '127.0.0.1'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  warnings.warn(
{'status': 'OK', 'services': {'Webserver': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'AMQP-Postgres': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'Management Worker': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'Manager Rest-Service': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'Cloudify API': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'Cloudify Execution Scheduler': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'Cloudify Console': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'PostgreSQL': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'RabbitMQ': {'status': 'Active', 'is_remote': False, 'extra_info': {'connection_check': 'OK'}}, 'Cloudify Composer': {'status': 'Active', 'is_remote': False, 'extra_info': {}}, 'Monitoring Service': {'status': 'Active', 'is_remote': False, 'extra_info': {}}}}
>>> 
```

rest-service/manager_rest/dsl_functions.py
```diff
@@ -246,6 +311,41 @@ def check_vault_connection(url, token, path):
...
+    except requests.exceptions.RequestException as e:
```
tehasdf, line 334: the restclient would throw a `CloudifyClientError` (`from cloudify_rest_client.exceptions import CloudifyClientError`) here, rather than a requests exception

rest-service/manager_rest/dsl_functions.py
```diff
@@ -246,6 +311,41 @@ def check_vault_connection(url, token, path):
...
+        status['status'] = False
+        status['message'] = e
```
tehasdf, line 336:
```suggestion
        status['message'] = str(e)
```
alas, an exception object won't be able to be jsonified

rest-service/manager_rest/dsl_functions.py
```diff
@@ -49,6 +55,15 @@
             'path',
         ],
     },
+    'cloudify': {
+        'connection_parameters': [
+            'host',
+            'username',
+            'password',
+            'tenant',
+            'cert',
```
tehasdf, line 64: let's not require `cert`, since we do allow http (not https). Also, even if we do use https, we can still omit the cert, and then we'll fall back to checking in the built-in CA store. That'll be the case when you're connecting to a manager that uses a cert signed by a public CA (eg. you used letsencrypt)

rest-service/manager_rest/dsl_functions.py
```diff
@@ -281,9 +381,10 @@ def get_node(self, node_id):
         return node.to_dict()
 
     def get_secret(self, secret_key):
-        secret = self.sm.get(Secret, secret_key)
-        decrypted_value = cryptography_utils.decrypt(secret.value)
-        return SecretType(secret_key, decrypted_value)
+        return get_secret_method(
+            secret_key,
+            sm=self.sm,
+        )
```
tehasdf, line 387: great

rest-service/manager_rest/rest/resources_v3/secrets.py
```diff
@@ -47,11 +54,13 @@ def get(self, key):
         else:
             # Returns the decrypted value
             try:
-                secret_dict['value'] = decrypt(secret.value)
+                secret_dict['value'] = get_secret_method(key)[1]
```
tehasdf, line 57:
```suggestion
                secret_dict['value'] = get_secret_method(key).value
```
let's keep using the attribute access, rather than index access, for this namedtuple. I don't really see a reason to switch to indexes here, and that's going to be more fragile (eg. if we add more fields there, or replace the SecretType with just the Secret model)

PR #3989, opened 2022-12-20
RD-6639 Remove /set-global endpoints

tehasdf:
set-visibility is available instead; the rest-client already uses set-visibility for its .set_global methods.

Set-global has been mentioned to be deprecated in the rest-docs since like 4.3, and we note that set-global was broken for some time already.

This is relevant for plugins, blueprints, secrets.

Also add a test for set-visibility, for each of the resources mentioned.


tehasdf:
Draft because Blueprints and Plugins need the same

mateumann (approved):

PR #4002, opened 2022-12-29
RD-6614 Deployment files sync in mgmtworker

mateumann:


rest-service/manager_rest/persistent_storage.py
```diff
@@ -11,6 +14,53 @@
...
 
 
+LAST_MODIFIED_FMT = '%Y-%m-%dT%H:%M:%S.%f%z'
+
+
+@dataclass(slots=True)
+class FileInfo:
+    """FileInfo is a class used to represent information about a file"""
+    filepath: str
+    checksum: str
```
tehasdf, line 24: we don't actually need the checksum, I think. Everything works off of just mtime

docker-compose.yml
```diff
@@ -11,6 +11,8 @@ services:
       - ssl:/etc/cloudify/ssl:ro
       - mgmtworker_workdir:/opt/mgmtworker/work
       - ./resources/rest-service/cloudify/types:/opt/manager/resources/cloudify/types
+    environment:
+      CFY_RESOURCES_ROOT: /tmp/resources
```
tehasdf, line 15: This really needs to stay the original `/opt/manager/resources/deployments/` (both here and in AIO), because of all the plugins/scripts that already have that path hardcoded

tehasdf:
```suggestion
      CFY_RESOURCES_ROOT: /opt/manager/resources
```

tehasdf:
This might still have some WIP tests, but I'll merge it, along with cloudify-cosmo/cloudify-manager-install#1457 in order to make the manager work again ;)

tehasdf (approved):

PR #4037, opened 2023-01-22
RD-6881 Component: don't break on zip blueprint_archive

tehasdf:
In here, we can throw more than just a JSONDecodeError: we can also throw a UnicodeDecodeError, and that commonly happens when `res` is a binary blob - which is common when the blueprint is a zip.


tehasdf:
Note: both JSONDecodeError and UnicodeDecodeError are derived from ValueError

adrian-polanczyk-codilime (approved):

PR #4047, opened 2023-01-31
RD-6913 Populate config table in configure_manager.py script

mateumann:


rest-service/manager_rest/configure_manager.py
```diff
@@ -623,8 +693,9 @@ def _wait_for_rabbitmq(address):
 
     config.instance.load_configuration(from_db=False)
     with setup_flask_app().app_context():
-        config.instance.load_from_db(session=db.session)
         user_config = _load_user_config(args.config_file_path)
+        populate_config_in_db(user_config)
```
tehasdf, line 697: move to inside `configure` I think, that's where all functions on this level are called from. Up here, at the top level, it only makes sense to place things that can be called separately, like the admin token thing

rest-service/manager_rest/configure_manager.py
```diff
@@ -588,6 +597,67 @@ def _wait_for_rabbitmq(address):
...
+def populate_config_in_db(cfg: dict):
+    config_for_db = generate_db_config_entries(cfg)
+    for scope, entries in config_for_db:
+        for name, value in entries.items():
+            inst = db.session.get(
+                models.Config,
+                {'name': name, 'scope': scope}
+            )
+            inst.value = value
+    db.session.commit()
```
tehasdf, line 658: no need to commit, we already do in `configure`. Committing here will make testing harder (which btw, might want to add some to `manager_rest/tests/infrastructure/test_configure_manager.py`

rest-service/manager_rest/configure_manager.py
```diff
@@ -509,8 +518,69 @@ def _add_deployments_filter(sys_filter_dict, creator, tenant, now):
...
 
 
+def _generate_db_config_entries(cfg):
+    manager_private_ip = cfg['manager'].get('private_ip', 'localhost')
+    default_file_server_url = f'https://{ipv6_url_compat(manager_private_ip)}'\
+                              f':{DEFAULT_INTERNAL_REST_PORT}/resources'
+
+    prometheus_cfg = cfg.get('prometheus', {})
+    rest_cfg = {
+        'rest_service_log_path': REST_LOG_DIR + '/cloudify-rest-service.log',
```
tehasdf, line 528:
```suggestion
        'rest_service_log_path': os.path.join(REST_LOG_DIR, 'cloudify-rest-service.log'),
```
(+flake, this line is too long i think)
You already use os.path.join up top, so let's maybe stay consistent

rest-service/manager_rest/test/infrastructure/test_configure_manager.py
```diff
@@ -329,11 +364,37 @@ def test_update_manager_cert_differs(self):
...
+
+        for scope, name, value in [
+            ('rest', 'public_ip', 'example.org'),
+            ('rest', 'file_server_url', 'https://localhost:53333/resources'),
+            ('mgmtworker', 'max_workers', 9),
+            ('agent', 'heartbeat', 99),
+        ]:
+            inst = db.session.scalars(
+                select(models.Config).filter_by(scope=scope, name=name)
+            ).first()
```
tehasdf, line 399:
```suggestion
            ).one()
```
if there's no entry, that should be already an error

tehasdf (approved):

PR #4059, opened 2023-02-14
RD-6908 dep-update: update plan with real instance counts

tehasdf:
If nodes were scaled, those new instance counts are not reflected in the plan.
Before creating the new deployment plan, update the blueprint plan with the current real instance counts.

### DIFF
--- a/workflows/cloudify_system_workflows/deployment_update/workflow.py
+++ b/workflows/cloudify_system_workflows/deployment_update/workflow.py
@@ -20,6 +20,43 @@ from .update_instances import update_or_reinstall_instances
 from .utils import clear_graph
 
 
+def _update_plan_nodes(plan):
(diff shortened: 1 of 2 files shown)


workflows/cloudify_system_workflows/deployment_update/workflow.py
```diff
@@ -20,6 +20,43 @@
...
+            continue
+
+        current_num = existing_node._node.get('number_of_instances')
+        current_planned = \
+            existing_node._node.get('planned_number_of_instances')
+
+        try:
+            # if the node has an `instances: deploy: {num}`, that overrides
+            # any other setting
+            force_deploy = node['instances']['deploy']
```
geokala, line 48: Huh, I would've expected this to have kept the original default (updated by whatever's in the update).

geokala (approved):

PR #4069, opened 2023-02-16
RD-7042 Execution groups: allow changing concurrency

tehasdf:
Just allow changing the setting at runtime, using a PATCH call. No need to start the executions up front: the de-queueing mechanism will pick it up, once an execution has finished.

And if you lower the setting, of course we won't immediately do anything either, it only really affects the de-queueing.


tehasdf:
needs cloudify-cosmo/cloudify-common#1245

adrian-polanczyk-codilime (approved):

PR #4064, opened 2023-02-16
[Snyk] Fix for 3 vulnerabilities

PatrycjaKubiczek:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to fix one or more vulnerable packages in the `pip` dependencies of this project.</h3>

#### Changes included in this PR

- Changes to the following files to upgrade the vulnerable dependencies to a fixed version:
    - api-service/requirements.txt

<details>
<summary>⚠️ <b>Warning</b></summary>

```
jsonschema 3.2.0 requires pyrsistent, which is not installed.
Flask-Security 3.0.0 requires Flask-WTF, which is not installed.

```
</details>

##### By pinning:
Severity                   | Priority Score (*)                   | Issue                   | Upgrade                   | Breaking Change                   | Exploit Maturity
:-------------------------:|-------------------------|:-------------------------|:-------------------------|:-------------------------|:-------------------------
medium severity  |  **581/1000**  <br/> **Why?** Recently disclosed, Has a fix available, CVSS 5.9  | Denial of Service (DoS) <br/>[SNYK-PYTHON-STARLETTE-3319937](https://snyk.io/vuln/SNYK-PYTHON-STARLETTE-3319937) |  `starlette:` <br> `0.22.0 -> 0.25.0` <br>  |  No  | No Known Exploit 
low severity  |  **416/1000**  <br/> **Why?** Recently disclosed, Has a fix available, CVSS 2.6  | Access Restriction Bypass <br/>[SNYK-PYTHON-WERKZEUG-3319935](https://snyk.io/vuln/SNYK-PYTHON-WERKZEUG-3319935) |  `werkzeug:` <br> `2.2.2 -> 2.2.3` <br>  |  No  | No Known Exploit 
high severity  |  **661/1000**  <br/> **Why?** Recently disclosed, Has a fix available, CVSS 7.5  | Denial of Service (DoS) <br/>[SNYK-PYTHON-WERKZEUG-3319936](https://snyk.io/vuln/SNYK-PYTHON-WERKZEUG-3319936) |  `werkzeug:` <br> `2.2.2 -> 2.2.3` <br>  |  No  | No Known Exploit 

(*) Note that the real score may have changed since the PR was raised.

Some vulnerabilities couldn't be fully fixed and so Snyk will still find them when the project is tested again. This may be because the vulnerability existed within more than one direct dependency, but not all of the affected dependencies could be upgraded.

[...]

PR #4088, opened 2023-03-06
Update dependencies in the restservice virtualenv

tehasdf:
This re-compiles requirements in the restservice venv; the main goal is updating `starlette` and `werkzeug`, for the snyk-reported vulns.

Note that all 3 of these are in the same venv, so they all must be updated in tandem...


tehasdf:
This expands on, and obsoletes, #4064

adrian-polanczyk-codilime (approved):

PR #4096, opened 2023-03-09
RD-6954 Add dep.latest_execution_workflow_id

tehasdf:
Just declaring the assoc-proxy, that's all.


tehasdf:
this puts #4095 in 7.0.0

mateumann (approved):

PR #4100, opened 2023-03-14
RND-176 - Configurable Prometheus URL

adrian-polanczyk-codilime:


tehasdf (approved):
we can start by merging this indeed, but eventually, it'd be good to move this into the config table

tehasdf (approved):

PR #4109, opened 2023-03-17
RND-206 storage_manager._get_query - load correct assoc-proxies

tehasdf:
This essentially rewrites most of the storage-manager query building (see first commit) in order to make sure filtering on assoc-proxies works correctly (see the "IDD filtering test" commit).

The first commit has the most of the "meat" of this. Sorry - I wasn't able to really make it smaller, for easier review! :(

Also, drop a few methods from the storage-manager.

Also, this should fix the inte-tests:
```
integration_tests/tests/agentless_tests/component/test_workflows.py -k test_components_update_inputs
integration_tests/tests/agentless_tests/component/test_workflows.py -k test_components_update_blueprint
```

A technically-back-incompatible change here, too: now we're no longer automatically including `schedules` in deployments: it seems it was never used, and every relationship is expensive to join there.


rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -128,151 +132,157 @@ def transaction(self):
...
-        query = self._add_substr_filter(query, substr_filters)
-        query = self._add_filter_rules(query, model_class, filter_rules, joins)
-        return query
+        query = model_class.query
+        resolved_fields = {}
+        rels = set()
+
+        for field_name in include:
+            field = getattr(model_class, field_name, None)
+            if not field:
```
mateumann, line 235: `is None`?

rest-service/manager_rest/storage/storage_manager.py
```diff
@@ -128,151 +132,157 @@ def transaction(self):
...
+        return query, resolved_fields, rels
+
+    def _resolve_value_filters(
+        self,
+        filters: dict[str, Any],
+        resolved_fields: dict[str, InstrumentedAttribute],
+    ) -> Iterable[sql.ClauseElement]:
+        for field_name, value in filters.items():
+            field = resolved_fields.get(field_name)
+            if not field:
```
mateumann, line 271: `is None`?

mateumann (approved):

PR #4128, opened 2023-03-28
RND-213 Add _common_only flag to searches/workflows

glukhman:
This allows using `/api/v3.1/searches/workflows?_common_only=True` to list only the workflows which are common between the filtered deployments

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/workflows.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/workflows.py
@@ -50,9 +50,10 @@ class Workflows(SecuredResource):
 
 
 def workflows_list_response(
-        deployments: Iterable[models.Deployment]
+        deployments: Iterable[models.Deployment],
+        common_only: bool = False,
 ) -> ListResponse:
-    workflows = _extract_workflows(deployments)
+    workflows = _extract_workflows(deployments, common_only=common_only)
     pagination = {
         'total': len(workflows),
         'size': len(workflows),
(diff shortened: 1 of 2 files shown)


rest-service/manager_rest/rest/resources_v3_1/workflows.py
```diff
@@ -50,24 +50,32 @@ def get(self, filters=None, search=None, filter_id=None, **kwargs):
...
 ) -> ListResponse:
-    workflows = _extract_workflows(deployments)
+    workflows = _extract_workflows(deployments, common_only=common_only)
     pagination = {
         'total': len(workflows),
         'size': len(workflows),
         'offset': 0,
     }
-    return ListResponse(items=[w.as_dict() for w in workflows],
+    return ListResponse(items=[w.name for w in workflows],
```
mateumann, line 62: Are you sure about this change (`w.as_dict()` → `w.name`)?  Is it enough?

rest-service/manager_rest/rest/resources_v3_1/workflows.py
```diff
@@ -50,24 +50,32 @@ def get(self, filters=None, search=None, filter_id=None, **kwargs):
...
 
 
 def _extract_workflows(
-        deployments: Iterable[models.Deployment]
+        deployments: Iterable[models.Deployment],
+        common_only: bool = False,
 ) -> List[Workflow]:
-    workflows = set()
+    if not deployments:
+        return list()
```
mateumann, line 71: or just `return []` (one function call less)

PR #4157, opened 2023-04-04
RND-373 Restore legacy snapshot code

mateumann:
It is possible now to create both legacy (default in 7.0) and new (default in 7.1) snapshots.


mgmtworker/cloudify_system_workflows/snapshots/snapshot_create_legacy.py
mateumann, line 1: This is simply a version of `snapshot_create.py` which we released with 6.4.1.

tehasdf (approved):

PR #4283, opened 2023-08-23
upload_missing_plugin: avoid calling .json() many times

tehasdf:
No need to parse the response more than once.


mgmtworker/cloudify_system_workflows/dsl_import_resolver/resolver_with_catalog_support.py
```diff
@@ -212,12 +216,16 @@ def _upload_missing_plugin(self, name, specifier_set, distribution,
...
         # available wagon of the most recent version
         if not matching_versions:
-            for version in plugin_versions.json().get('items'):
+            for version in available_versions:
                 parsed_version = parse_version(version.get('version'))
                 yaml_urls = version.get('yaml_urls')
+                if not yaml_urls:
+                    continue
                 if version.get('wagon_urls'):
                     wagon_url = version['wagon_urls'][0]['url']
```
tehasdf, line 225: if we don't go under this `if`, then `wagon_url` is still kept from the previous loop, leading to the wrong url, or NameError if we never set it

glukhman (approved):

PR #4323, opened 2023-10-17
NE-7622 - Upgrade dependencies for subdirs

adrian-polanczyk-codilime:


adrian-polanczyk-codilime:
License issue for `psycopg2` package was previously and is not caused by changes relative to this PR.

mateumann (approved):

PR #4465, opened 2024-07-31
NE-35712 Update rest-service dependencies

korbajan:
Set min required version with solved vulnerabilities


lukaszd128 (approved):
Looks ok

PR #4467, opened 2024-08-06
Fix blueprint API - in some cases the visibility parameter is optional

lukaszd128:
Fixes blueprint API (blueprint validate and blueprint upload) so the visibility argument is truly optional when parameters are passed as a form parameters. Additional fix for blueprint validate API endpoint to pass further parameter values as strings (not as one element list of strings which later fails validation).

### DIFF
--- a/rest-service/manager_rest/rest/resources_v3_1/blueprints.py
+++ b/rest-service/manager_rest/rest/resources_v3_1/blueprints.py
@@ -100,7 +100,7 @@ class BlueprintsId(resources_v2.BlueprintsId):
         created_at = args.get('created_at')
         created_by = args.get('created_by')
 
-        visibility = args.get('visibility')
+        visibility = args.get('visibility', None)
         private_resource = args.get('private_resource')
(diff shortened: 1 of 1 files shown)


mateumann (dismissed):

rest-service/manager_rest/rest/resources_v3_1/blueprints.py
```diff
@@ -100,7 +100,7 @@ def put(self, blueprint_id, **kwargs):
         created_at = args.get('created_at')
         created_by = args.get('created_by')
 
-        visibility = args.get('visibility')
+        visibility = args.get('visibility', None)
```
korbajan, line 103: `None` is the default value that is returned from dict get().

korbajan (dismissed):

rest-service/manager_rest/rest/resources_v3_1/blueprints.py
lukaszd128: I reverted that change. Thanks

korbajan (approved):

